Issue #10, opened 2021-08-23
RunGrepTest issue on OpenBSD


PhilipHazel:
This is #2680 in the old Bugzilla, submitted by Nam Nguyen.

RunGrepTest fails on Test 132 for me on OpenBSD.

```
./testdata/grepoutput expects an 'a' at the end.
---------------------------- Test 132 -----------------------------
match 1:
 a
match 2:
 b
---
 a
RC=0

```
I actually get no 'a' when I run RunGrepTest:
```
---------------------------- Test 132 -----------------------------
match 1:
 a
match 2:
 b
---
RC=0

```
Manually running this test I can't see how it produces 'a' after the '---':
```
$ (pcre2grep -m1 -A3 '^match'; echo '---'; head -1) < testdata/grepinput
match 1:
 a
match 2:
 b
---
```

PH: The idea of this test is to check that the standard input is left in the right place when pcre2grep stops because it has reached the -m limit. The "a" line is generated by the "head -1" command when run under Linux (which is all I have):

```
$ (./pcre2grep -m1 -A3 '^match'; echo '---'; head -1) < testdata/grepinput
match 1:
 a
match 2:
 b
---
 a
```

Looks like this is yet another Linux/BSD difference. Sigh. I think the relevant code is around line 2589 in pcre2test.c:

```
  /* If the -m option set a limit for the number of matched or non-matched
  lines, check it here. A limit of zero means that no matching is ever done.
  For stdin from a file, set the file position. */                         
                                                                            
  if (count_limit >= 0 && count_matched_lines >= count_limit)             
    {                                                                   
    if (frtype == FR_PLAIN && filename == stdin_name && !is_file_tty(handle))
      (void)fseek(handle, (long int)filepos, SEEK_SET);             
    rc = (count_limit == 0)? 1 : 0;                                     
    break;                                                             
    }
```

carenas:
FWIW, it also fails in Minix, NetBSD, FreeBSD and AIX (might fail also in other *NIX, but I don't have access to them to check).

Solaris also fails if using the default /bin/sh but succeeds if using bash, but somehow it works in Windows, macOS and Linux.

The underlying code is not compatible with ANSI C 17[1] which specifies in 7.21.9.2.4:

"For a text stream, either offset shall be zero, or offset shall be a value returned by an earlier
successful call to the ftell function on a stream associated with the same file and whence shall be
SEEK_SET."

[...]

PR #19, opened 2021-09-19
Add a GitHub Actions build workflow

ltrzesniewski:
Let's inaugurate the pull requests 🙂 

Here's a suggestion: you may use GitHub Actions for continuous integration. If you merge this, each time you push or receive a pull request, GitHub will build and test the code, then report success or failure with a ✔️ or ❌ sign on the commit or PR.

I wrote a very basic workflow (see the YAML file, it is pretty self-explanatory), but you may expand this to include more config options or combinations of options.

You can also see an [example of a run in my fork](https://github.com/ltrzesniewski/pcre2/runs/3644688236?check_suite_focus=true), or consult the [documentation](https://docs.github.com/en/actions).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,23 @@
+
+name: Build
+on: [push, pull_request]
+
+jobs:
+  linux:
+    name: Linux
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        
+      - name: Autogen
+        run: ./autogen.sh
+        
+      - name: Configure
+        run: ./configure --enable-jit --enable-pcre2-8 --enable-pcre2-16 --enable-pcre2-32
+        
+      - name: Build
+        run: make
+        
+      - name: Test
+        run: make check


PhilipHazel:
Thanks for the suggestion. I am a newbie to both Git and GitHub, and I'm not sure that I'm ready for sophistications like this yet. Also, there is currently a Release Candidate out for testing, so non-urgent changes are frozen (and I'm busy on an entirely different project).

PhilipHazel:
I have finally got round to merging this and will now research what it can do (including #10).

PhilipHazel:
Thank you. The first thing I want to do is to run the workflow manually,
which is clearly something needed for testing. Looks like I have to add the
workflow-despatch event to the YAML file, and then there will a "run"
button for the action. Is that right? I have to go out now for several
hours, so no hurry to respond.

ltrzesniewski:
The workflow is configured to run automatically:

- Every time you push, on any branch. If you push several commits at once, it will run on the last commit.
- Each time someone submits a PR, or adds a commit to a PR. This is useful for quickly making sure a PR is viable.

This means that, normally, you shouldn't need to run a workflow manually. If you want to test a change, all you need to do is push it. That's why I recommended testing workflow changes on a different branch: you may need to push several small commits, and squash them to a single commit on the master branch when you're satisfied (this will keep a clean history on the master branch).

But if you want to run a workflow a second time on the same commit, you can already do that without any change. Open the run, and you'll see a "Re-run all jobs" button on the right:

image

You may also re-run a single job:

image

(there's currently a single job in the workflow, but you may want to add a second one for Alpine)

PhilipHazel:
Thanks for your help. I have been experimenting in a branch called ymltest.
After 6 attempts I have got it to configure and compile PCRE2 in an Alpine
container. The main tests run OK, but the Grep test fails - presumably this
is expected because of the issue raised in #10, though I thought this had
been fixed. (Another point is that I used libc but in #10 it says Alpine
uses musl. Perhaps I should change to use that?) I have not merged the
ymltest branch, pending more work. It would be useful somehow to see the
output from the RunGrepTest script.

ltrzesniewski:
> Another point is that I used libc but in #10 it says Alpine uses musl. Perhaps I should change to use that

Yes, I think the main goal of using Alpine is to test PCRE2 with musl. @carenas suggested installing `musl-dev` instead of `libc-dev`.

> It would be useful somehow to see the output from the RunGrepTest script.

The workflow will display any output of specified the commands. I used `make check` for simplicity, but it looks like it discards the output of the commands. You may split this step into 3 separate steps which call the test scripts directly. This should display their output.

carenas:
Alpine's failure is because of the reordering of the callout output and corresponds to the following log (extracted from RunGrepTest.log) output:

```
Testing pcre2grep script callouts
--- ./testdata/grepoutputC
+++ testtrygrep
@@ -1,22 +1,22 @@
 Arg1: [T] [he ] [ ] Arg2: |T| () () (0)
+The quick brown
 Arg1: [T] [his] [s] Arg2: |T| () () (0)
 Arg1: [T] [his] [s] Arg2: |T| () () (0)
 Arg1: [T] [he ] [ ] Arg2: |T| () () (0)
 Arg1: [T] [he ] [ ] Arg2: |T| () () (0)
 Arg1: [T] [he ] [ ] Arg2: |T| () () (0)
-The quick brown
 This time it jumps and jumps and jumps.
 This line contains \E and (regex) *meta* [characters].
 The word is cat in this line
 The caterpillar sat on the mat
 The snowcat is not an animal
 Arg1: [qu] [qu]
+The quick brown
 Arg1: [ t] [ t]
 Arg1: [ l] [ l]
 Arg1: [wo] [wo]
 Arg1: [ca] [ca]
 Arg1: [sn] [sn]
-The quick brown
 This time it jumps and jumps and jumps.
 This line contains \E and (regex) *meta* [characters].
 The word is cat in this line
FAIL RunGrepTest (exit status: 1)
```

PhilipHazel:
Inserting fflush *before* as well as after a forked callout has fixed it!
Yay.

PhilipHazel:
Yes, indeed. I changed the .yml file to run each test separately, and got
the same output. I have just tried inserting a call to fflush() after a
forked callout in pcre2grep in the hope that this might "standardize" the
ordering of the output. It has changed the order on my box (Arch linux) and
the plain Ubuntu test on GitHub (which still passes), but sadly the Alpine
test still fails (though the output is different). Continuing to think....

PR #42, opened 2021-11-09
jit: avoid wraparound in stack size

carenas:
An integer overflow that is slightly more relevant in the old PCRE codebase (where it will lead to Undefined Behaviour and if able to be triggered correctly a bug that might have security implications), but still worth addressing explicitly in the new codebase IMHO.

Avoid the overflow by checking and failing the value early, and
while at it make the check clearer and document the failure mode.


zherczeg (approved):
LGTM

Issue #43, opened 2021-11-11
perl compatible matching of '\n' with '.' when (*NUL)


carenas:
perl doesn't match '\n' with '.' unless the "s" modifier is provided and regardless of what the input separator is as shown by: 

```
$ printf '\n\na\0' | perl -ne 'BEGIN { $/="\0" } /(?<=\n)(.*)$/ and print $1' | od -c | head -1
0000000    a  \0
$ printf '\n\na\0' | perl -ne 'BEGIN { $/="\0" } /(?<=\n)(.*)$/s and print $1' | od -c | head -1
0000000   \n   a  \0
```

GNU grep (that uses the old PCRE) shows a similar behaviour when using NUL as a line delimiter (-z is for NUL separated input, and also sets the line terminator of output, just like perl's "-0 -l") as  shown by GNU grep

```
$ printf '\n\na\0' | ggrep -Pzo '(?<=\n).*$' | od -c | head -1
0000000    a  \0
```

but PCRE2 does not, if the newline is not LF (or a compatible ANY or ANYCRLF) and that is actually validated by the testsuite (set 2) and IMHO makes more sense, but that will prevent grep (that has been recently updated to pcre2 on its unreleased version) to use PCRE2_NEWLINE_NUL as this change of behaviour might be considered a regression.

```
$ printf '\n\na\0' | pcre2grep -o -NNUL '(?<=\n).*$' | od -c | head -1
0000000   \n   a  \n
```

the documentation explicly says that no changes on the matching are expected when the newline definition is changed, and when newline is '\n' an equivalent "s" mode is provided through PCRE2_DOTALL making the result the same than perl for the modes that have '\n' as a valid new line delimiter, but not if CR or NUL are used, so there is at least a possibility this might be a "bug"?

FWIW, confirmed at least it is not a regression, as 8.x, while not having NUL, behaves the same when using CR and which is consistent with the behaviour observed in 10.x

carenas:
Not a bug, but a difference in the design decisions between PCRE and perl which makes the behaviour naturally incompatible.

PR #45, opened 2021-11-13
perldot: perl compatible matching of "." when using NUL or CR terminated subjects

carenas:
The first patch could be taken AS-IS, but the rest are only a POC (meant more for design discussion), since they are  probably incomplete, missing tests, documentation and will need serious refactoring.

It also has probably questionable names, and might be missing pieces, as I am not familiar with this part of the codebase and I simply mindlessly did the changes that were needed to compile, but it seems to work fine on my tests, and therefore think could do with some peer review.

Fixes: #43

### DIFF
--- a/HACKING
+++ b/HACKING
@@ -389,6 +389,7 @@ These items are all just one unit long:
 
   OP_END                 end of pattern
   OP_ANY                 match any one character other than newline
+  OP_ANY_NOTNL           match any one character other than '\n' or newline
   OP_ALLANY              match any one character, including newline
   OP_ANYBYTE             match any single code unit, even in UTF-8/16 mode
   OP_SOD                 match start of data: \A
@@ -426,6 +427,9 @@ These items are all just one unit long:
 OP_ASSERT_ACCEPT is used when (*ACCEPT) is encountered within an assertion.
 This ends the assertion, not the entire pattern match. The assertion (?!) is
 always optimized to OP_FAIL.
+
+OP_ANY_NOTNL is used for '.' when PCRE2_NO_DOTNL is set and unless
+PCRE2_DOTALL is also set as explained below.
 
 OP_ALLANY is used for '.' when PCRE2_DOTALL is set. It is also used for \C in
 non-UTF modes and in UTF-32 mode (since one code unit still equals one
--- a/doc/pcre2api.3
+++ b/doc/pcre2api.3
@@ -497,12 +497,13 @@ maximum.
 .SH NEWLINES
 .rs
 .sp
-PCRE2 supports five different conventions for indicating line breaks in
+PCRE2 supports six different conventions for indicating line breaks in
 strings: a single CR (carriage return) character, a single LF (linefeed)
-character, the two-character sequence CRLF, any of the three preceding, or any
-Unicode newline sequence. The Unicode newline sequences are the three just
-mentioned, plus the single characters VT (vertical tab, U+000B), FF (form feed,
-U+000C), NEL (next line, U+0085), LS (line separator, U+2028), and PS
+character, the two-character sequence CRLF, any of the three preceding, any
+Unicode newline sequence, or the NUL character (binary zero).
+The Unicode newline sequences are the three just mentioned, plus the
+single characters VT (vertical tab, U+000B), FF (form feed, U+000C),
+NEL (next line, U+0085), LS (line separator, U+2028), and PS
 (paragraph separator, U+2029).
 .P
 Each of the first three conventions is used by at least one operating system as
--- a/src/pcre2.h.in
+++ b/src/pcre2.h.in
@@ -143,6 +143,7 @@ D   is inspected during pcre2_dfa_match() execution
 #define PCRE2_EXTENDED_MORE       0x01000000u  /* C       */
 #define PCRE2_LITERAL             0x02000000u  /* C       */
 #define PCRE2_MATCH_INVALID_UTF   0x04000000u  /*   J M D */
+#define PCRE2_NO_DOTNL            0x08000000u  /* C       */
 
 /* An additional compile options word is available in the compile context. */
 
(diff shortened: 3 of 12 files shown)


zherczeg:
This looks very special. Is there a reason for this (other than perl)? I mean what is the typical use case?

carenas:
the "use case" came from an old GNU grep regression report[1] and the corresponding test which I came along while migrating that codebase to PCRE2.

I don't find it very useful either, and indeed prefer PCRE's behaviour but the fact that changing the underlying library and starting to use PCRE2_NEWLINE_NUL (which should help with several other issues) prevents anyone to get the perl behaviour seems like a compatibility bug worth solving IMHO, and of course, as you said it is perl behaviour.

[1] https://lists.gnu.org/archive/html/bug-grep/2015-07/msg00002.html

zherczeg:
I am sorry, but I don't really understand the bug. The pattern is `/(?<=\n\n\n).*/` and grep (2.21) does not match to `\n\n\nThis line has three blank lines above it.` If I try it in PCRE2, it matches without any problem.

zherczeg:
Just a thought: `.*` matches to an empty string, and this might confuse grep internals.

carenas:
> I am sorry, but I don't really understand the bug. The pattern is `/(?<=\n\n\n).*/` and grep (2.21) does not match to `\n\n\n\nThis line has three blank lines above it.` If I try it in PCRE2, it matches without any problem.

It should match "anything that is preceded by 3 '\n', and indeed it matches it you happen to have 4; the difference in output between perl and pcre comes from the definition of what matches '.'

perl (and grep, because it is using PCRE*_NEWLINE_LF in a weird way), will ignore that extra '\n', which is what we can't do in PCRE2 unless something like the proposed patch is added.

agree the behaviour on the bug report wasn't correct either, and indeed my suggestion[1] after "fixing" it was to consider the PCRE behaviour (with PCRE2_NEWLINE_NUL) as the correct one, but it will be technically a regression and probably a surprise to people that expect the "P" in PCRE to mean full compatibility with perl.

[1] https://git.savannah.gnu.org/cgit/grep.git/commit/?id=015d028d0598f31d5aa25e5c47dfe8872afb4e6e

zherczeg:
I know little about perl, but here is my source code:
```
$str = "\n\n\n\nThis";
if ($str =~ /(?<=\n\n\n)(.*)/) {
   print "'$1'\n";
}
```
It prints `''`. If `$str` is `"\n\n\nThis"` it prints `'This'`. This is exactly how pcre2 works.

PCRE2 has a "not empty" flag for matching, would it be possible to use that flag?

carenas:
> It prints `''`. If `$str` is `"\n\n\nThis"` it prints `'This'`. This is exactly how pcre2 works.

correct; the "empty match" is because the expression doesn't have the "s modifier" and so can't include '\n' as part of '.' so it gets an empty match that then can't multiply, returning an empty match, which I have to admit I didn't expect.

but that is not the issue discussed here, and indeed when the line delimiter is '\n' perl and PCRE are in perfect alignment, the  divergence only occurs when using something that doesn't have '\n' like '\r' or '\0' as a line delimiter.

> PCRE2 has a "not empty" flag for matching, would it be possible to use that flag?

it might work for this specific case, but still will fail when the '\n' is embedded in the middle of the subject as shown in the test cases.  any string that has '\n' won't match, regardless of what the line delimiter is (-0 means NUL for perl)

```
$ printf "a\nb" | perl -0ne 'print length($1)+1,"\n" if /(a.b)/'
$ printf "a\0b" | perl -0ne 'print length($1)+1,"\n" if /(a.b)/'
$ printf "a\nb" | pcre2grep -o -NNUL -a '(a.b)' | wc -c
       4
$ printf "a\0b" | pcre2grep -o -NNUL -a '(a.b)' | wc -c
       0
```

zherczeg:
In other words, `\n` should not match to dot, even if the newline type(s) does not include `\n`? And this behaviour is hidden behind an options flag. Well it should have no effect on jit, since it won't complie anything when not needed, but it looks like bigger impact on the interpreter when the dot is repeated. So from my side this is ok, but lets wait for @PhilipHazel 's opinion.

carenas:
> "Perl uses \n to represent the "logical" newline, where what is logical may depend on the platform in use. In MacPerl, \n always means \015. On EBCDIC platforms, \n could be \025 or \045. In DOSish perls, \n usually means \012, but when accessing a file in "text" mode, perl uses the :crlf layer that translates it to (or from) \015\012, depending on whether you're reading or writing. Unix does the same thing on ttys in canonical mode. \015\012 is commonly referred to as CRLF."

This paragraph explains IMHO clearly why in perl '.' never matches '\n'; because the definition of it is done at compile time to match what would be expected of the text files feed to it.

It also explains why -0 doesn't have the same effect on it that the equivalent option in PCRE has, and as you pointed out leads to confusion.

> I am now confused. However, given that PCRE2 has a different approach to newlines than Perl (recognizing more than one type at once, for example), I think the current behaviour makes sense. The documentation could always be improved, of course, especially if we can figure out exactly how Perl behaves, in order to describe the differences.

Agree the current approach makes sense and matches the documentation; my concern again is about perl compatibility and the fact that with the current implementation there is just no way to keep the matches consistent between PCRE and perl because of that special place that perl has for '\n'.

I was hoping with this POC to prove that the ammount of code changes required to support it (at least in *NIX) wasn't that big of a maintenance burden IMHO, and doing so will also probably allow GNU grep that has moved to use PCRE2 the opportunity to finally remove that "-Pz is highly experimental and might not do what you expect" note in their documentation with the library support of PCRE2_NEWLINE_NUL.

zherczeg:
There is another way to implement this, and that does not need any matching changes: convert dot to [^\0\n] internally during parsing. These conversions can be done by repan ( https://github.com/zherczeg/repan ) easily.

PhilipHazel:
I have just committed a couple of documentation updates (pcre2compat and pcre2syntax) that are intended to clearly explain PCRE2's behaviour. I don't think we should make any change.

PR #49, opened 2021-11-14
pcre2grep: remove JFRIEDL_DEBUG obsoleted code

carenas:


PhilipHazel:
How do you know that this code is obsolete? It is several years since I corresponded with Jeffrey Friedl, but he may well still be running tests.

carenas:
pcre_exec() doesn't exist in any 10.x version and therefore it doesn't build.

a "port" to 10.x might be possible, but it didn't seem to fit well with the surrounding code at first, and the functionality it has seems to be already available with other options (like timing report)

PhilipHazel:
Ah, I had overlooked that. I guess JF hasn't used this facility since PCRE2 came along, and as that was nearly 7 years ago, I now agree that the code is indeed obsolete. Thanks for all the work you are doing, by the way. I will look at all the PRs in due course.

Issue #54, opened 2021-11-21
allow NULL to be considered as "" if used together with a length of 0


carenas:
I usually prefer to support `ptr:NULL, length:0` for string manipulations (all `string.h` functions support it) but it looks like this is not the case for pcre2. I wouldn't mind supporting it though.

_Originally posted by @zherczeg in https://github.com/PhilipHazel/pcre2/issues/53#issuecomment-974762207_

PhilipHazel:
The original intention for PCRE2 was that NULL always gives an error, whatever the length, for both patterns and subjects, whereas an empty string is of course a valid pattern and a valid subject. If NULL with zero length is supported as an empty subject, it should also be supported as a pattern, and as a replacement string in pcre2_substitute() (and any others I've overlooked) for consistency. Is it really worth doing this work? How much would such a feature be used?

zherczeg:
In most system, don't supporting it requires effort, not supporting it :) Anyway, it is definitely not worth to spending much effort on it.

ltrzesniewski:
I brought that up because in C#, `ReadOnlySpan<char>.Empty` is a struct which contains a NULL pointer and a zero length, and PCRE2 didn't like that kind of input, which surprised me. I had to [special-case](https://github.com/ltrzesniewski/pcre-net/blob/2d4c01f625e49ef0d6d9ad7f27f5d1ff236c3340/src/PCRE.NET/Internal/InternalRegex.cs#L211-L212) this in PCRE.NET, and I thought maybe other people stumbled upon it as well.

I don't know if I'd even call that a feature, it just seemed like a small inconsistency to me. I can send a PR to support this later on if you deem it's worth the trouble.

carenas:
> The original intention for PCRE2 was that NULL always gives an error, whatever the length

I think this makes sense design wise, and the only interface that is currently missing AFAIK (other than the fixes needed so that we don't crash instead) is the use of NULL for the replacement string in pcre2_substitute, which is not documented/coded to return PCRE2_NULL_ERROR.

presume that since we will crash now instead, changing that wouldn't require changes to the API and is unlikely to get anyone surprised.

would be great if this could be tested after to avoid regressions, but I am not sure what would be the best way to do so

PhilipHazel:
[...]

PR #55, opened 2021-11-21
pcre2_string_utils: avoid segfault with strlen(NULL)

carenas:
As the first step towards allowing the use of NULL together with a length of 0 as equivalent to "" for a subject as suggested in #54, but also as an alternative fix for the failures addressed by #53.

### DIFF
--- a/doc/pcre2_substitute.3
+++ b/doc/pcre2_substitute.3
@@ -53,7 +53,7 @@ changed to the length of the new string, excluding the trailing zero that is
 automatically added.
 .P
 The subject and replacement lengths can be given as PCRE2_ZERO_TERMINATED for
-zero-terminated strings. The options are:
+zero-terminated strings. if used with a replacement string of NULL, then it is assumed to be equivalent to the behaviour expected from a replacement string of NULL and a length of 0. The options are:
 .sp
   PCRE2_ANCHORED             Match only at the first position
   PCRE2_ENDANCHORED          Pattern can match only at end of subject
--- a/src/pcre2_string_utils.c
+++ b/src/pcre2_string_utils.c
@@ -209,6 +209,7 @@ PCRE2_SIZE
 PRIV(strlen)(PCRE2_SPTR str)
 {
 PCRE2_SIZE c = 0;
+if (str == NULL) return 0;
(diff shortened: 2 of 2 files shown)


ltrzesniewski:
I don't think that's what we want, see [my comment here](https://github.com/PhilipHazel/pcre2/pull/53#issuecomment-974803818).

carenas:
> I don't think that's what we want

(NULL, PCRE2_ZERO_TERMINATED) is an special case of (NULL, 0) as they will be equivalent if you consider (or are forced to, because your language doesn't differenciate NULL) that NULL and "" are somehow representing an "empty string".

since from the discussion in #54 it would seem that there is no point on accomodating for that in the library, this is likely not needed as well, and specially because it is likely that most languages interfacing with the library would hit this already and have a workaround like the one you have in PCRE.NET, and which will need to be replicated inside the library if we were to fully implement that equivalence.

carenas:
discarded as the solution in #53 is likely a better fit and this will require implementation of #54 which seems unlikely

ltrzesniewski:
> (NULL, PCRE2_ZERO_TERMINATED) is an special case of (NULL, 0)

I disagree. Here is my interpretation:

- (NULL, PCRE2_ZERO_TERMINATED) is the equivalent of (NULL, strlen(NULL)), which is invalid.
- (NULL, 0) is a special case of (*anything*, 0)

> if you consider that NULL and "" are somehow representing an "empty string"

I don't consider that. They are definitely different.

We apparently still have a misunderstanding about that, but I don't think it's worth pursuing it further.

> it would seem that there is no point on accomodating for that in the library

Yes, let's drop this. It's not really worth the time we all already spent on this.

PhilipHazel:
After some thought, I do have some sympathy with ignoring the value of the pointer when the string length is given as 0. When I get round to looking at the code - possibly in a couple of weeks as there is no hurry and I'm on another project - I will certainly put in Carlo's crash avoidance patches. To support "length=0 means empty string" is perhaps just a matter of inserting lines like

if (length == 0) subject = (appropriate cast)"";

in the right places. I would have to see if that's OK at all code unit sizes.

PR #67, opened 2021-12-30
Auto generate unicode property tests.

zherczeg:
I have created a new test (test26) which is auto-generated by a script in `maint`. This test should be recreated after a unicode update to a new version.


zherczeg:
I wanted to generate negative tests, but I got unexpected (I could says disturbing) results.

For example `Scripts.txt` says:
`A8F2..A8F7    ; Devanagari # Lo   [6] DEVANAGARI SIGN SPACING CANDRABINDU..DEVANAGARI SIGN CANDRABINDU AVAGRAHA`

And `ScriptExtensions.txt` says:
`A8F3          ; Deva Taml # Lo       DEVANAGARI SIGN CANDRABINDU VIRAMA`

My question is: if a character is defined as Devanagari, why it is defined as a script extension for Devanagari? This is completely illogical and looks like a bug in the data files. Does anybody know the reason? If we cannot trust the Script Extension data, we need to rewrite the generator completely :(

zherczeg:
I think the script overlapping problem does not cause any practical problem at the moment, but it is good to keep in mind. The generator removes the not-working negative tests by parsing the Scripts.txt. It does not try to find another negative test though, so currently there is no negative test for "Cyrillic" script.

zherczeg:
Reworked the script to include normal scripts. This way we can check that our hand maintained datasets are correct.

PR #71, opened 2022-01-04
RunGrepTest: Fix tests 132 and 133 when srcdir is relative

jrtc27:
Notably, running the script directly from a build subdirectory will infer srcdir as .. if not otherwise set, but doesn't work for these. With this commit sh pcre2_grep_test.sh works as expected.


PhilipHazel:
Oops, yes, that is an obvious mistake. Thanks for the fix.

PR #74, opened 2022-01-06
Improve unicode property abbreviation support

zherczeg:


zherczeg:
I have added a separate patch which fully auto-generates script names and their abbreviations. Of course the order of the script names is changed, so a lot of files changed as well. It also turned out that the abbreviation of Tangsa is tnsa, not tngs.

PhilipHazel:
I have merged these patches and updated the ucptest test data for the new script numbers. I have also put a temporary note in the README just in case someone wants to update the Unicode version. The lists of scripts in the user documentation are really getting too long to maintain by hand. I propose to add a new feature to pcre2test e.g. "pcre2test -list scripts" which will give a list of supported scripts. "-list properties" might also be useful. But I'll do the new property support first.

zherczeg:
When a new script is added, is it added to the end of `Scripts.txt`? If they randomly reorder them, we could force some ordering (e.g. alphabetic).

PhilipHazel:
I do not know, but I do not think so. I am sure that Scripts.txt is
auto-generated from a big database that they must have. I am busy working
on GenerateUcd.py at the moment, for the boolean properties, so it will
make less merging problems if you hold off making other changes. It might
be an idea to make them alphabetic. It might reduce matching errors in
ucptestdata when new scripts are added.

Regards,
Philip

zherczeg:
It was just thinking loudly. It is easy to do, but will change the generated files, so the change would be big.

Issue #77, opened 2022-01-11
Heap buffer overflow


rlohning:
Hi,

I'm a developer in the Qt project which uses pcre2. I think I found an issue. Do demonstrate this, I'll use your oss-fuzz image with a changed fuzz target which mimics how Qt uses your library. I did not change anything in pcre2 itself. Should any of my steps be incorrect, please let me know.

1. Check out [my branch of oss-fuzz](https://github.com/rlohning/oss-fuzz/tree/break_pcre2).
This will clone your latest sources to the oss-fuzz image with [changes I did to the fuzz target in my fork of your repo](https://github.com/rlohning/pcre2/tree/rebuild_qt).
2. Build your fuzzer in oss-fuzz:
`python infra/helper.py build_image pcre2`
`python infra/helper.py build_fuzzers --engine libfuzzer --sanitizer address --architecture x86_64 pcre2`
3. Run this with the input file I will send to you via mail.
`python infra/helper.py reproduce pcre2 pcre2_fuzzer <input_file>`
You will see output like:

> Running: /testcase
=================================================================
==18==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6030000001a1 at pc 0x00000058a03f bp 0x7ffd1c26fab0 sp 0x7ffd1c26faa8
READ of size 1 at 0x6030000001a1 thread T0
SCARINESS: 12 (1-byte-read-heap-buffer-overflow)
    #0 0x58a03e in get_ucp /src/pcre2/src/pcre2_compile.c
    #1 0x56f0a6 in parse_regex /src/pcre2/src/pcre2_compile.c:3152:14
    #2 0x56452e in pcre2_compile_8 /src/pcre2/src/pcre2_compile.c:10147:13
    #3 0x55e3df in LLVMFuzzerTestOneInput /src/pcre2/src/pcre2_fuzzsupport.c:68:23
    #4 0x455283 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) cxa_noexception.cpp
    #5 0x440ec2 in fuzzer::RunOneTest(fuzzer::Fuzzer*, char const*, unsigned long) /src/llvm-project/compiler-rt/lib/fuzzer/FuzzerDriver.cpp:324:6
    #6 0x44671c in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) cxa_noexception.cpp
[...]

PR #78, opened 2022-01-12
Merge scriptx and bidi fields

zherczeg:
This way the scriptx data can grow from 8 bit to 11 bit, which should be enough in the foreseeable future.


PhilipHazel:
I will look at this when I've finished documenting/tidying previous work and checked out #77. There will be additional work to do in the comments and in ucptest.c.

PR #81, opened 2022-01-14
Extend unicode boolean property bitset index to 12 bit

zherczeg:


zherczeg:
12 bit should be enough for a while and we still have 4 bit free field

PR #83, opened 2022-01-26
Fix typo `with-match-limit_depth` -> `with-match-limit-depth`

adsr:
Noticed while reading pcre2.txt. Apologies if this is the wrong way to update docs.


PhilipHazel:
This is a good place for updating docs - but you can save yourself some trouble next time. The HTML docs and the .txt file are automatically built from the .1 and .3 files so there's no need to update them manually (maybe you didn't, but I thought I'd just mention this). I've merged your patch.

PR #93, opened 2022-03-28
Add OpenSSF Scorecards to impove the security posture

flyfishzy:
# What is Scorecards?
Scorecards is an [`Open Source Security Fundation`](https://openssf.org/) porject.
> Scorecards is an automated tool that assesses a number of important heuristics [("checks")](https://github.com/ossf/scorecard#scorecard-checks) associated with software security and assigns each check a score of 0-10. You can use these scores to understand specific areas to improve in order to strengthen the security posture of your project.
https://github.com/ossf/scorecard

# Troubleshooting
In order to make Scorecards works correctly,  a `Read-only Access Token` needs to be added to the project.
You can reference https://github.com/ossf/scorecard-action#authentication

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,70 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master, dev ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '36 3 * * 4'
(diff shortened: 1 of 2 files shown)


zherczeg:
What is the benefit of this? It seems it gives you numbers which you can compare between projects. I don't have good experiences with such numbers they are often confusing.

flyfishzy:
It can help improve security posture, eg. `Static Application Security Testing (SAST)`,`CI-Tests`,`Branch-Protection` and so on.
This PR add two github actions which are automatically triggered whenever there is a code commit.

PhilipHazel:
I'm pleased to note that running this check found only one issue: a grumble about a potential integer overflow in pcre2test which in practice could never happen. Nevertheless, it is always helpful to have more checking going on, so I have merged this PR. As well as fixing the reported issue, this did get me to add some other small related tidies to pcre2test.

PhilipHazel:
Spoke too soon ... there seems to be a problem with running the workflow.

flyfishzy:
It looks like the scorecards analysis run failed.
In order to make Scorecards works correctly, a Read-only Access Token needs to be added to the project.
You can reference https://github.com/ossf/scorecard-action#authentication

PhilipHazel:
I believe I have sorted this out. The action has run clean.

PR #100, opened 2022-04-14
Enable fuzzing of the JIT engine

pagabuc:
This PR enables fuzzing of the JIT engine when PCRE2 is compiled with `--enable-fuzz-support --enable-jit`. 
I have tested this patch locally and it found some crashes in the JIT compiler (fixed in 50a51cb7e67268e6ad417eb07c9de9bfea5cc55a and 4ca0530b9bdf00d2485d612ddfab8d21c72dc83a).

### DIFF
--- a/src/pcre2_fuzzsupport.c
+++ b/src/pcre2_fuzzsupport.c
@@ -148,6 +148,9 @@ for (i = 0; i < 2; i++)
 
   if (code != NULL)
     {
+#ifdef SUPPORT_JIT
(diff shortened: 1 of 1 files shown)


zherczeg:
Is this patch forcing to enable jit? I don't know anything about this system, but such features are usually enabled by command line arguments (of course the command line argument works only if the feature is enabled, otherwise it notifies the user that it is ignored).

PhilipHazel:
No, the patch is to run with JIT if enabled. I have merged it.

PR #105, opened 2022-04-20
ci: windows support

carenas:
Configure and build (using cmake with Visual Studio) in Windows and validate no regressions are introduced by running some of the tests

### DIFF
--- a/RunTest.bat
+++ b/RunTest.bat
@@ -135,9 +135,9 @@ if "%all%" == "yes" (
   set do7=yes
   set do8=yes
   set do9=yes
-  set do10=yes
+  set do10=no
   set do11=yes
-  set do12=yes
+  set do12=no
   set do13=yes
   set do14=yes
   set do15=yes
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,3 +21,21 @@ jobs:
         
       - name: Test
         run: make check
+
+  windows:
+    name: 32bit Windows
+    runs-on: windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Configure
+        run: cmake -DPCRE2_SUPPORT_JIT=ON -DPCRE2_BUILD_PCRE2_16=ON -DPCRE2_BUILD_PCRE2_32=ON -B build -A Win32
+
+      - name: Build
+        run: cmake --build build
+
+      - name: Test
+        run: |
+          cd build\Debug
+          ..\..\RunTest.bat


RunTest.bat
```diff
@@ -135,9 +135,9 @@ if "%all%" == "yes" (
   set do7=yes
   set do8=yes
   set do9=yes
-  set do10=yes
+  set do10=no
   set do11=yes
-  set do12=yes
+  set do12=no
```
zherczeg, line 140: This is my worry. If something does not work, we might not be able to do anything with it, just disable tests. CI is usually good for actively developed targets.

carenas: well, we don't have yet functionality that allow us to debug remotely broken tests, which is why the only option now is to disable them.

but I happened to already build that for windows just because I was curious on why those 2 tests were failing and you can see both the code and a run with a downloadable set of files showing the output in:

  https://github.com/carenas/pcre2/actions/runs/2199761640

there is also a tentative solution for Linux there, and of course fixes pending for the real underlying issue (which I am still puzzled about, because it will seem to be an extra empty line from pcre2test, and is consistently broken in that old OS github provides for its action VM)

PhilipHazel:
I have merged this, but like Zoltan I don't like the idea of disabling some of the tests. Let's hope this can be resolved before the next release.

ltrzesniewski:
Note that the x86 build with VS produces a [warning](https://github.com/PCRE2Project/pcre2/runs/6126362412?check_suite_focus=true#step:4:20):

> D:\a\pcre2\pcre2\src\pcre2_jit_compile.c(5889,40): warning C4018: '>=': signed/unsigned mismatch [D:\a\pcre2\pcre2\build\pcre2-16-static.vcxproj]

But this looks like a compiler issue (there's no warning on x64 for instance). I always apply this patch: https://github.com/ltrzesniewski/pcre-net/commit/3c8c0e98cd4ff02e1da7cfd7063087fb09fa315b to fix this, since I compile with warnings as errors. You didn't want to include this patch upstream, but maybe it could be considered in an `#if` which checks for VS?

I don't get the C4146 warnings in my builds though.

zherczeg:
PCRE2_UCHAR is converted to int by msvc? Strange. Signedness usually kept when promoting to larger value size.

ltrzesniewski:
I remember I didn't understand why it emitted this warning back when I had the issue the first time. The code looked correct. 😕

PhilipHazel:
I suspect it's not the PCRE2_UCHAR that is the problem. My guess is that "2" is being given the type int. Could you try change that to "2u"? (Just guessing.)

ltrzesniewski:
Changing `2` to `2u` didn't fix the issue, it's caused by the `a_pri + b_pri >= max_pri` part of the expression.

I *think* I understand what's going on though. I'll open a separate PR so we can discuss this there.

Issue #115, opened 2022-05-03
A clean and modern CMake build system


dg0yt:
This is a summary of points for discussion about improving the CMake build system.
 - Do actually export config.
   The current config is written like a find module (legacy: inspecting the system), instead of using build-time information (modern: export from the build).
 - Do not build static and shared libs in the same time.
   This looks familiar for users of autotools, but it comes with a price: It adds complexity, including additional user-facing variables. Users must select the matching target name even if they just build a single config. It requires a different name for MSVC import libs, currently breaking the unmodified pc files. ...
 - Add aliases of the exported targets, for embedding the PCRE2 build via `add_directory`.
 - Document the canonical CMake config usage.
 - Add a post-install usage test to CI.
 - Follow widely used CMake style.
   `lower_case(commands)`. Don't repeat condition in `else()` and `endif()`. Less whitespace.

References to previous discussion:
 - https://github.com/PCRE2Project/pcre2/pull/113
 - https://github.com/PCRE2Project/pcre2/commit/2410fbe3869cab403f02b94caa9ab37ee9f5854b#commitcomment-68176307
 - https://lists.exim.org/lurker/message/20210701.070030.8dbb0b5b.en.html

Who uses the CMake build system?
 - [vcpkg](https://github.com/microsoft/vcpkg/tree/master/ports/pcre2/). Apart from avoiding slow autotools builds on Windows, the default configurations build for release and debug, for the benefit of Visual Studio users.
- (I can add more here base on comments.)

PhilipHazel:
Thank you for picking this up. This project is entirely reliant on the CMake community for the maintenance of the CMake files, for which I'm grateful. (Come to think of it, the Autotools files were also a user contribution - I'm pretty ignorant about both build systems.) I'm happy to go along with whatever comes out of this discussion. Note that there is also some information about previous changes in the PCRE2 ChangeLog file, release 10.38 comment #2.

dg0yt:
> Note that there is also some information about previous changes in the PCRE2 ChangeLog file, release 10.38 comment #_2.

https://github.com/PCRE2Project/pcre2/blob/a334ea2a341cc785d8693e12979fdf6661b0698e/ChangeLog#L196-L219

Let's CC @jwsblokland and hear his POV on simultaneous static + dynamic.

GregThain:
Thanks again, @dg0yt.  I don't claim to be a CMake expert, but I'm willing to help out here.  Some other suggestions for improvements:

[...]

PR #116, opened 2022-05-06
Include specific .pdb files only for chosen char size libs when shared

wrowe:
This resolves an error that the listed .pdb files won't exist and make install fails
if 8, 16 and 32 bit libraries weren't all built.

Signed-off-by: William A Rowe Jr <wrowe@vmware.com>


PhilipHazel:
I will merge this once the checks have run, BUT: Please take a look at #115 which is a discussion about a major revision of the CMake configuration.

PR #134, opened 2022-07-10
Provide pcre2_match_data_create_with_frames() to avoid start stack frames.

ylavic:
pcre2_match() uses START_FRAMES_SIZE (20KiB) of stack space for the start
(initial) frames, which may be too large for systems with small per-thread
stacks (e.g. 128KiB by default on Alpine Linux).

Provide pcre2_match_data_create_with_frames() to allow for the user to
create match_data containing context/heap allocated start frames, which
will be used by pcre2_match() if available instead of the stack frames.

* src/pcre2.h.generic, src/pcre2.h.in:
  Declare pcre2_match_data_create_with_frames().

* src/pcre2_intmodedep.h:
  Add a pointer to start_frames to struct pcre2_real_match_data, will be
  NULL if stack frames are to be used.

* src/pcre2_match_data.c:
  Provide a common match_data_create() helper to be called by all the
  pcre2_match_data_create*() functions without duplicating code.
  When called from pcre2_match_data_create_with_frames(), match_data_create()
  will allocate the start frames in (at the end of) the ovector and point
  match_data->start_frames to that.

* src/pcre2_match.c:
  Split pcre2_match() in two match_start() and match_start_on_stack()
  helpers, the former doing the usual work given some start frames, the
  latter calling the former with start frames reserved on the stack.
  pcre2_match can then call the one or the other depending on whether
  match_data->start_frames exist or not, playing some noinline/indirection
  games to do the stack allocation from match_start_on_stack() only.

### DIFF
--- a/src/pcre2.h.generic
+++ b/src/pcre2.h.generic
@@ -673,6 +673,8 @@ PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
 PCRE2_EXP_DECL pcre2_match_data PCRE2_CALL_CONVENTION \
   *pcre2_match_data_create(uint32_t, pcre2_general_context *); \
 PCRE2_EXP_DECL pcre2_match_data PCRE2_CALL_CONVENTION \
+  *pcre2_match_data_create_with_frames(uint32_t, pcre2_general_context *); \
+PCRE2_EXP_DECL pcre2_match_data PCRE2_CALL_CONVENTION \
   *pcre2_match_data_create_from_pattern(const pcre2_code *, \
     pcre2_general_context *); \
 PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
@@ -868,6 +870,7 @@ pcre2_compile are called by application code. */
 #define pcre2_match_context_create            PCRE2_SUFFIX(pcre2_match_context_create_)
 #define pcre2_match_context_free              PCRE2_SUFFIX(pcre2_match_context_free_)
 #define pcre2_match_data_create               PCRE2_SUFFIX(pcre2_match_data_create_)
+#define pcre2_match_data_create_with_frames   PCRE2_SUFFIX(pcre2_match_data_create_with_frames_)
 #define pcre2_match_data_create_from_pattern  PCRE2_SUFFIX(pcre2_match_data_create_from_pattern_)
 #define pcre2_match_data_free                 PCRE2_SUFFIX(pcre2_match_data_free_)
 #define pcre2_pattern_convert                 PCRE2_SUFFIX(pcre2_pattern_convert_)
--- a/src/pcre2.h.in
+++ b/src/pcre2.h.in
@@ -673,6 +673,8 @@ PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
 PCRE2_EXP_DECL pcre2_match_data PCRE2_CALL_CONVENTION \
   *pcre2_match_data_create(uint32_t, pcre2_general_context *); \
 PCRE2_EXP_DECL pcre2_match_data PCRE2_CALL_CONVENTION \
+  *pcre2_match_data_create_with_frames(uint32_t, pcre2_general_context *); \
+PCRE2_EXP_DECL pcre2_match_data PCRE2_CALL_CONVENTION \
   *pcre2_match_data_create_from_pattern(const pcre2_code *, \
     pcre2_general_context *); \
 PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
@@ -868,6 +870,7 @@ pcre2_compile are called by application code. */
 #define pcre2_match_context_create            PCRE2_SUFFIX(pcre2_match_context_create_)
 #define pcre2_match_context_free              PCRE2_SUFFIX(pcre2_match_context_free_)
 #define pcre2_match_data_create               PCRE2_SUFFIX(pcre2_match_data_create_)
+#define pcre2_match_data_create_with_frames   PCRE2_SUFFIX(pcre2_match_data_create_with_frames_)
 #define pcre2_match_data_create_from_pattern  PCRE2_SUFFIX(pcre2_match_data_create_from_pattern_)
 #define pcre2_match_data_free                 PCRE2_SUFFIX(pcre2_match_data_free_)
 #define pcre2_pattern_convert                 PCRE2_SUFFIX(pcre2_pattern_convert_)
--- a/src/pcre2_intmodedep.h
+++ b/src/pcre2_intmodedep.h
@@ -615,6 +615,8 @@ here.) */
 #undef  LOOKBEHIND_MAX
 #define LOOKBEHIND_MAX UINT16_MAX
 
+struct heapframe; /* see below */
+
 typedef struct pcre2_real_code {
   pcre2_memctl memctl;            /* Memory control fields */
   const uint8_t *tables;          /* The character tables */
@@ -661,6 +663,7 @@ typedef struct pcre2_real_match_data {
   uint8_t          flags;         /* Various flags */
   uint16_t         oveccount;     /* Number of pairs */
   int              rc;            /* The return code from the match */
+  struct heapframe *start_frames; /* Initial heap frames (NULL for stack) */
   PCRE2_SIZE       ovector[131072]; /* Must be last in the structure */
 } pcre2_real_match_data;
 
--- a/src/pcre2_match.c
+++ b/src/pcre2_match.c
@@ -6268,35 +6268,10 @@ switch (Freturn_id)
 #undef LBL
 }
 
-
-/*************************************************
-*           Match a Regular Expression           *
-*************************************************/
-
-/* This function applies a compiled pattern to a subject string and picks out
-portions of the string if it matches. Two elements in the vector are set for
-each substring: the offsets to the start and end of the substring.
-
-Arguments:
-  code            points to the compiled expression
-  subject         points to the subject string
-  length          length of subject string (may contain binary zeros)
-  start_offset    where to start in the subject string
-  options         option bits
-  match_data      points to a match_data block
-  mcontext        points a PCRE2 context
-
-Returns:          > 0 => success; value is the number of ovector pairs filled
-                  = 0 => success, but ovector is not big enough
-                  = -1 => failed to match (PCRE2_ERROR_NOMATCH)
-                  = -2 => partial match (PCRE2_ERROR_PARTIAL)
-                  < -2 => some kind of unexpected problem
-*/
-
-PCRE2_EXP_DEFN int PCRE2_CALL_CONVENTION
-pcre2_match(const pcre2_code *code, PCRE2_SPTR subject, PCRE2_SIZE length,
+static int
+match_start(const pcre2_code *code, PCRE2_SPTR subject, PCRE2_SIZE length,
   PCRE2_SIZE start_offset, uint32_t options, pcre2_match_data *match_data,
-  pcre2_match_context *mcontext)
+  pcre2_match_context *mcontext, heapframe *start_frames)
 {
 int rc;
 int was_zero_terminated = 0;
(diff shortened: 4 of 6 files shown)


ylavic:
Supposedly `match_data` can be reused for multiple `pcre2_match()` calls (e.g. some `thread_local` pointer), so using `pcre2_match_data_create_with_frames()` thus allocating >20KiB bytes of data does necessarily happen before each `pcre2_match()`.

chemodax:
I would suggest another solution: make START_FRAMES_SIZE configurable at compile time. In this case it can be tweaked on platforms with small stack size.

ylavic:
That would make it a distro setting for most users/projects (and distros are likely to keep the current START_FRAMES_SIZE value anyway), so pcre2_match() is still going to consume 20K of stack for most users.
The proposed changes allow projects using pcre2 and that don't need/want this stack space to opt-out, with any libpcre2 (future?) version.

PhilipHazel:
I too had thought about making START_FRAMES_SIZE configurable. It would be a much simpler patch, which I could easily do. Would than meet your need? Oh, I see you have just commented as I write. Of course, BOTH strategies could be implemented; they are not in conflict.

ylavic:
Note that the changes to [src/pcre2test.c](https://github.com/PCRE2Project/pcre2/pull/134/files#diff-8e816db40d244630db9665260092b29678ae73d55cefc6207d8fce92f3330a00) to make it use `pcre2_match_data_create_with_frames()` are possibly not needed/suitable, I used this for testing with `make check` but it might or not be the suitable default for the `pcre2test` binary..

ylavic:
> Oh, I see you have just commented as I write. Of course, BOTH strategies could be implemented; they are not in conflict.

Thanks for considering it. Yeah the main issue with compile time setting is that I (as an OOS developper) can do little about it. For instance the Apache httpd server is going to be built and run against the distro libs most of the time, reusing match_data there with no stack overhead/overflow would be great.

PhilipHazel:
I've started to look at your code. I don't (yet?) understand the fancy games with the inlining options. Can it not be done more simply like this:

* If pcre2_match() sees that the startframe pointer in the match data is NULL, it calls pcre2_start_use_stack(). This function allocates the frames on the stack, then calls pcre2_match_start(), passing the stack frames.

* If pcre2_match() finds the startframe pointer not NULL, it calls pcre2_match_start() itself, passing the frames that are on the heap.

The test could be done after testing for JIT, thereby avoiding stack usage when JIT is called.

Am I missing something? 

As for pcre2test, there will have to be some way of testing this code, but I haven't looked in detail yet. No doubt some new option will be needed.

PhilipHazel:
OK, I now understand about the inline stuff. I have been pondering how to solve your problem in a simpler manner. One idea is to abandon the use of the on-stack initial frames and always use the heap. This would mean an extra malloc/free cost for every call of pcre2_match(). I come from an era where allocating and freeing memory was expensive, but I've just done some Googling, and it seems that in modern operating systems this is no longer the case. I will do some timing tests to see if there is a detectable effect for pcre2_match. 

A slightly more complicated solution is not to modify the match_data block, but instead to add an option to pcre2_match(). This would add the additional malloc/free only to the times the option is set, but involves all the complication of indirect function calls, though it saves having to change match_data_create() with its associated pcre2test complications.

chemodax:
@PhilipHazel On modern OS heap alloc is really fast if malloc and free performed from the same thread: heap allocate have some kind of internal per-thread or per-processor free memblock list.

PhilipHazel:
However, a quick test on my Arch Linux system, using pcre2test's crude timing feature, shows a consistent 20% increase in CPU time when every call to pcre2_match() does a malloc/free. This was using the testdata/testinput1 sequence of tests, after a quick hack to disable the use of the stack-based backtracking frames (there's already code to do this if the stack vector is too small for at least 10 frames - I just made it unconditional). More thought/experiment needed (but not today).

chemodax:
@PhilipHazel Could you please share patch you are testing -- I'm interested to test it on Windows.

zherczeg:
This topic is strange for me. Modern OS-es use page based allocation, and only allocate memory pages on memory write. If you allocate 1M data with malloc, but never writes anything to it, no memory is allocated by the OS for that memory region. Same with stack. If your max stack is 100 MByte, but your application uses 2K only, your OS allocates a single 4K page for it. I don'tget the 128K limit at all, maybe it is a misconfiguration?

PhilipHazel:
I realized that my simple patch was in fact too simple; I will try a different one later today and report back.

ltrzesniewski:
I've noticed a quite large perf improvement on Windows after I removed a malloc/free pair I was doing for each match. I don't have the numbers though, but I remember the difference was significant.

PhilipHazel:
Here is the patch I have just tested:

--- a/src/pcre2_match.c
+++ b/src/pcre2_match.c
@@ -6827,14 +6827,21 @@ large. Ensure that there are at least 10 available frames by getting an initial
 vector on the heap if necessary, except when the heap limit prevents this. Get
 fewer if possible. (The heap limit is in kibibytes.) */
 
+#ifdef CUTOUT
 if (frame_size <= START_FRAMES_SIZE/10)
   {
   mb->match_frames = mb->stack_frames;   /* Initial frame vector on the stack */
   mb->frame_vector_size = ((START_FRAMES_SIZE/frame_size) * frame_size);
   }
 else
+#endif
+
   {
   mb->frame_vector_size = frame_size * 10;
+  
+  if (mb->frame_vector_size < START_FRAMES_SIZE)
+    mb->frame_vector_size = ((START_FRAMES_SIZE/frame_size) * frame_size); 
+ 
   if ((mb->frame_vector_size / 1024) > mb->heap_limit)
     {
     if (frame_size > mb->heap_limit * 1024) return PCRE2_ERROR_HEAPLIMIT;

Running pcre2test on testdata/testinput1 with this patch applied does 3787 allocs/frees (according to valgrind) whereas without it there are 1346. However, using pcre2test's relatively crude timing options showed a slowdown of only around 2%. This is on Arch Linux. On some of the shorter test files the difference was much greater, but this might be rounding type errors. I'm not sure quite what to make of this -- I await the information from Windows tests with interest.

chemodax:
On Windows:
Baseline:
> pcre2test.exe -8 -TM testinput1
Total match time 7.3721 milliseconds

With patch:
> pcre2test.exe -8 -TM testinput1
Total match time 7.5713 milliseconds

I.e. with patch it takes 20% more time.

PhilipHazel:
Either I'm not thinking straight, or your figures are a bit off: Surely a difference of 0.2 in 7.37 is around 2%, not 20%? (2.7%, I think). This is similar to what I see on Linux. What do you see for tests 4 and 5?

ylavic:
Replacing the current large on-stack initial buffer with a full malloc/free implementation is always going to be costly IMO, even if a microbenchmark shows ~2% loss only it might be worse for a real application with more concurrency/pressure on the system allocator.
In Apache httpd [for instance](https://github.com/apache/httpd/blob/f589ecaa68cfe3be8694042035efc8343715e25c/server/util_pcre.c#L277), a custom pcre2_general_context (with custom malloc/free functions) is used to redirect to a per-thread memory pool/cache that efficiently reuses memory chunks between the calls to pcre2_match(). So malloc() is going to be called a few times (for the first/largest regexes run by the thread), and free() almost never (unless the total exceeds some configurable limit per cache/pool).

With this kind of caching mechanism there is negligible overhead in switching to heap memory allocation, but since it's not something that most PCRE users have (supposedly), keeping the current stack reserve by default sounds wise.
That's why I proposed an opt-in with pcre2_match_data_create_with_frames(), but a new option to pcre2_match() could work too (still this makes sense only if stack_frames_vector is reserved in match_start_on_stack() only, when the option is not set).

Regarding JIT, it seems that pcre2_jit_match() has its own jit_stack mechanism, using some context's jit_callback_data instead of the stack_frames_vector, so the stack_frames_vector is not needed/used there (nor would be the match_data->start_frames from this PR then), right?
If that's the case then a new pcre2_match() option looks more flexible indeed, and match_start could skip the allocation until after (and if not) `use_jit`.

zherczeg:
Hm, allocating a stack space by both the interpreter and JIT could be a waste, that is a good point. Currently JIT allocates a 32 KB stack space when the custom JIT stack data is not provided (using a non-inlined match with stack function call). However, if the interpreter already allocates that space, JIT could also use it.

It could be interesting to use the JIT stack by the interpreter as well. It allocates virtual memory, but it requires OS support, so it does not work on some exotic environments, Probably not worth it.

PhilipHazel:
I have thought of a new Cunning Plan which I will try out, possibly later today. It is as follows: Add a pointer and a length to the match_data structure, but initialize the pointer as NULL. Then pcre2_match(), when it sees NULL, calls malloc. The length is needed because pcre2_match() may extend the block if it needs to. But do NOT free. Then, if the caller uses the same match_data block for multiple pcre2_match() calls, which I suspect many programs do, the block is re-used. It is eventually freed when pcre2_match_data_free() is called. The malloc() will happen only when JIT is not being called. This has the advantage of being simple, needing no action on the part of callers, and no need to explain these internals in the documentation.

ltrzesniewski:
> if the caller uses the same match_data block for multiple pcre2_match() calls, which I suspect many programs do

That may be true for C programs, but I suppose some bindings to other languages can't afford that luxury, since they don't know what their users do.

For instance, I create a match data block per match in my C# binding library (in its idiomatic API, I expose a more optimized API as well, but it imposes additional constraints on the user).

> no need to explain these internals in the documentation

Due to the above, I think that documenting this behavior could be worthwhile for *libraries* built on top of PCRE2.

ylavic:
It seems that the only place where heapframes_size should be a multiple of frame_size is in the match() function, so a local heapframes_size variable there can do (like in the proposed patch)?

PhilipHazel:
Yes, I realized that too. I'm working on it.

PhilipHazel:
I've pushed a patch to the heapframe branch, which also does a minor code tidy to the heap limit - do the *1024 once only.

Do you think that functions pcre2_match_data_{add,free}_frames() would be useful?

chemodax:
> Either I'm not thinking straight, or your figures are a bit off: Surely a difference of 0.2 in 7.37 is around 2%, not 20%? (2.7%, I think). This is similar to what I see on Linux. What do you see for tests 4 and 5?

Yes, you are right: it's 2%, not 20%.

ylavic:
(sorry for the delay)

> I've pushed a patch to the heapframe branch, which also does a minor code tidy to the heap limit - do the *1024 once only.

My testing/debugging worked perfectly, +1 from me (FWIW).

> Do you think that functions pcre2_match_data_{add,free}_frames() would be useful?

I think that it would be a good API (possibly with pcre2_match_data_frames_size()) for maintaining the match_data with the default memctl (system's malloc/free).
Honestly I won't use them for now because a custom memctl better fits my needs so far.

Anyway thanks a lot for the changes and your time, great improvements IMO.

PhilipHazel:
Thank you for testing. I have realized that work is needed in pcre2test to get the code that finds a minimum heap limit working with the new arrangements. Once I have done that, and updated the documentation, I will merge this branch. 

I have also realized that the pcre2_dfa_match() function also uses the same stack trick and could be updated in the same way. However, I do not propose to spend time doing that work unless somebody asks for it.

PhilipHazel:
The heapframe branch is now merged into master, so I am going to close this pull request.

chemodax:
@PhilipHazel As far I understand stack_frames optimization is completely removed. Is it correct? I think this will introduce performance regression for platforms with enough stack, because of extensive malloc()/free() usage for match data.

Is it possible to bring back stack_frames optimization with configurable size and smaller default? I think it will be safe for most users, but still allow use match_data without malloc() in most cases.

What do you think?

PhilipHazel:
Yes, the stack frames usage has gone, but there is a new optimization: the heap block is remembered in the match_data block. Note that this change will not affect JIT usage in any way, except that an unused block is no longer allocated on the stack. I would hope that applications that really, really care about performance would use JIT. 

For non-JIT uses of pcre2_match() there will be winners and losers. If a caller uses the same match_data block for multiple matches, there will either be no noticeable difference (one extra malloc/free) if the block never has to be extended, or the code will run faster if lots of the matches need more than the minimal amount, because the extended block is remembered. The pcre2test program, when run on the testinput1 file, runs substantially faster because of this. The losers are programs that use a new match_data block for each match. Of course, the importance depends on how much matching such programs do and how much performance matters to them. There is always the possibility of custom memory allocation.

I do not like the idea of re-introducing the stack usage. As was pointed out before, many applications want to use whatever the system installs, so making it a build-time option doesn't help them, and making it a run-time option is messy and involves compiler-specific hacks.

PR #136, opened 2022-07-12
Bazel and bazel central registry support

zaucy:
I mainly use [bazel](https://bazel.build/) as my build system and I was wondering if pcre2 would be willing to accept these files upstream to add pcre2 to the [bazel central registry](https://github.com/bazelbuild/bazel-central-registry). I _can_ write patches for the repository and submit it that way, but it is preferred to have them in the repository.

At the moment this is a minimal configuration that successfully runs the `pcre2demo`. e.g.

```sh
bazel run //:pcre2demo -- -g "\\w+" "hello world"
```
```
Match succeeded at offset 0
 0: hello
No named substrings

Match succeeded again at offset 6
 0: world
No named substrings
```

If this was accepted and submited to the bazel central registry then bazel users would only have to add the following to their `MODULE.bazel`.

```python
bazel_dep(name = "pcre2", version = "10.40")
```

And then any target that uses pcre2 would simply be added to their `deps` list.

```python
cc_library(
    name = "my_cool_library_that_uses_pcre2",
    # ...
    deps = [
        # ...
        "@pcre2",
    ],
)
```

I'm happy to add tests that cover bazel and make any other modifications you require.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,3 @@
+common --experimental_enable_bzlmod
+build --incompatible_enable_cc_toolchain_resolution
+build --incompatible_strict_action_env
--- a/.gitignore
+++ b/.gitignore
@@ -75,4 +75,6 @@ src/pcre2.h
 src/pcre2_chartables.c
 src/stamp-h1
 
+/bazel-*
+
 # End
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -1,0 +1,72 @@
+load("@rules_cc//cc:defs.bzl", "cc_library", "cc_test")
+load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
+
+copy_file(
+    name = "config_h_generic",
+    src = "src/config.h.generic",
+    out = "src/config.h",
+)
+
+copy_file(
+    name = "pcre2_h_generic",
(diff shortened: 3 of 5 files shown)


PhilipHazel:
Am I right in thinking this just involves adding some files to the GitHub repo (and a minor tweak to .gitignore)? Or would it be necessary also to include these files in the release tarballs?

zaucy:
Just the GitHub repo is fine. Unless you'd prefer one of the tarball over the repository archive to be in the bazel central registry. e.g. https://github.com/bazelbuild/bazel-central-registry/blob/main/modules/zlib/1.2.11/source.json#L8

PhilipHazel:
OK, I have merged this pull request but it's up to you or any other Bazel user to do any necessary maintenance. I note that MODULE.bazel contains the PCRE2 version number. Will that have to be updated for each release?

zaucy:
Yeah the version will have to be updated each release. Is there a script I can modify to make that happen? `PrepareRelease`?

PhilipHazel:
PrepareRelease is probably the right place to do something like that. Unfortunately, it does not at the moment make any use of the release number, so you would have to write code to pick that out of configure.ac. PrepareRelease already makes use of Perl, so adding a short in-line Perl script would be fine.

PR #165, opened 2022-12-01
jit: fix pcre2_jit_free_unused_memory() if sljit not using allocator

carenas:
sljit allows building without an internal allocator, but instead using an external one.

make sure to only invoke the corresponding sljit call if an internal allocator is in use (the default and as coded in pcre integration) to avoid problems if the code is changed to use an external allocator instead.


carenas:
FWIW the 32bit Windows failure is unrelated and independently fixed by #166

zherczeg (approved):
LGTM

PR #166, opened 2022-12-01
ci: avoid pulling 64bit libraries from strawberry perl

carenas:
Workaround for Windows builds 20221120.1[1]

Regardless of cmake version there is no point pulling any extra libraries

[1] https://github.com/actions/runner-images/issues/6627


zherczeg (approved):
LGTM

PR #168, opened 2022-12-03
jit: correctly implement utf_continue() for utf16

carenas:
Resulting in a constant false result in aarch64 when using SIMD


zherczeg (approved):
LGTM. Nice catch

PR #170, opened 2022-12-07
ci: add macOS universal build

carenas:
using cmake (not recommended) so a naive way to make an universal build could be achieved, and which has the main purpose of building the arm64 code to catch any issues there as early as possible.


ltrzesniewski:
Note that the macOS VM runs on x86_64, so arm64 code will be compiled but won't be tested with this change.

Issue #173, opened 2022-12-08
10.41 breaks compilation of MariaDB and other consumers


thesamesam:
MariaDB fails to compile with libpcre2-10.41 like so:
```
FAILED: client/CMakeFiles/mariadb-test.dir/mysqltest.cc.o 
[...]

PR #174, opened 2022-12-08
Tried to fix issue #173 (using pcre2posix.h in MariaDB)

cbouc:
#173 Added default (empty) definition for PCRE2_CALL_CONVENTION in prec2posix.h just like  that in prec2.h


src/pcre2posix.h
```diff
@@ -116,6 +116,20 @@ typedef struct {
...
+void __cdecl function(....)
+
+might be needed. In order so make this easy, all the exported functions have
+PCRE2_CALL_CONVENTION just before their names. It is rarely needed; if not
+set, we ensure here that it has no effect. */
+
+#ifndef PCRE2_CALL_CONVENTION
+#define PCRE2_CALL_CONVENTION
+#endif
+
```
carenas, line 132: the equivalent code should be then removed from pcre2posix.c

carenas (changes requested):

PhilipHazel:
Unfortunately, this issue was not caught by the tests because pcre2test includes both pcre2.h and pcre2posix.h. I will do two things: I will move the default definition (and its comment) from pcre2posix.c to pcre2posix.h and I will write a little test program that just includes pcre2posix.h (without pcre2.h) and add this test to the test suite.

thesamesam:
@PhilipHazel Once a test is added, would you consider a new release for this? Otherwise all distributors of pcre2 are likely to have to hit the problem and then backport the fix after hopefully stumbling upon the bug I filed.

PhilipHazel:
Fixed by 16802f39. I will put out a new release fairly soon.

PR #177, opened 2022-12-25
Fix integer overflow undefined behaviour

nielsdos:
Even though delta's type is PCRE2_SIZE, the computation for delta uses all integers at the right hand side.
This means that there is a potential integer overflow. The if below then checks whether a computation equivalent to delta is larger than INT_MAX: the overflow check.
However, since integer overflow is undefined behaviour, the compiler may assume it never happens. Therefore, the overflow check can be assumed to always be false even though there is casting because according to the compiler if an overflow doesn't happen for ints, it surely does not happen for INT64_OR_DOUBLE...

I found this issue using the stack static analysis tool. I verified that the compiler actually optimizes away the check by looking at the IR.

Fix it by computing delta with the casts first, and use that same delta in the if.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -7117,10 +7117,8 @@ for (;; pptr++)
 
         if (lengthptr != NULL)
           {
-          PCRE2_SIZE delta = replicate*(1 + LINK_SIZE);
-          if ((INT64_OR_DOUBLE)replicate*
-                (INT64_OR_DOUBLE)(1 + LINK_SIZE) >
-                  (INT64_OR_DOUBLE)INT_MAX ||
+          INT64_OR_DOUBLE delta = (INT64_OR_DOUBLE)replicate*(INT64_OR_DOUBLE)(1 + LINK_SIZE);
+          if (delta > (INT64_OR_DOUBLE)INT_MAX ||
               OFLOW_MAX - *lengthptr < delta)
             {
             *errorcodeptr = ERR20;


carenas:
LINK_SIZE is a very tiny integer (usually 2), so the check (as mentioned in the comment above) is just being extra paranoid.

I suspect though, that if your compiler is optimizing away that check, it might be because of a bug with it, or you should be configuring your pcre2 to use double (as a 64bit int) instead.

src/pcre2_compile.c
```diff
@@ -7117,10 +7117,8 @@ for (;; pptr++)
 
         if (lengthptr != NULL)
           {
-          PCRE2_SIZE delta = replicate*(1 + LINK_SIZE);
-          if ((INT64_OR_DOUBLE)replicate*
-                (INT64_OR_DOUBLE)(1 + LINK_SIZE) >
-                  (INT64_OR_DOUBLE)INT_MAX ||
+          INT64_OR_DOUBLE delta = (INT64_OR_DOUBLE)replicate*(INT64_OR_DOUBLE)(1 + LINK_SIZE);
```
carenas, line 7120: if a broken int64 cast is being used here, this is likely to make the problem worst by potentially halving the width of delta.

it migth be better IMHO to only cast "replicate" to ensure the multiplication is not being done with plain ints

carenas (changes requested):

src/pcre2_compile.c
carenas: a maybe cleaner way to approach this sort of check pushed (not tested) in the [following branch](https://github.com/carenas/pcre2/tree/intbug).

note that similar logic is used in several other places, so not sure why the static analyzer used didn't pick on those.

nielsdos: Your approach is indeed cleaner.
System details: Linux 6.1.3, Clang 14.0.6

carenas: could you confirm the check is indeed getting removed?, still can't reproduce that with the same compiler as shown by the following simplified code:

```
$ cat i.c
#include <stdio.h>
#include <limits.h>
#include <stdlib.h>

int main(int argc, char *argv[])
{
	int a = atoi(argv[1]);
	
	size_t m;
	       
	m = a * 2;
	if ((long)a * 2 > INT_MAX)
		printf("OVERFLOW\n");
	printf("%zu\n", m);
	return 0;
}
$ clang --version
Debian clang version 14.0.6
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
$ clang -O3 -o i i.c
$ ./i 2147483647
OVERFLOW
18446744073709551614
```

PhilipHazel:
What is the status of this, please? Should I implement the one-line patch in the conversation above?

src/pcre2_compile.c
nielsdos: That example gives the same result on my system, so this PR was probably a mistake.

carenas:
This PR could be closed, but I think we should do some related improvements anyway, as I still think that there might be some truth to the original static analyzer report

PhilipHazel:
Can we close this now?

PhilipHazel:
OK, I' closing this.

PR #182, opened 2022-12-29
correctly set heap_limit in KiB

carenas:
Noticed while doing sanitizer runs, as the multiplication will overflow an uint32_t, which for the default would result in an even larger value than expected.


carenas:
this is incorrect and was indeed part of the change

Issue #183, opened 2022-12-29
setting heap_limit could be truncated since PCRE2 10.41


carenas:
After d90fb23 (Refactor match_data() to always use the heap instead of having an initial frames vector on the stack..., 2022-07-27) the value for `mb->heap_limit` (using a PCRE2_SIZE) is set to the number of bytes, instead of KB (as used everywhere else).

The problem with doing that is that the variable that holds this in the match/expression context is only 32bit wide and therefore setting the value itself requires multiplying that value by 1024 and that can overflow.

Changing the code to widen the multiplication itself helps when PCRE2_SIZE is 64bit but it won't in 32bit systems where PCRE2_SIZE isn't just wide enough:

```
diff --git a/src/pcre2_match.c b/src/pcre2_match.c
index 168b9fa..7e65985 100644
--- a/src/pcre2_match.c
+++ b/src/pcre2_match.c
@@ -6816,7 +6816,7 @@ frame_size = (offsetof(heapframe, ovector) +
 smaller. */
 
 mb->heap_limit = ((mcontext->heap_limit < re->limit_heap)?
-  mcontext->heap_limit : re->limit_heap) * 1024;
+  mcontext->heap_limit : re->limit_heap) * 1024ul;
 
 mb->match_limit = (mcontext->match_limit < re->limit_match)?
   mcontext->match_limit : re->limit_match;
```

Additionally, since the default for heap_limit is set to 20000000KB, it is already larger (when converted into bytes) than what a 32bit PCRE2_SIZE can hold, and while that specific issue was already present in previous code, meaning that the limit was practically unreachable, it was at least representable and useful for evaluation.

PR #184, opened 2022-12-30
avoid LIMIT_HEAP truncation since d90fb23 (included in PCRE2 10.41)

carenas:
Avoid the overflow and use instead an uint32_t variable to hold the heap_limit (in KiB)

Fixes: d90fb23 (Refactor match_data() to always use the heap instead of having an initial frames vector on the stack..., 2022-07-27)
Closes: #183


src/pcre2_intmodedep.h
```diff
@@ -858,7 +858,7 @@ doing traditional NFA matching (pcre2_match() and friends). */
 
 typedef struct match_block {
   pcre2_memctl memctl;            /* For general use */
-  PCRE2_SIZE heap_limit;          /* As it says */
+  uint32_t heap_limit;            /* As it says */
```
zherczeg, line 861: PCRE2_SIZE is 64 bit long on 64 bit systems, isn't it?

carenas: correct, but the heap_limit value can only be lower than UINT_MAX even if it can represent memory sizes larger than 4GB in 64-bit systems because it uses KB for units.

PR #186, opened 2023-01-08
add PCRE2_ASCII (RFC)

carenas:
A prototype (but funtionally implementing the basic expression '\d' of the suggested feature), of ways to disable the expansion of \d even when PCRE2_UCP (or its equivalent verb) might be used.

Implemented in a way that minimizes code change so it could be reviewed more easily, and with a bare bones documentation and no tests, which will obviously be corrected for any (non draft) version.

### DIFF
--- a/doc/pcre2_compile.3
+++ b/doc/pcre2_compile.3
@@ -44,6 +44,7 @@ The option bits are:
   PCRE2_ALT_BSUX           Alternative handling of \eu, \eU, and \ex
   PCRE2_ALT_CIRCUMFLEX     Alternative handling of ^ in multiline mode
   PCRE2_ALT_VERBNAMES      Process backslashes in verb names
+  PCRE2_ASCII              Prefer ASCII in conflicting UTF classes
   PCRE2_AUTO_CALLOUT       Compile automatic callouts
   PCRE2_CASELESS           Do caseless matching
   PCRE2_DOLLAR_ENDONLY     $ not to match newline at end
--- a/doc/pcre2api.3
+++ b/doc/pcre2api.3
@@ -1447,6 +1447,12 @@ or PCRE2_EXTENDED_MORE option is set with PCRE2_ALT_VERBNAMES, unescaped
 whitespace in verb names is skipped and #-comments are recognized, exactly as
 in the rest of the pattern.
 .sp
+  PCRE2_ASCII
+.sp
+When PCRE2_UTF and PCRE2_UCP are both being used, some classes are changed in
+ways that conflict between UTF and ASCII characters. This option can be set
+to restrict \ed to only match the non UTF digits.
+.sp
   PCRE2_AUTO_CALLOUT
 .sp
 If this bit is set, \fBpcre2_compile()\fP automatically inserts callout items,
--- a/doc/pcre2pattern.3
+++ b/doc/pcre2pattern.3
@@ -73,10 +73,11 @@ appearance in a pattern causes an error.
 .sp
 Another special sequence that may appear at the start of a pattern is (*UCP).
 This has the same effect as setting the PCRE2_UCP option: it causes sequences
-such as \ed and \ew to use Unicode properties to determine character types,
-instead of recognizing only characters with codes less than 256 via a lookup
-table. If also causes upper/lower casing operations to use Unicode properties
-for characters with code points greater than 127, even when UTF is not set.
+such as \ed (unless PCRE2_ASCII was set) and \ew to use Unicode properties
+to determine character types, instead of recognizing only characters with
+codes less than 256 via a lookup table. It also causes upper/lower casing
+operations to use Unicode properties for characters with code points greater
+than 127, even when UTF is not set.
 .P
 Some applications that allow their users to supply patterns may wish to
 restrict them for security reasons. If the PCRE2_NEVER_UCP option is passed to
@@ -670,7 +671,8 @@ determine character types, as follows:
   \ew  any character that matches \ep{L} or \ep{N}, plus underscore
 .sp
 The upper case escapes match the inverse sets of characters. Note that \ed
-matches only decimal digits, whereas \ew matches any Unicode digit, as well as
+matches only decimal digits and could be forced to match only the original
+set with PCRE2_ASCII, whereas \ew matches any Unicode digit, as well as
 any Unicode letter, and underscore. Note also that PCRE2_UCP affects \eb, and
 \eB because they are defined in terms of \ew and \eW. Matching these sequences
 is noticeably slower when PCRE2_UCP is set.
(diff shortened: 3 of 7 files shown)


zherczeg:
I don't really like the "add a feature and then add another feature for partially disable that feature" design. Instead I would add a full set of flags (can be set by a separate api call, 0 by default) which allows fine control over these features. E.g. `enable_unicode_d` enables the operation even if ucp is not set, `disable_unicode_d` disables it even if ucp is set. Disable could be the strongest. I know this is a bigger scope of work, but needed in the long run. The actual set of features could be computed with binary logical operations easily.

PhilipHazel:
I'm not sure we need quite that much fine control, but I understand your concern. I have been thinking about this and will post some ideas soon. The problem with using an api call to set these flags is that it can't also be set in the pattern in the way (*UTF) or (*UCP) are now used.

carenas:
While the new flag was meant to be used to move from C3 to C2, that part wasn't implemented yet.

What was implemented though (at least partially), was something missing from the matrix above in the "B" dimension and that matches B2 in the modified list:

B1: Only ASCII characters match (the default)
B2: Only relevant "locale" UTF characters are added
B3: All UTF characters are added (what PCRE2_UCP selects)

The result was that `\d` would only match [0-9] (and maybe the fullwidth digits[1]), instead of the other ~650 numerals that have `\p{Nd}`

The use case I had on mind was `git`, where assuming the content is unicode programming languages, you can expect to have identifiers with utf-8 characters and therefore need those characters for a correct identification with `\w` and `\b`, but wouldn't need the same for `\d`

Outside of that use case, I can also think that limiting `\w` to only expand characters of the script that is identified by the locale might be useful.

Of course overloading all that meaning in a single flag is difficult, and the fact that it interacts with the other two makes it hard to use and I can see why Zoltan suggested a different design.

[1] https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms

PhilipHazel:
I have continued to think ... luckily there is no rush on this ... I am beginning to think Zoltan's idea might be useful. There are plenty of bits available in the "extra" options and there is an existing api - pcre2_set_compile_extra_options(). Having independent control over \d, \w, and \s would be possible. However, there is also the question of the "in pattern" settings such as (*UTF) that should, if possible, mirror the external settings - but there are also external lockouts such as PCRE2_NEVER_UTF. Will we need more of those? More thought needed.....

PhilipHazel:
I lied above. There are only 2 bits left in the options argument to pcre2_compile(). Here is a proposal that retains compatibility. It isn't very pretty, but we have to work from where we are.

1. Use one of the two available bits for a new option PCRE2_CASELESS_RESTRICT. This behaves like PCRE2_CASELESS, except that it restricts ASCII characters to have only ASCII other cases, and non-ASCII characters to have non-ASCII other cases.

2.  Add a number of new PCRE2_EXTRA_xxx options to control the behaviour of \d etc independently:
  PCRE2_EXTRA_ASCII_D   \d matches ASCII only
  PCRE2_EXTRA_ASCII_S   \s matches ASCII only
  PCRE2_EXTRA_ASCII_W  \w matches ASCII only - and this also affects \b which is defined in terms of \w
  PCRE2_EXTRA_ASCII_POSIX   the POSIX [,,,] classes match ASCII only

These new options would be needed only if you want to have different treatment for those four things. Otherwise, as now, you either set UCP to get Unicode treatment for all, or don't set UCP, to get ASCII treatment for all, using the new PCRE2_CASELESS_RESTRICT bit if you want that kind of caselessness.

Changes would also be needed to the (*xxx) settings that are allowed in patterns. I'm not sure if we would need all the possibilities, though is should be easy to implement. There could be an overall (*ASCII) that sets them all.

All this need changes to pcre2_compile(), both interpretive matching functions, and the JIT code, of course.

carenas:
I think that even the CASELESS flag might be enough of a niche case, that it could be relegated to an EXTRA option.

One useful feature we might copy from [Rust's regex](https://docs.rs/regex/latest/regex/#opt-out-of-unicode-support) is the `(?u)` option to disable/enable UCP in parts of the expression.

Something I am curious about, is why were `(*UTF)` and `(*UCP)` invented, specially since if we are to invent `(*ASCII)` as well, we will end up burning that bit in the main option anyway.

PhilipHazel:
(*UTF) and (*UCP) were invented at user request, for situations where the user controls the pattern, but not the code. The original addition was in 2009, but the ChangeLog doesn't record any details. It's possible that it relates to the use of PCRE from non-C languages.

I suggested making PCRE2_CASELESS_RESTRICT a "main" option so that a user could just set that as an alternative to PCRE2_CASELESS, without having to mess with the extra options. But I willing to be persuaded. Anybody else have a view on this?

PhilipHazel:
I have changed my mind. I now agree the all the new options can be "extra"s. I intend to work on this shortly.

PhilipHazel:
I have just pushed a commit that implements PCRE2_EXTRA_CASELESS_RESTRICT. I also added tracking apparatus for the extra options as for the main options, in order to implement (?r) within a pattern to change this option dynamically. You can also use /caseless_restrict or /r in pcre2test. I have NOT updated the user documentation because I'm planning on doing it all in one go once I have implemented the other EXTRA options mentioned. The code changes in this commit are all in pcre2_compile() so the new option needed no changes to the matching functions or JIT. There are some new tests but I would not be at all surprised to learn that I have overlooked something,

PhilipHazel:
I have now pushed commits that implement PCRE2_EXTRA_ASCII_BSD, _BSS, _BSW, and _POSIX. The BS stands for "backslash", as used in some other option names, and I chose these names to make it clear exactly what was being forced to ASCII. The implementation is entirely in pcre2_compile.c. There is no documentation yet - that's coming next.

carenas:
Implemented already

PhilipHazel:
Bad news: I have realized that there is a major bug. The \b and \B escapes, which are documented as being dependent on \w, don't work correctly. This is because the change in behaviour of \b is implemented at match time, not compile time. Fixing this is inevitably going to involve changes to pcre2_match(), pcre2_dfa_match() and (sorry Zoltan) the JIT. My current plan is to add to the OP_WORDCHAR opcode a new one called OP_ASCII_WORDCHAR, so that all the distinguishing happens at compile time. (And the same for the negative ones, of course.)

PhilipHazel:
Oops, I meant OP_WORD_BOUNDARY, of course.

PhilipHazel:
I have pushed a commit that fixes this bug in the interpreters, but of course JIT is now broken. In the end, I did it the opposite way to what I described above. I invented OP_UCP_WORD_BOUNDARY and OP_NOT_UCP_WORD_BOUNDARY, so the original opcodes, like all the other opcodes for \d etc, are ASCII-only.

zherczeg:
I can fix the jit code later.

zherczeg:
I have fixed the jit code, please check

Issue #187, opened 2023-01-11
The code unit rules for _16 and _32 functions need to be more clearly explained


tempelmann:
After asking on the mailing list, I learned that I was wrong assuming that prce_match_16 would take the search buffer size in byte units. I had read https://www.pcre.org/current/doc/html/pcre2api.html#SEC15 before, and still misunderstood it.

What also misled me was that it was explained that `PCRE2_SIZE` is same as `size_t`, which is clearly a byte size (otherwise, it could overflow for units > 1 byte).

May I ask that this is explained more clearly, with these points:

1. When using the 16 or 32 bit functions, then the buffer length is half or a fourth of the buffer's length in bytes.
2. The 16 or 32 bit unichars are expected at their word-alignment. That means that (a) UTF-16BE chars won't be found in a buffer containing UTF-16BE and (b) if strings are searched in binary data (with the option `PCRE2_MATCH_INVALID_UTF`), they also won't be found if they're not aligned (in this case, a work-around might be to search multiple times with the start offset moved by 1 or more bytes).

Also, not sure if that's explained: The search code is not unicode-composition aware, i.e. when searching for unicode chars with accents (or umlauts), one should perform multiple searches with all possible composition forms.

tempelmann:
That's a bit imprecise. It's not "number of characters" but "number of code points", because there are characters that consist of two code points.

zherczeg:
This is easier to understand this way. Btw characters and code points are the same thing, code units are the atomic part of encoding.

PhilipHazel:
PCRE2 expects 16--bit data to be in vectors of uint16_t -- that is, in the natural BE/LE order for the current environment (same for 32-bit data). If a user is handling 16-bit data as individual bytes it is their responsibility to get it into uint16_t code units before passing it to PCRE2. I will work on the documentation to try to make this more clear.

PhilipHazel:
I have added a few sentences to pcre2api.1 to try to state more clearly that string lengths are always in code units, not bytes.

PR #188, opened 2023-01-12
doc: document limitations with PCRE2_MATCH_INVALID_UTF

carenas:
Closes: #187

### DIFF
--- a/doc/pcre2unicode.3
+++ b/doc/pcre2unicode.3
@@ -442,6 +442,10 @@ would match an instance of WORD that is surrounded by invalid UTF code units.
 Using PCRE2_MATCH_INVALID_UTF, an application can run matches on arbitrary
 data, knowing that any matched strings that are returned are valid UTF. This
 can be useful when searching for UTF text in executable or other binary files.
+.P
+The test for validity of Unicode characters is done for whole code units, so
+if the valid Unicode string is not aligned to a multiple of the code point size
+it won't be identified as valid.
 .
 .
 .SH AUTHOR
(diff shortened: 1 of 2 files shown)


doc/pcre2unicode.3
```diff
@@ -442,6 +442,12 @@ would match an instance of WORD that is surrounded by invalid UTF code units.
 Using PCRE2_MATCH_INVALID_UTF, an application can run matches on arbitrary
 data, knowing that any matched strings that are returned are valid UTF. This
 can be useful when searching for UTF text in executable or other binary files.
+.P
+The test for validity of Unicode characters is done for whole code units, so
+if the valid Unicode string is not aligned to a multiple of the code point size
+it won't be identified as valid. If that is a possibility on your data and
+unaligned access is not a problem in your system, you might want to try the
+matches with different starting byte offsets as a work around.
```
zherczeg, line 450: This can be used only for utf16. And I think we should not encourage this use case.

PhilipHazel:
I agree there should be some rewording, but I'm not sure it should mention breaking down 16-bit units. PCRE expects 16-bit data to be in uint16_t vectors, that is, in the natural BE/LE arrangement for the environment. If somebody is handling 16-bit data in bytes, it is their responsibility to get it into the right format for PCRE. I will work on the documentation to try to make this clear.

carenas:
The proposed change doesn't mention breaking down 16-bit units.  The misguided attempt to suggest doing so as a workaround was only part of the original version, because it was explicitly mentioned in the ticket, and only was added because I found it funny while testing as it would match an UTF-16 'a' out of pure luck when using an BE 'a' after half a broken other composite character:

```
code = pcre2_compile(L'a', 1, PCRE2_MATCH_INVALID_UTF, &error, &erroff, NULL);
..
int buf = L'a';
buf <<= 8;
buf |= 0xd8000000;
memcpy(subject, &buf, sizeof(int));
error = pcre2_match(code, (char *)subject + 1, 3, 0, 0, data, NULL);
assert(d == 1);
```

PR #190, opened 2023-01-13
add pcre2_match_data_reset()

carenas:
This is just a mirror of `pcre2_match_data_free()` (and indeed changes its implementation to reuse the new code), but that would be useful when match_data is being reused (specially since 10.41, as it now contains the heapframes)

It adds a knob that could be useful when getting rid of the heapframes might be needed (ex: they'd grown abnormally large because of a bad regex) and plugs a possible leak when PCRE2_COPY_MATCHED_SUBJECT was set.

I'd seen also at least 1 case where an implementation was copying the match_data to a stack variable (for thread safety) and then would leak itself to dead as it was never calling `free` but that just happened to work before.


carenas:
This is a bad idea, if a match_data is being reused this saves just the free/malloc calls of the match_data itself. the improved wording in the documentation was worth saving but will be included in a different request.

PR #191, opened 2023-01-13
add pcre2_get_match_data_heapframes_size()

carenas:
The proposed code, allows for an external application to implement logic similar to:
```
PCRE2 version 10.42 2022-12-11
  re> /\[(a)]{1000}/expand,framesize
Frame size for pcre2_match(): 16128
data> \[a]{1000}\=ovector=1
Matched, but too many substrings
 0: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
data> 
  re> /a/memory,heapframes_size
Memory allocation (code space): 9
Heapframes size for pcre2_match(): 20643840
data> a\=ovector=0
 0: a
data> 
  re> /a/heapframes_size
Heapframes size for pcre2_match(): 20480
```
So that the system could recover from memory pressure in cases were match_data is being reused.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -5,14 +5,18 @@ Change Log for PCRE2 - see also the Git log
 Version 10.43 xx-xxx-202x
 -------------------------
 
-1. The test program added by change 2 of 10.42 didn't work when the default 
-newline setting didn't include \n as a newline. One test needed (*LF) to ensure 
+1. The test program added by change 2 of 10.42 didn't work when the default
+newline setting didn't include \n as a newline. One test needed (*LF) to ensure
 that it worked.
 
-2. Added the new freestanding POSIX test program to the ManyConfigTests script 
-in the maint directory (overlooked in 2 below). Also improved the selection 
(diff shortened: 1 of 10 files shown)


PhilipHazel:
I have now played with this a bit. I'm happy with the new function, but I'm going to re-work the changes to pcre2test. It seems wrong to me for the heapframes_size option to be a pattern option; I'm going to change it to a subject option. As a pattern option, it gives information about the *previous* match, which doesn't seem right.

Issue #196, opened 2023-01-24
8-byte Out of Bounds Write onto the Heap


WhatTheFuzz:
Found a small faulting test case for `pcre2test` that causes a segmentation fault on some architectures (x86). It looks like the fault is in `process_data`. The logic for filling `ovector` may be off by one. Specifically, the offending line is

```c
for (j = 0; j < 2*oveccount; j++) ovector[j] = JUNK_OFFSET;
```

in the file:line `pcre2test.c:7634`

```shell
# Show compiler and arch. 
$ clang --version
Ubuntu clang version 14.0.0-1ubuntu1
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

# Compile with ASAN to show the bug on amd64 platforms. On x86 (-m32), it crashes.
$ CC=clang CFLAGS="-fsanitize=address -Og -ggdb" LD_FLAGS="-fsanitize=address" ./autogen.sh
$ CC=clang CFLAGS="-fsanitize=address -Og -ggdb" LD_FLAGS="-fsanitize=address" ./configure
$ CC=clang CFLAGS="-fsanitize=address -Og -ggdb" LD_FLAGS="-fsanitize=address" make

# Show exploitation mitigations.
$ checksec .libs/pcre2test 
    Arch:     amd64-64-little
    RELRO:    Partial RELRO
    Stack:    No canary found
    NX:       NX enabled
    PIE:      PIE enabled
    ASAN:     Enabled
    UBSAN:    Enabled
    
$  .libs/pcre2test  ./80271ad5c66883facd8415181d5c632ff297b23a45a70760d831d2d8cd4a6ddd /dev/null
=================================================================
==134099==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x612000000180 at pc 0x5619337c3587 bp 0x7ffcc068faf0 sp 0x7ffcc068fae8
WRITE of size 8 at 0x612000000180 thread T0
    #0 0x5619337c3586 in process_data pcre2/src/pcre2test.c:7634:48
    #1 0x5619337b7ec7 in main pcre2/src/pcre2test.c:9441:12
    #2 0x7f548e8e1d8f in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #3 0x7f548e8e1e3f in __libc_start_main csu/../csu/libc-start.c:392:3
[...]

PR #198, opened 2023-01-25
add a C23 inspired checked integer multiplication helper

carenas:
A combined approach that addresses #196 and replaces #177

Unlike the original It allows in both cases to handle larger sizes than INT_MAX when possible

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -73,6 +73,32 @@ AC_SYS_LARGEFILE
 
 PCRE2_VISIBILITY
 
+# Check for the mul_overflow() builtin
+
+AC_MSG_CHECKING([for __builtin_mul_overflow()])
+AC_LANG_PUSH([C])
+tmp_CFLAGS=$CFLAGS
+CFLAGS="$CFLAGS -Werror"
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
+		#ifdef HAVE_SYS_TYPES_H
+		#include <sys/types.h>
+		#endif
+		#ifdef HAVE_STDLIB_H
+		#include <stdlib.h>
+		#endif
+		int a, b;
+		size_t m;
+	]], [[__builtin_mul_overflow(a, b, &m)]])],
+	[pcre2_cc_cv_builtin_mul_overflow=yes],
+	[pcre2_cc_cv_builtin_mul_overflow=no])
+AC_MSG_RESULT([$pcre2_cc_cv_builtin_mul_overflow])
+if test "$pcre2_cc_cv_builtin_mul_overflow" = yes; then
+	AC_DEFINE([HAVE_BUILTIN_MUL_OVERFLOW], 1,
+		[Define this if your compiler provides __builtin_mul_overflow()])
+fi
+CFLAGS=$tmp_CFLAGS
+AC_LANG_POP([C])
+
 # Check for Clang __attribute__((uninitialized)) feature
 
 AC_MSG_CHECKING([for __attribute__((uninitialized))])
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -7112,15 +7112,12 @@ for (;; pptr++)
 
         /* In the pre-compile phase, we don't actually do the replication. We
         just adjust the length as if we had. Do some paranoid checks for
-        potential integer overflow. The INT64_OR_DOUBLE type is a 64-bit
-        integer type when available, otherwise double. */
+        potential integer overflow. */
 
         if (lengthptr != NULL)
           {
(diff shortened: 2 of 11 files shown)


WhatTheFuzz:
Closes #196

PhilipHazel:
I support this idea of making a common source for this check. However, as pcre2_ckd_smul() is an internal function, and not part of the external API, it should really be defined using the PRIV macro (see comments in pcre2_internal.h).

carenas:
I know, but it seems ackward, since unlike all other helper functions, this one doesn't require a different version per codesize.

Would it be better to get rid of the C file and put it directly in the header, that way it would be also be also inlined where needed and work almost like the original?

src/pcre2_internal.h
```diff
@@ -2042,6 +2042,38 @@ extern void *       _pcre2_memmove(void *, const void *, size_t);
...
+PRIV(ckd_smul)(PCRE2_SIZE *r, int a, int b)
+{
+INT64_OR_DOUBLE m;
+
+#ifndef NDEBUG
+if (a < 0 || b < 0) abort();
+#endif
+
+m = (INT64_OR_DOUBLE)a * (INT64_OR_DOUBLE)b;
+if (!(PCRE2_SIZE_MAX - m > 0)) return TRUE;
```
zherczeg, line 2071: `m < PCRE2_SIZE_MAX` is enough
btw if either argument is negative this is true. You could use unsigned multiplication if negative numbers are not possible. 
I also remember something when int max is converted to infinity, but it might be the opposite conversion direction.

PhilipHazel:
The thing about the helper functions is that there may be environments in which they are not exported beyond the library. This means you cannot use them in pcre2test. That is one of the reasons why the PRIV stuff exists.

PhilipHazel:
Also, the use of the PRIV notation makes it clear that this is a library-wide helper function, and not something local.

src/pcre2_internal.h
carenas: That is how I wrote it [originally](https://github.com/PCRE2Project/pcre2/actions/runs/4006442948/jobs/6877970854), but it triggered unhelpful (-Wsign-compare) warnings in most builds.

Note that using unsigned here isn't useful as the original variables are signed, and indeed this whole function exists solely to prevent triggering undefined behaviour with signed integer multiplications that overflow.

configure.ac
```diff
@@ -638,6 +638,8 @@ fi
 if test "$enable_debug" = "yes"; then
   AC_DEFINE([PCRE2_DEBUG], [], [
     Define to any value to include debugging code.])
+else
+  CFLAGS="$CFLAGS -DNDEBUG"
```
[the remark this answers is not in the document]
carenas: NDEBUG is fairly well known, indeed it is enabled by default by our own cmake build process when indicating a production build and is used elsewhere to disable asserts (which is what it is doing here).

What would you suggest we do then, delete the open coded assert from the code?

src/pcre2_compile.c
```diff
@@ -7334,21 +7328,19 @@ for (;; pptr++)
...
-                    > (INT64_OR_DOUBLE)INT_MAX ||
+            PCRE2_SIZE delta;
+            if (PRIV(ckd_smul)(&delta, repeat_max,
+                               length_prevgroup + 1 + 2 + 2*LINK_SIZE) ||
                 OFLOW_MAX - *lengthptr < delta)
               {
               *errorcodeptr = ERR20;
               return 0;
               }
+            delta -= (2 - 2*LINK_SIZE);   /* Last one doesn't nest */
```
carenas, line 7343: should be `(2 + 2*LINK_SIZE)` and because is not considered in the check with OFLOW_MAX will change behaviour

carenas:
An alternative to the current proposal which doesn't use the `static inline` function available in the [external branch](https://github.com/carenas/pcre2/pull/new/overflow-semiprivate).

Also included some changes raised in the review, but not sure if they are sufficient.

PhilipHazel:
I prefer not to put code into .h files, and I'm happy with a separate function, but it must use the PRIV mechanism, for consistency if nothing else. I will do some work on this after I have dealt with #186.

carenas:
FWIW, both (this version with the code in the header) and the alternative use PRIV.

There is also an additional [branch](https://github.com/carenas/pcre2/tree/overflow-builtin) with a patch on top that uses the compiler builtins if available (which works for gcc and clang) so the assert() might not be even needed and that hopefully works also for Zoltan.

configure.ac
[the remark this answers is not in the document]
zherczeg: Yes. There is PCRE2_DEBUG for debugging.

PR #199, opened 2023-01-30
Ucptest updates

carenas:
The first patch, just updates the output from ucptest, which has been out of sync one way or another since the Bidi changes.  Some of the results at least in testoutput2 might not be correct, but at least this way they are visible.

The second patch does fix a minor issue in the CASELESS_RESTRICT code that was just merged and another old bugs from ucptest that I found while testing as well as further cleanups


PhilipHazel:
Looks reasonable to me, but I'm going to finish current work on PCRE2_EXTRA_ASCII_xxx before trying to merge your patches. (I consider myself to be a Git novice and I want to keep things simple if possible.)

PhilipHazel:
After merging this and the Unicode update, I see a lot of differences in the ucptestdata output. I'm wondering if something has gone wrong somewhere. The main PCRE2 tests run clean.

carenas:
I think we have an old bug with boolean properties but still unsure of its impact, as mentioned before not all changes in the `ucptestdata` output made sense

PhilipHazel:
It may be the Unicode update that's the problem. I am in the process of merging the other PCRE2_EXTRA_ASCII options, but have to stop now until tomorrow, when I hope I'll be able to push the result. Then I must do the documentation but after that I'll take a closer look at the properties stuff - I've forgotten the details so will have to relearn how the code works, and where all the data comes from.

PR #202, opened 2023-02-03
maint: avoid duplicated boolean properties and bad script extensions

carenas:
`ucptest` was misbehaving and showing the wrong properties and finding the wrong characters.


PhilipHazel:
Thanks for sorting this out. I am always happy when someone else fixes a problem before I have got round to it. :-)

PR #208, opened 2023-02-09
pcre2grep: documentation updates

carenas:
using -E without -i doesn't make sense, but it could give confusing results otherwise


PhilipHazel:
It would make sense to have -E without -i if a search pattern contains (?i). That's why I didn't make -E imply -i, but perhaps the documentation could be more explicit.

carenas:
-E as documented makes sense, as well as your use case so I am just including the unrelated documentation changes instead

PR #213, opened 2023-03-01
Add unsupported error code for jit

zherczeg:


minipli-oss (changes requested):
I'm just questioning myself if this change breaks existing applications because of the new error code? It did, at least, had an impact on the tests, so other projects might be impacted as well?

_/me goes looking..._

For example, GNU [grep](https://git.savannah.gnu.org/cgit/grep.git/tree/src/pcresearch.c?h=v3.8#n213) would be affected, as it explicitly tests the possible error codes:
```c
  ec = pcre2_jit_compile (pc->cre, PCRE2_JIT_COMPLETE);
  if (ec && ec != PCRE2_ERROR_JIT_BADOPTION && ec != PCRE2_ERROR_NOMEMORY)
    die (EXIT_TROUBLE, 0, _("JIT internal error: %d"), ec);
```
With your change it would `die()` on certain patterns instead of falling back to intepreter mode.

Judging by previous arguments made by yourself and @PhilipHazel in https://github.com/PCRE2Project/pcre2/pull/157, such an API-breaking change is stuff for a PCRE3 only.

zherczeg:
You are right, I need to read the documentation first:
https://www.pcre.org/current/doc/html/pcre2_jit_compile.html

"In particular, PCRE2_ERROR_JIT_BADOPTION is returned if JIT is not supported or if an unknown bit is set in options." - So I cannot change the badoption case (need to revert).

The code in git is simply wrong, they also did not read the documentation. In general we are talking about pcre2 api compatilbity, not bug compatibility with existing applications.

minipli-oss:
Ahh, saw your edit. So you meant `grep`, not `git`. Still, breaking `grep` won't be welcomed by many, I fear ;)

src/pcre2_error.c
```diff
@@ -272,6 +272,7 @@ static const unsigned char match_error_texts[] =
   /* 65 */
   "internal error - duplicate substitution match\0"
   "PCRE2_MATCH_INVALID_UTF is not supported for DFA matching\0"
+  "pattern does not supported by jit\0"
```
carenas, line 275: removing or using "is" instead of "does" would be a more appropriate phrase

carenas:
note some documentation might need updating as well.

the implications for this change from 7a233130adeedc0e37cd10a1b31ddafa8db6b65a, for example, is likely to require changes in users of this API.

zherczeg:
This patch is just exploring options. I have another idea: we could add flag to jit compile, which allows returning other error codes than nomemory / badoption. Will not make the code much more complex, and we have plenty of options. What do you think?

PhilipHazel:
I think it's a good idea to require a new option flag to enable the new return codes. That way there is no danger of breaking existing code, and applications can update in their own time if they want to make use of the new codes.

carenas:
> The code in grep is simply wrong, they also did not read the documentation. In general we are talking about pcre2 api compatilbity, not bug compatibility with existing applications.

FWIW GNU grep has been fixed

https://git.savannah.gnu.org/cgit/grep.git/commit/src/pcresearch.c?id=6e7253de1dd443be5f498078e1807ddfa4d364de

PR #222, opened 2023-04-06
pcre2grep: avoid UCP mode for \d

carenas:
As suggested in the discussion[1] from `git` by the GNU grep developers, and with the hope to prevent further divergence.

It is important to note that this will make matching from `\d` and `[:digit:]` inconsistent in pcre2grep hence published as a draft.

[1] https://lore.kernel.org/git/ed237a07-2f77-74eb-2f52-49b9b8f08873@cs.ucla.edu/


PhilipHazel:
I agree with this in principle, but it probably needs leaving until #223 gets sorted out.

PR #223, opened 2023-04-06
avoid inconsistency between \d and [:digit:] when using /a

carenas:
Address the inconsistency raised by the suggested change in #222 and that would also affect other users of the API.

for added flexibility PCRE2_EXTRA_ASCII_DIGIT is independent of PCRE2_EXTRA_ASCII_BSD.


PhilipHazel:
Thanks for doing a lot of work on this. My feeling is that I don't really like making \s, \w, and \d different in this regard, and therefore I prefer to keep all three PCRE2_EXTRA_ASCII_BS{S,W,D}  options as they are. There could be an *additional* option PCRE2_EXTRA_ASCII_DIGIT that does what you want - affects both \d and [:digit:] but is it really necessary to single out [:digit;] as a special case of POSIX classes? Or do we need separate PCRE2_EXTRA_ASCII_POSIX_{DIGIT,SPACE,WORD} options?

carenas:
> do we need separate PCRE2_EXTRA_ASCII_POSIX_{DIGIT,SPACE,WORD} options?

A few arguments against adding SPACE and WORD:

* they are already defined differently in UCP mode by using different Unicode classes, so for example \w != [:word:] already.
* reverting them back to ASCII can already be done with PCRE2_EXTRA_ASCII_POSIX.
* I don't see a use case that wouldn't be supported already by either not setting PCRE2_UCP or the current flags.

PhilipHazel:
I'm happy to leave SPACE and WORD until somebody actually wants them. You will see that I've done the merge. There is one further point: I added new internal option settings such as (?aP) for PCRE2_EXTRA_ASCII_POSIX and the other previous ASCII options. Do we need to add in something for DIGIT? I've already used (?aD) for BSD and I'd rather not change that because it matches (?aS) for BSS and (?aW) for BSW. I wondered about (?aPD) but that seems a bit clumsy.

carenas:
> Do we need to add in something for DIGIT? I've already used (?aD) for BSD and I'd rather not change that because it matches (?aS) for BSS and (?aW) for BSW. I wondered about (?aPD) but that seems a bit clumsy.

While not the same (?aP) can already be used for that effect:

```
PCRE2 version 10.43-DEV 2023-01-15
  re> /(?aP)[[:digit:]](?-aP)[[:digit:]]/B,ucp
        Bra
        [0-9]
        [\p{Nd}]
        Ket
        End
```

My thinking was to punted for now and see if that wouldn't be enough, and if eventually required, see what would make more sense, so either:

* reuse (?aD)
* add something like (?ad) which I agree seems as clumsy

PhilipHazel:
Agreed. I also thought of (?ad) and disliked it. Let's see if a need arises.

PR #228, opened 2023-04-17
left shift of 3 by 30 places cannot be represented in type 'int'

JetXujing:
Although the use of type 'int' here has no effect on the result, there are warnings in some cases.


zherczeg:
JIT compiler changes should be opened in its own repository:
https://github.com/zherczeg/sljit

Furthermore use `sljit_ins` type not `sljit_u32`

JetXujing:
OK, thanks for the reminder, this is my first time contributing to pcre2 and I will contribute to https://github.com/zherczeg/sljit later.

JetXujing:
I have created a new PR in https://github.com/zherczeg/sljit/pull/159. Close it here.

PR #237, opened 2023-04-20
pcre2test: avoid printing invalid utf trail in partial match

carenas:
This could result in a buffer over read as reported in the linked ticket and that was only possible after 10.34 when match_invalid_utf was introduced, or if invalid data was matched without UTF validation, which is known to have undefined behaviour.

Also makes utf82ord() save when used with invalid UTF, as the same issue could be triggered in other ways (ex: using aftertext) as shown by:

```
PCRE2 version 10.34 2019-11-21
  re> /(?<=..)X/match_invalid_utf,aftertext
data> \xc3\xb1x{10ffff}X\xb1\xf4\x8f\xbf
```

Fixes: #235

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,6 +66,10 @@ testtemp2
 testtemp2grep
 testtry
 testtrygrep
+testSinput
+testbtables
+testsaved1
+testsaved2
 
 m4/libtool.m4
 m4/ltoptions.m4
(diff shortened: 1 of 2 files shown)


PhilipHazel:
The tests are failing with JIT, it seems.

carenas:
Shouldn't fail when stacked on top of #238, but will update it once that is settled

Issue #243, opened 2023-05-01
MinGW: Failed to build static lib `undefined reference to '__imp_pcre2_regcomp'`


rockdaboot:
### The issue
`pcre2` is a dependency of [GNU Wget2](https://gitlab.com/gnuwget/wget2).
We have two MinGW CI images to build deps like `pcre2`, one for dynamic linking and one for static linking.
The static build of `pcre2` fails since ~6 months.

The build command sequence is
```
git clone https://github.com/PCRE2Project/pcre2
cd pcre2
export CFLAGS="-O2 -g"
./autogen.sh
./configure --host=x86_64-w64-mingw32 --prefix=/usr/local/x86_64-w64-mingw32 --disable-shared --enable-static
make
```
(Please see the Dockerfile below for the ENV variables set)

### Error message
```
...
#25 37.55   CCLD     pcre2posix_test.exe
#25 37.67 /usr/bin/x86_64-w64-mingw32-ld: src/pcre2posix_test-pcre2posix_test.o: in function `main':
#25 37.68 /usr/local/pcre2/src/pcre2posix_test.c:134: undefined reference to `__imp_pcre2_regcomp'
```

### Dockerfile
The failed build is part of a list of dependency builds.
<details>
FROM debian:bookworm-slim

LABEL maintainer "Tim Rühsen <tim.ruehsen@gmx.de>"

WORKDIR /usr/local

RUN apt-get update -y && apt-get install --no-install-recommends -y \
        git \
        autoconf \
        autoconf-archive \
        autopoint \
        automake \
        autogen \
        libtool \
        make \
        wine \
        flex \
        bison \
        gettext \
        gperf \
        mingw-w64 \
        pkg-config-mingw-w64-x86-64 \
        ca-certificates \
        wget \
        patch \
        texinfo \
        gengetopt \
        curl \
        lzip \
        pandoc \
        rsync \
        ccache \
        python3 \
        binfmt-support \
        bc

ENV PREFIX="x86_64-w64-mingw32"
ENV INSTALLDIR="/usr/local/$PREFIX"
ENV PKG_CONFIG_PATH="$INSTALLDIR/lib/pkgconfig:/usr/$PREFIX/lib/pkgconfig" \
    PKG_CONFIG_LIBDIR="$INSTALLDIR/lib/pkgconfig" \
    PKG_CONFIG="/usr/bin/${PREFIX}-pkg-config" \
    CPPFLAGS="-I$INSTALLDIR/include" \
[...]

PR #246, opened 2023-05-07
pcre2posix_test: fix build when using a static library

carenas:
Only relevant for Windows builds, where the lack of a PCRE2_STATIC definition implies the use of a DLL

Fixes: #243

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -566,6 +566,7 @@ endif # WITH_PCRE2_8
 ## If the 8-bit library is enabled, build the POSIX wrapper test program and
 ## arrange for it to run.
 
+if WITH_DLL
 if WITH_PCRE2_8
 TESTS += pcre2posix_test
 noinst_PROGRAMS += pcre2posix_test
@@ -573,6 +574,7 @@ pcre2posix_test_SOURCES = src/pcre2posix_test.c
 pcre2posix_test_CFLAGS = $(AM_CFLAGS)
 pcre2posix_test_LDADD = libpcre2-posix.la libpcre2-8.la
 endif # WITH_PCRE2_8
+endif # WITH_DLL
 
 ## If JIT support is enabled, arrange for the JIT test program to run.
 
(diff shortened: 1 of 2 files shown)


PhilipHazel:
I don't think this patch can be the right thing to do. The program pcre2posix_test is supposed to be an ordinary client program that uses the POSIX interface to PCRE2. It should not need access to the internal config file. I have no access to Windows, nor do I really know anything about it, so I'm not sure what needs to be done here, but all the client should need to #include is pcre2posix.h. Maybe there's something in pcre2.h that should also be there.

carenas:
Ordinary programs (at least in Windows), would define PCRE2_STATIC before including the pcre2 headers, if they intend to link statically and as documented in pcre2demo.c with:

> Building under Windows:
>
> If you want to statically link this program against a non-dll .a file, you must
> define PCRE2_STATIC before including pcre2.h, so in this environment, uncomment
> the following line. */
>
> /* #define PCRE2_STATIC */

I was tempted to do that as a "solution", but then the program will fail to build instead if there is no static library to link with.

carenas:
Alternatively; since the objective is not to break the build unnecessarily, it could be skipped unless a DLL is built to link with.

PhilipHazel:
Yes, I think that is a better solution than changing either of the pcre2posix.{c,h} files, which have, over the years, been refined for many different environments, so I am very reluctant to make changes there.  **However**, (a) I think the plain WITH_DLL test should at the very least include a test for Windows and could perhaps be renamed as ENABLE_SHARED, or maybe a flag called BUILD_PCRE2POSIX_TEST could be set up to make things absolutely clear. (b) There needs to be similar updating for CMake.

If am happy to take over making these changes if you like and maybe I can also add to the build.yml file to create a static test for Windows. Let me know what you prefer.

carenas:
> There needs to be similar updating for CMake.

CMake works fine and doesn't need a similar "fix" because the way it defines linking with the library itself, includes a rule to add that "define" through a compiler flag when needed.

Something similar could be done with "autotools", if the building of pcre2posix_test would be required to find the correct CFLAGS to use by using pkg-config to query the just generated definition file, but though it was overkill for what was needed of this test. 

> If am happy to take over making these changes if you like and maybe I can also add to the build.yml file to create a static test for Windows. Let me know what you prefer.

Since you seem to have a particular solution in mind, it might be better for you to iterate on it directly, but my guess is that (as mentioned in the linked issue) since the problem runs deeper we are likely to need changes to the pcreposix.{c,h} files eventually so there is no longer a conflict with the visibility feature.

carenas:
superseded by #249 which allow for building `pcre2posix_test` regardless of the type of library being built

PR #248, opened 2023-05-08
avoid the possibility of dangling pointers while processing capture items in pcre2_compile

carenas:
Raised by gcc 13 (from MinGW and Fedora 38):

```
src/pcre2_compile.c: In function 'compile_regex':
src/pcre2_compile.c:8319:17: warning: storing the address of local variable 'capitem' in '*cb.open_caps' [-Wdangling-pointer=]
 8319 |   cb->open_caps = &capitem;
      |   ~~~~~~~~~~~~~~^~~~~~~~~~
```

While the cb.open_caps is not needed outside the possibly recursive calls to compile_regex() and compile_branch(), it is safer to not keep it with possibly dangling references once is no longer needed.


PhilipHazel:
I'm missing something here. I don't see how passing the chain of open_capitems via its own argument is any different to passing it as a member of the compile block structure. The items themselves are still on the stack. Oh, perhaps the idea is to confine this value to the nested compile_{branch,regex} functions? OK, I think I can understand that. In which case, the patch looks straightforward.

carenas:
Updated the commit message to hopefully make it clearer.

PR #249, opened 2023-05-09
windows: integrate visibility attributes without conflicts

carenas:
Add PCRE2POSIX_SHARED which will be used Instead of PCRE2_STATIC to indicate that linking with a shared library is expected (specially relevant in Windows), but unlike the previous solution, it will be able to find the function names even if no DLL was available by not changing the names of the functions to ONLY fit the ones expected from a DLL.  This also makes the use of PCRE2POSIX_SHARED "optional", as it allows linking with a static or dynamic library when not used.

To avoid the conflict that was triggered by building with MinGW, set a different variable (PCRE2_EXPORT) that could be then embedded as needed into the corresponding `PCRE2*_EXP_*` variables.

As a bonus, allow for detecting the visibility attributes also with cmake.

Fixes: #104
Fixes: #243


src/pcre2_internal.h
```diff
@@ -122,17 +122,13 @@ only if it is not already set. */
...
+#      define PCRE2_EXP_DECL       extern PCRE2_EXPORT
 #      define PCRE2_EXP_DEFN
 #    endif
 #  else
-#    ifdef __cplusplus
-#      define PCRE2_EXP_DECL       extern "C"
-#    else
-#      define PCRE2_EXP_DECL       extern
-#    endif
+#    define PCRE2_EXP_DECL         extern PCRE2_EXPORT
```
carenas, line 129: this will break using a C++ compiler

Makefile.am
```diff
@@ -304,9 +303,7 @@ src/config.h.generic: configure.ac
 	test -f _generic/src/config.h
 	perl -n \
 	  -e 'BEGIN{$$blank=0;}' \
-	  -e 'if(/PCRE2_EXP_DEFN/){print"/* #undef PCRE2_EXP_DEFN */\n";$$blank=0;next;}' \
-	  -e 'if(/to make a symbol visible/){next;}' \
-	  -e 'if(/__attribute__ \(\(visibility/){next;}' \
+	  -e 'if(/(.+?)\s*__attribute__ \(\(visibility/){print"$$1\n";$$blank=0;next;}' \
```
carenas, line 306: this could break if the compiler used to build this doesn't support visibility attributes (but that could be addressed later if the maintainer moves his workstation to Windows)

PR #253, opened 2023-05-13
Add some test cases

JetXujing:
"Add tests which use recurse to test \K and Mark in atomic scope" is used to override two branches of the init_frame function, as follows:
https://github.com/PCRE2Project/pcre2/blob/1bc34ffa64c33381d793fb5cdddf3f484e603d23/src/pcre2_jit_compile.c#L2188
https://github.com/PCRE2Project/pcre2/blob/1bc34ffa64c33381d793fb5cdddf3f484e603d23/src/pcre2_jit_compile.c#L2197

"Add the test which use firstline,utf,match_invalid_utf together" is used to overwrite a branch of do_utfreadnewline_invalid:
https://github.com/PCRE2Project/pcre2/blob/1bc34ffa64c33381d793fb5cdddf3f484e603d23/src/pcre2_jit_compile.c#L4657

### DIFF
--- a/testdata/testinput10
+++ b/testdata/testinput10
@@ -636,4 +636,8 @@
     qchq\=ph
     qchq\=ps
 
+/line1\nbreak/firstline,utf,match_invalid_utf
+    line1\nbreak
+    line0\nline1\nbreak
+
 # End of testinput10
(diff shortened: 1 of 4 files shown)


JetXujing:
In fact, the test cases I added simply covered the corresponding branch, but didn't really test the functionality of the options. In fact, I don't have a good idea for these test cases.

Anyone with a good idea to optimize these test cases.

PhilipHazel:
As this stands, the tests are not running. Looks like you updated the tests without updating the output?

Issue #257, opened 2023-05-30
Matching a 32 MB binary with mostly NULs freezes on arm64 with JIT


tempelmann:
I have a sample file (attached) then, when searching for a string in the file, the match function won't return if JIT is enabled and if it's on an ARM (Apple Silicon) machine. Without JIT, or on Intel arch, the problem does not occur.

I have reproduced this issue with both 10.42 and the current master (as of May 30, 2023).

I'm attaching the source code for testing and the zipped sample file (unpack it and move "HFS4_small.dmg" into the same folder where to the built test program lies). I also attach the built test program (built for ARM and x64 for macOS).

To reproduce: Run the pcre2demo program. On an Intel Mac, it should print "finished". On an ARM, it'll not finish, at least not within the < 1 second it takes on Intel, nor within a minute.

[pcre2demo.zip](https://github.com/PCRE2Project/pcre2/files/11606601/pcre2demo.zip)
[HFS4_small.dmg.zip](https://github.com/PCRE2Project/pcre2/files/11606603/HFS4_small.dmg.zip)
[pcre2demo.c.zip](https://github.com/PCRE2Project/pcre2/files/11606605/pcre2demo.c.zip)

zherczeg:
Would it be possible to run your code with gdb, stop after a second, and get a backtrace?

tempelmann:
Zoltan, can you give me detailed instructions on which commands I need to use? Especially, how do I break into gdb after the code hangs?

zherczeg:
run your binary with gdb: `gdb --args prog_name prog_args...`
type `r` (short form of run)
wait 1 sec
ctrl+c to stop execution
type `bt` (short form of backtrace)
copy the output here

tempelmann:
```
% lldb pcre2demo 
(lldb) target create "pcre2demo"
Current executable set to '/Users/tt/Desktop/pcre2demo' (arm64).
(lldb) r
Process 14298 launched: '/Users/tt/Desktop/pcre2demo' (arm64)
Process 14298 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
    frame #0: 0x000000010003e1a8 pcre2demo`ffcs_mask + 148
pcre2demo`ffcs_mask:
->  0x10003e1a8 <+148>: fmov   x9, d2
    0x10003e1ac <+152>: cbnz   x9, 0x10003e1e0           ; <+204>
[...]

PR #259, opened 2023-05-31
jit: add support for partial matches to arm64 simd helper

carenas:
Do the minimal change required to address the underlying issue (not a hang, but very degraded performance, with the possibility of corruption that could result in an infinite loop by accidentally decrementing `str_ptr`) that was reported.

Fixes: #257

### DIFF
--- a/src/pcre2_jit_neon_inc.h
+++ b/src/pcre2_jit_neon_inc.h
@@ -82,7 +82,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #  endif
 # endif
 
-static sljit_u8* SLJIT_FUNC FF_FUN(sljit_u8 *str_end, sljit_u8 *str_ptr, sljit_uw offs1, sljit_uw offs2, sljit_uw chars)
+static sljit_u8* SLJIT_FUNC FF_FUN(sljit_u8 *str_end, sljit_u8 **str_ptr, sljit_uw offs1, sljit_uw offs2, sljit_uw chars)
 #undef FF_FUN
 {
 quad_word qw;
@@ -171,7 +171,7 @@ else
   }
 # endif
 
-str_ptr += IN_UCHARS(offs1);
+*str_ptr += IN_UCHARS(offs1);
 #endif
 
 #if PCRE2_CODE_UNIT_WIDTH != 8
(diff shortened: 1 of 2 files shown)


src/pcre2_jit_simd_inc.h
```diff
@@ -944,9 +944,12 @@ if (common->mode == PCRE2_JIT_COMPLETE)
 
 /* Fast forward STR_PTR to the result of memchr. */
 OP1(SLJIT_MOV, STR_PTR, 0, SLJIT_RETURN_REG, 0);
-
 if (common->mode != PCRE2_JIT_COMPLETE)
+  {
```
carenas, line 948: this needs to be skipped if !STR_PTR

src/pcre2_jit_neon_inc.h
```diff
@@ -82,7 +82,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #  endif
 # endif
 
-static sljit_u8* SLJIT_FUNC FF_FUN(sljit_u8 *str_end, sljit_u8 *str_ptr, sljit_uw offs1, sljit_uw offs2, sljit_uw chars)
+static sljit_u8* SLJIT_FUNC FF_FUN(sljit_u8 *str_end, sljit_u8 **str_ptr, sljit_uw offs1, sljit_uw offs2, sljit_uw chars)
```
zherczeg, line 85: This function should return with str_ptr.

carenas: it does, but only if the match succeeds.

the str_ptr "reference" is needed to keep track of all the progress made if there is no match, specially as it loops over until the end_ptr looking for matches.

zherczeg: Something is not right here. This variant (with two offsets) of the code should not be used when partial matching is enabled.

carenas: Not sure if I follow, but there s plenty wrong with the arm64 simd helper; this ONLY tackles the issues that were reported and that affect mostly the single character + mask, one offset function.

 Since they all use the same template function, the changes to the second parameter to be a pointe to pointer so it can be used as IO propagate to all functions though, which also include the one with 4 characters and 2 offsets which wasn't being triggered AFAIK but that would be otherwise unaffected by the change (except that it might perform slightly better, just like the equivalent one from other architectures).

zherczeg:
Is the patch ready?

carenas:
AFAIK It addresses the reported issue and was not planning any further changes to it.

It would be nice to get feedback from @tempelmann, more testing (specially since as you mentioned is a little less surgical than needed) or a regression test added though but couldn't figure out how to do that and my mac mini is crashing after I installed BBEdit

tempelmann:
I'll build a test version and see if it fixes my particular issue.

tempelmann:
I can confirm that this fixes #257

Issue #276, opened 2023-07-31
Arm64 crash on JIT call to incorrect address


sebpop:
On arm64, JIT generated code crashes when executing a call to a function with an address + an incorrect offset `-0x10000000 = 2^26 (base 10)`.

The jit code contains several other calls to the same function address without the incorrect offset:

```
   0xffffa10a6298:      mov     x9, #0x67a8
   0xffffa10a629c:      movk    x9, #0xa90a, lsl #16
   0xffffa10a62a0:      movk    x9, #0xffff, lsl #32
   0xffffa10a62a4:      blr     x9               // correct address
[...]
   0xffffa10a67a4:      bl      0xffff990a67a8   // this is incorrectly offset by -0x10000000
                                                 // crashes when this call executes.
[...]
   0xffffa10a6840:      bl      0xffffa90a67a8   // correct address
[...]
      // several other calls with the correct address.
```

I do not know where the incorrect offset is coming from.

I have seen that when sljit generates a call to a function, sljit uses an offset `SLJIT_IMM()`.
For example the doc for `sljit_emit_icall` says:
https://github.com/PCRE2Project/pcre2/blob/master/src/sljit/sljitLir.h#L1540
```
/* Emit a C compiler (ABI) compatible function call.
Direct form: set src to SLJIT_IMM() and srcw to the address
```
Is there a way for SLJIT_IMM to be modified during the jit compilation, or is that an immutable immediate constant?
```
src/sljit/sljitLir.h:#define SLJIT_IMM          0x40
```
If SLJIT_IMM does not change, then the incorrect offset can only be part of srcw address computation.

zherczeg:
SLJIT_IMM is a constant, since it represents an immedate value. Probably the compiler tries to optimize the call with a bl instruction and doing it wrong during the executable code generation.

https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitNativeARM_64.c#L187

sebpop:
In `detect_jump_type` I see that `executable_offset` is directly used in an add then a sub: https://github.com/zherczeg/sljit/blob/master/sljit_src/sljitNativeARM_64.c#L201
and this without going through the usual macro as in the rest of the code.  Shouldn't this line be written:
```
target_addr = (sljit_uw)(SLJIT_ADD_EXEC_OFFSET(code, executable_offset) + jump->u.label->size);
```
[...]

PR #277, opened 2023-08-02
Fix incorrect branch offset calculation

ilink-splunk:
This now uses the same offset calculation as sljit_generate_code. Before, if you had an address that was 16 bytes off from the maximum, it would overflow the 26 bit signed int available to the "bl" instruction. In other words, it would choose "bl" when it should have chosen "blr". This typically caused a crash because the jump was to an invalid address.


ilink-splunk:
We've done quite a bit of testing on this internally, using code which scans the compiled regexes for out-of-bounds `bl` instructions. Without this fix, we get a surprisingly high failure rate. With the fix, we haven't detected any problems.

ilink-splunk:
Per discussion in https://github.com/PCRE2Project/pcre2/issues/276 i'll move this over to the JIT repo

ilink-splunk:
Per discussion in https://github.com/PCRE2Project/pcre2/issues/276 i'll move this over to the JIT repo

PR #281, opened 2023-08-10
jit: add LoongArch support

carenas:
While at it, make sure the list of files by sljit is updated after last import.


zherczeg (approved):
LGTM

PR #285, opened 2023-08-16
Fix and improve zig build

plajjan:
MAX_VARLOOKBEHIND now has to be set, so we do, which fixes builds on latest version of pcre2. The build.zig was created for some older version where this was not necessary. There are many many other variables that we should set, but this is something. Baby steps.

The flag handling has been revamped, mostly to prepare it for having varying flags set depending on input options and similar.

On Linux, Zig defaults to targetting the Musl C library. PCRE2, at least pcre2grep only works with GNU libc and so we now rewrite the default to make it work out of the box.

pcre2grep is now built when the -Dbuild_pcre2grep flag is specified.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -1,8 +1,32 @@
 const std = @import("std");
 
 pub fn build(b: *std.Build) !void {
-    const target = b.standardTargetOptions(.{});
+    const target_input = b.standardTargetOptions(.{});
     const optimize = b.standardOptimizeOption(.{});
+    const build_pcre2grep = b.option(bool, "build_pcre2grep", "") orelse false;
+
+    var target = target_input;
+
+    if (target_input.getOs().tag == .linux) {
+        // Zig defaults to musl libc on linux, but we want glibc as pcre2grep,
+        // pcre2test and perhaps other parts of pcre2 do not work with musl.
+        // This silently changes the target to glibc, which can be a little
+        // confusing if someone tries to explicitly target musl. It would be
+        // better if we could differentiate between target_input being
+        // explicitly set or if it has the defualt value.
+        // TODO: Add a way to check if a target is the default target or
+        //       explicitly set, if explicitly set, raise a warning and do not
+        //       set it to glibc so that the user sees the error
+        target.abi = .gnu;
+    }
+
+    var flags = std.ArrayList([]const u8).init(b.allocator);
+    defer flags.deinit();
+    flags.append("-std=c99") catch unreachable;
+    flags.append("-DHAVE_CONFIG_H") catch unreachable;
+    flags.append("-DPCRE2_CODE_UNIT_WIDTH=8") catch unreachable;
+    flags.append("-DPCRE2_STATIC") catch unreachable;
+    flags.append("-DMAX_VARLOOKBEHIND=255") catch unreachable;
 
     const copyFiles = b.addWriteFiles();
     copyFiles.addCopyFileToSource(.{ .path = "src/config.h.generic" }, "src/config.h");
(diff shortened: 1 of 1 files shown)


carenas:
as shown by CI (by using alpine), `pcre2grep` builds and works fine when linked against musl

carenas (changes requested):
needs more analysis

plajjan:
Ok. Not sure what you mean by analysis but FYI the current zig build on master is broken and this fixes it.

carenas:
> Not sure what you mean by analysis

the "fix" to define `MAX_VARLOOKBEHIND` is misplaced, this should be rebased on top of #286 one that is merged as the problem affects also all other builds that rely in the manual method of building.

it is not clear why you assumed that only glibc would work, but that might be also displaced (AFAIK PCRE2 build/works fine with musl and we even include it in CI, as well as other libc, so maybe some other config changes might be needed instead)

plajjan:
@carenas ok, thanks for clarifying. 

With regards to glibc vs musl, the build fails
```
zig build-exe pcre2grep Debug native: error: the following command failed with 2 compilation errors:
/home/kll/terastream/acton/dist/zig/zig build-exe -cflags -std=c99 -DHAVE_CONFIG_H -DPCRE2_CODE_UNIT_WIDTH=8 -DPCRE2_STATIC -DMAX_VARLOOKBEHIND=255 -- /home/kll/terastream/acton-deps/pcre2/src/pcre2grep.c /home/kll/terastream/acton-deps/pcre2/zig-cache/o/98a38a4938b000dd35845851627eb591/libpcre2.a -lc --cache-dir /home/kll/terastream/acton-deps/pcre2/zig-cache --global-cache-dir /home/kll/.cache/zig --name pcre2grep -I /home/kll/terastream/acton-deps/pcre2/src -I /home/kll/terastream/acton-deps/pcre2/zig-out/include --listen=- 
Build Summary: 4/7 steps succeeded; 1 failed (disable with --summary none)
install transitive failure
└─ install pcre2grep transitive failure
   └─ zig build-exe pcre2grep Debug native 2 errors
error: ld.lld: undefined symbol: sys_nerr
    note: referenced by pcre2grep.c:1174 (src/pcre2grep.c:1174)
    note:               /home/kll/terastream/acton-deps/pcre2/zig-cache/o/b1a1dea106f86d595e2347d222905220/pcre2grep.o:(strerror)
    note: did you mean: sys_nerr@GLIBC_2.12
    note: defined in: /lib/x86_64-linux-gnu/libc.so.6
error: ld.lld: undefined symbol: sys_errlist
    note: referenced by pcre2grep.c:1175 (src/pcre2grep.c:1175)
    note:               /home/kll/terastream/acton-deps/pcre2/zig-cache/o/b1a1dea106f86d595e2347d222905220/pcre2grep.o:(strerror)
    note: did you mean: sys_errlist@GLIBC_2.3
    note: defined in: /lib/x86_64-linux-gnu/libc.so.6
```

because musl doesn't have str_nerr, but this is conditioned in pcre2grep on HAVE_STRERROR so that's really what should have been defined. My bad.

zig has some support for dealing with config.h. It has notion of automake and cmake flavours, or no flavor at all. I'm not sure of the relation between src/config.h.in and src/config.h.generic and what's the best pick here.

I'm not sure I have the time to fix this up, so it might be better to just close this PR.

plajjan:
Ok, I did hack a little bit on using config.h.generic which seems to work. I'll see if I can push that work instead! :)

PhilipHazel:
As the zig config has subsequently been updated (01-Nov-2023) I am going to close this as obsolete. Please open a new PR if there is something still active.

PR #287, opened 2023-08-16
jit: rebase #259 on top of 3c93cdc

carenas:
Recent JIT does a backward incompatible change that removes CMOV so update code accordingly to fix arm64 builds.


zherczeg (approved):
LGTM.
Now the jit compiler supports some simd instructions, so part of this code could be rewritten. But that is a longer term plan.

PR #290, opened 2023-08-18
jit: Add LoongArch SX 128 bit SIMD Support.

lrzlin:
This patch enables SIMD support on LoongArch platform, already tested on real hardware and qemu.

### DIFF
--- a/src/pcre2_jit_simd_inc.h
+++ b/src/pcre2_jit_simd_inc.h
@@ -42,7 +42,8 @@ POSSIBILITY OF SUCH DAMAGE.
 #if !(defined SUPPORT_VALGRIND)
 
 #if ((defined SLJIT_CONFIG_X86 && SLJIT_CONFIG_X86) \
-     || (defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X))
+     || (defined SLJIT_CONFIG_S390X && SLJIT_CONFIG_S390X) \
+     || (defined SLJIT_CONFIG_LOONGARCH_64 && SLJIT_CONFIG_LOONGARCH_64))
 
 typedef enum {
   vector_compare_match1,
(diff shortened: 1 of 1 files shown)


src/pcre2_jit_simd_inc.h
```diff
@@ -1860,4 +1861,515 @@ if (common->match_end_ptr != 0)
...
+sljit_s32 cmp1a_ind = 4;
+sljit_s32 cmp1b_ind = 5;
+sljit_s32 cmp2a_ind = 6;
+sljit_s32 cmp2b_ind = 7;
+struct sljit_label *start;
+#if defined SUPPORT_UNICODE && PCRE2_CODE_UNIT_WIDTH != 32
+struct sljit_label *restart;
+#endif
+struct sljit_jump *jump[2];
+int i;
```
carenas, line 2196: unused variable

src/pcre2_jit_simd_inc.h
```diff
@@ -1860,4 +1861,515 @@ if (common->match_end_ptr != 0)
...
+jump_list *not_found = NULL;
+vector_compare_type compare_type = vector_compare_match1;
+sljit_s32 tmp1_reg_ind = sljit_get_register_index(SLJIT_INT_REGISTER, TMP1);
+sljit_s32 str_ptr_reg_ind = sljit_get_register_index(SLJIT_INT_REGISTER, STR_PTR);
+sljit_s32 data_ind = 0;
+sljit_s32 tmp_ind = 1;
+sljit_s32 cmp1_ind = 2;
+sljit_s32 cmp2_ind = 3;
+sljit_u32 bit = 0;
+int i;
```
carenas, line 2085: unused variable

src/pcre2_jit_simd_inc.h
```diff
@@ -1860,4 +1861,515 @@ if (common->match_end_ptr != 0)
...
+push_inst(compiler, VOR_V | VD(dst_ind) | VJ(tmp_ind) | VK(dst_ind));
+return;
+}
+
+#define JIT_HAS_FAST_FORWARD_CHAR_SIMD 1
+
+static void fast_forward_char_simd(compiler_common *common, PCRE2_UCHAR char1, PCRE2_UCHAR char2, sljit_s32 offset)
+{
+DEFINE_COMPILER;
+sljit_u8 instruction[8];
```
carenas, line 1936: unused variable

src/pcre2_jit_simd_inc.h
```diff
@@ -1860,4 +1861,515 @@ if (common->match_end_ptr != 0)
...
+struct sljit_jump *partial_quit[2];
+vector_compare_type compare_type = vector_compare_match1;
+sljit_s32 tmp1_reg_ind = sljit_get_register_index(SLJIT_INT_REGISTER, TMP1);
+sljit_s32 str_ptr_reg_ind = sljit_get_register_index(SLJIT_INT_REGISTER, STR_PTR);
+sljit_s32 data_ind = 0;
+sljit_s32 tmp_ind = 1;
+sljit_s32 cmp1_ind = 2;
+sljit_s32 cmp2_ind = 3;
+sljit_u32 bit = 0;
+int i;
```
carenas, line 1951: unused variable

src/pcre2_jit_simd_inc.h
```diff
@@ -1860,4 +1861,515 @@ if (common->match_end_ptr != 0)
...
+
+/* VSEQ.B/H/W vd, vj, vk */
+push_inst(compiler, VSEQ | VD(tmp_ind) | VJ(tmp_ind) | VK(cmp2_ind));
+
+/* VOR vd, vj, vk */
+push_inst(compiler, VOR_V | VD(dst_ind) | VJ(tmp_ind) | VK(dst_ind));
+return;
+}
+
+#define JIT_HAS_FAST_FORWARD_CHAR_SIMD 1
```
carenas, line 1931: LSX extension is optional, so it will be better to expand sljit to call `cpucfg` looking for the correct bit instead

src/pcre2_jit_simd_inc.h
```diff
@@ -1860,4 +1861,515 @@ if (common->match_end_ptr != 0)
...
+  add_jump(compiler, &common->failed_match, CMP(SLJIT_GREATER_EQUAL, STR_PTR, 0, STR_END, 0));
+  OP1(SLJIT_MOV, TMP2, 0, STR_PTR, 0);
+  JUMPTO(SLJIT_JUMP, restart);
+
+  JUMPHERE(quit);
+  }
+#endif
+}
+
+#define JIT_HAS_FAST_REQUESTED_CHAR_SIMD 1
```
carenas, line 2069: ditto `cpucfg`

src/pcre2_jit_simd_inc.h
```diff
@@ -1860,4 +1861,515 @@ if (common->match_end_ptr != 0)
...
+push_inst(compiler, CTZ_W | RD_V(tmp1_reg_ind) | RJ_V(tmp1_reg_ind));
+
+OP2(SLJIT_ADD, TMP1, 0, TMP1, 0, STR_PTR, 0);
+add_jump(compiler, &not_found, CMP(SLJIT_GREATER_EQUAL, TMP1, 0, STR_END, 0));
+
+OP1(SLJIT_MOV, STR_PTR, 0, TMP3, 0);
+return not_found;
+}
+
+#define JIT_HAS_FAST_FORWARD_CHAR_PAIR_SIMD 1
```
carenas, line 2169: `sljit_has_cpu_feature(SLJIT_HAS_SIMD)`

carenas (changes requested):
Alternatively, recent `sljit` has support for SIMD (currently unused for PCRE) that could be used instead to implement this:

```
sljit_emit_simd_mov()
sljit_emit_simd_replicate()
sljit_emit_simd_lane_mov()
sljit_emit_simd_lane_replicate()
```

src/pcre2_jit_simd_inc.h
```diff
@@ -1860,4 +1861,515 @@ if (common->match_end_ptr != 0)
...
+if (compare_type != vector_compare_match2)
+  {
+  if (compare_type == vector_compare_match1i)
+    {
+    /* VOR.V vd, vj, vk */
+    push_inst(compiler, VOR_V | VD(dst_ind) | VJ(cmp2_ind) | VK(dst_ind));
+    }
+
+  /* VSEQ.B/H/W vd, vj, vk */
+  push_inst(compiler, VSEQ | VD(dst_ind) | VJ(dst_ind) | VK(cmp1_ind) );
```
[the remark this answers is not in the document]
lrzlin: Deleted

src/pcre2_jit_simd_inc.h
lrzlin: Deleted

src/pcre2_jit_simd_inc.h
lrzlin: Deleted

src/pcre2_jit_simd_inc.h
lrzlin: Deleted

src/pcre2_jit_simd_inc.h
lrzlin: Deleted

src/pcre2_jit_simd_inc.h
lrzlin: According to [this pr of openssl](https://github.com/openssl/openssl/pull/21509#issue-1815069650), we need to introduce `getauxval` to avoid the case that hardware supports LSX but the kernel does not. I'll request this change in sljit repo soon.

lrzlin:
@carenas Thanks for your advice! I have not noticed that the sljit has already introduced SIMD support, in that case I'll add support for LoongArch SX in sljit repo for future changes.

carenas:
@lrzlin do you have an update on at least adding `sljit_has_cpu_feature(SLJIT_HAS_SIMD)` to sljit so it could be used here to avoid unnecessary crashes with old software?

lrzlin:
> @lrzlin do you have an update on at least adding `sljit_has_cpu_feature(SLJIT_HAS_SIMD)` to sljit so it could be used here to avoid unnecessary crashes with old software?

Added here, however this change will cause the sljit_test in sljit repo failed (because SLJIT_HAS_SIMD implies SIMD
 support), should I commit this change now or at the time that I have finished all sljit SIMD functions?

src/pcre2_jit_simd_inc.h
lrzlin: Added

src/pcre2_jit_simd_inc.h
lrzlin: Added

carenas:
> however this change will cause the sljit_test in sljit repo failed

usually to avoid those issues, and while the code is still in development, it is just #ifdef out with the test skipping (see test92 for atomics, which are also not yet used by PCRE).

IMHO and since you already have a working "legacy" SIMD implementation it will be better to get this version out first that way.

eitherway, all sljit changes go first to that repository and they are merged back from there.

lrzlin:
> eitherway, all sljit changes go first to that repository and they are merged back from there.

In that case, I think the best way is putting the LSX detection here, until I have finished all LoongArch sljit SIMD support. Also it would convinent if we need to drop the legacy SIMD support and switch to sljit SIMD implement.

zherczeg:
I am ok with this code. I cannot maintain it though. So if no more style issues, I can land it.
Moving to sljit simd will take a longer time.

lrzlin:
> I am ok with this code. I cannot maintain it though. So if no more style issues, I can land it. Moving to sljit simd will take a longer time.

Thanks for that! I think there is no more style issues. Also I'll maintain the LoongArch SIMD code if there are any issues.

carenas:
It is missing an autoconf/cmake detection for getauxval() and having the code #ifdef around it; presume that the fallback defauting to 1 would be ok since all available hardware and the latest release of qemu have LSX support.

carenas (changes requested):

src/pcre2_jit_simd_inc.h
```diff
@@ -1860,4 +1861,516 @@ if (common->match_end_ptr != 0)
 
 #endif /* SLJIT_CONFIG_S390X */
 
+#if (defined SLJIT_CONFIG_LOONGARCH_64 && SLJIT_CONFIG_LOONGARCH_64)
+
+/* Using HWCAP for detecting whether LSX is available */
+#include <sys/auxv.h>
+#define LOONGARCH_HWCAP_LSX  (1 << 4)
+#define HAS_LSX_SUPPORT getauxval(AT_HWCAP) & LOONGARCH_HWCAP_LSX
```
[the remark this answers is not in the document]
lrzlin: Done

lrzlin:
> It is missing an autoconf/cmake detection for getauxval() and having the code #ifdef around it; presume that the fallback defauting to 1 would be ok since all available hardware and the latest release of qemu have LSX support.

@carenas Is the `#ifdef __linux__` guard okay? AFAIK, the Linux `getauxval` syscall and glibc wrapper are available since day one for LoongArch. Also, some embedded LoongArch SOCs (such as Loongson 2K500) do not support LSX extension, so it would be safer to disable LSX if the detection fails.

carenas:
Interesting, would require changes later if uclibc (or other libc except musl and bionic) are supported by Linux in Loongarch but should be fine for now.

carenas (approved):

zherczeg (approved):
LGTM

PR #292, opened 2023-08-30
Add a TH (title) macro to pcre2demo.3

MatthewVernon:
per groff_man(7):

> A man page should contain exactly one .TH call at or near the
> beginning of the file, prior to any other macro calls.

This manual page didn't (and as a result `man --warnings` emits a warning about the first `.SH` macro).

Make `PrepareRelease` add such a header to pcre2demo.3 along with a comment noting that pcre2demo.3 is automatically generated (to avoid any further confusion on that front).

This necessitates teaching `132html` to know about that warning (which we don't want to have appearing in the HTML output).

Also, update the EX and EE macros to reflect the current content of `/usr/share/groff/1.22.4/tmac/an-ext.tmac`.


PhilipHazel:
Thanks for looking at this. However, pcre2demo.3 is not manually maintained; it is created by the script called PrepareRelease, to ensure that it always contains the current pcre2demo.c code. If you don't want to make another pull request, I can copy your patch to PrepareRelease. It's so long ago that that script was created that I cannot remember details about the EX and EE macros. I'm happy to update them if it seems reasonable.

MatthewVernon:
Oh, since it was committed in the tree I hadn't considered it was essentially a build artifact.
I'll make a patch to PrepareRelease (and presumably also pcre2demo.3 to reflect the new output?), and maybe also stick a comment at the top to say where it comes from :)

PhilipHazel:
Just patch PrepareRelease; running "./PrepareRelease doc" should make a new pcre2demo.3. I guess it's more of a packaging artifact than a build one. It's in the repo so that those who clone it get the full documentation - same goes for the HTML docs, which are also made by PrepareRelease. And yes, you are right, there should be a comment!

MatthewVernon:
I've now done this (see revised commit & comment).

PhilipHazel:
I've merged, but there is a problem. The new man page shows everything with an indent, whereas the old one does not. This means that many lines overflow in an 80-character window, making it all hard to read. I will investigate, but I think it might be a consequence of adding the TH macro.

MatthewVernon:
You can put that back by adding
```
.RS -7
```
before the `.EX`; this effectively undoes the default 7 or 7.2n indentation that man pages have. I could add that as a new PR if you like?

PhilipHazel:
No need. I will do it.

PR #297, opened 2023-09-28
Spaces are not allowed within \u{} in BSUX mode

carenas:
Avoid regressions by matching the observed behaviour of Javascript when the PCRE2_EXTENDED_ALT_BSUX option is in effect, and make sure to document the exception.

### DIFF
--- a/doc/pcre2pattern.3
+++ b/doc/pcre2pattern.3
@@ -425,8 +425,9 @@ interpreted as a literal "u" character.
 .P
 PCRE2_EXTRA_ALT_BSUX has the same effect as PCRE2_ALT_BSUX and, in addition,
 \eu{hhh..} is recognized as the character specified by hexadecimal code point.
-There may be any number of hexadecimal digits. This syntax is from ECMAScript
-6.
+There may be any number of hexadecimal digits, but unlike other places that
+also use brackets, spaces are not allowed and would result in the string being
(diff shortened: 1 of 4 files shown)


PhilipHazel:
Hmm. Yes, this is a moot point. I was making PCRE2 consistent in always allowing spaces for data in braces (like Perl) but since this is not a Perl-compatible feature, I guess we should go along with Javascript.

PhilipHazel:
Oh dear. We now have another issue. The sequence \u{ 12} (when in BSUX mode) should be treated as a literal, and indeed it is in 10.42. But now, because spaces are allowed in quantifiers, it is treated as u{12}, because the \u is treated as a literal "u". This is going to be tedious to fix, but I guess I'll have to do it. The law of unintended consequences strikes again.

carenas:
Since in BSUX mode any `\x` character represents itself then the effect is indeed wider and so it might be worth disabling the space cleaning even for the quantifier to solve it, at least when in BSUX mode.

Note that Javascript (as well as POSIX) does not recognize `x{,m}` as valid either.

PhilipHazel:
I have fixed the \u{  12}  case so it is now compatible. The fix is in
HEAD. I added a note about {,m} to the docs; there will always be
differences between engines, but I think Perl compatibility has to come
first. BSUX mode is documented to affect only some very specific things.

Issue #261, opened 2023-06-05
Different PCRE result /w and /wo JIT enabled


mvorisek:
see https://github.com/php/php-src/issues/11374

repro in C code: https://github.com/php/php-src/issues/11374#issuecomment-1577098261

/cc @iluuu1994 @Wirone

zherczeg:
Summary:
Pattern: `(?<all>(?:(?:a(?&all))|(b))(c?))`
Input: `aabc`

It looks like the code works as intended, but things might be changed. As far as I remember, recursive calls are considered as a "subroutine calls", which uses brackets as private variables. Since `(b)` is part of the recursion, its match is not visible outside. I remember I used this fact for various optimizations, e.g. when atomic groups are failed to match. Is there any change of recursions since then?

mvorisek:
I agree with this JITted definition as returning "subroutine calls" matches (when not matched in the "main call") is completely useless. So non-JITted version should be probably fixed.

Also JITted/non-JITted results should be fuzzed and compared to prevent such inconsistencies in the future.

zherczeg:
It seems perl agrees with me:
```
$str = "aabc";
if ($str =~ /(?<all>(?:(?:a(?&all))|(b))(c?))/) {
   print "'$1' '$2' '$3'\n";
}
```
$2 and $3 are empty

zherczeg:
I completely misunderstood this request, I thought the jit code is wrong.

PhilipHazel:
This is an area where PCRE and Perl have differed and not differed over the years. PCRE had recursion (= subroutine calls) before Perl did. There is a section in pcre2compat.3 entitled "Differences in recursion processing between PCRE2 and Perl", but Perl is a moving target, so it may no longer be accurate.

You are right in thinking that JIT and non-JIT should give the same answers. At least, most of the time. The majority of the standard tests are run with and without JIT to ensure that this is the case. Clearly, none of the existing tests show up this anomaly. 

Since JIT agrees with Perl, I suppose it is the interpreter (non-JIT match) that should change. I have not looked at the interpreter code, but I suspect that it is not trivial to arrange to unset (or possibly unwind) captures from inside a subroutine call.

[...]

PR #300, opened 2023-10-02
match: also restore originally unset entries in recursion

carenas:
Fixes: #261


PhilipHazel:
Thank you for figuring out the solution to this issue. I have been poring over the code, trying to remember how it works, and failing to spot what was going on, mainly because I was looking too hard at the _start_ of a recursion rather than the end. Doh!

PR #301, opened 2023-10-04
Allow restricting [:xdigit:] to ASCII for POSIX compatibility

carenas:
Just like in the case of [:digit:], and as suggested in [UTS#18](https://unicode.org/reports/tr18/#Compatibility_Properties)

### DIFF
--- a/testdata/testinput5
+++ b/testdata/testinput5
@@ -1234,6 +1234,8 @@
 
 /[[:xdigit:]]/B,ucp
 
+/[[:xdigit:]]/B,ucp,ascii_digit
+
 # Unicode properties for \b and \B
 
 /\b...\B/utf,ucp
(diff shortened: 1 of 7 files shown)


PhilipHazel:
Thanks for reviewing my previous commits and finding the typos. This patch looks good to me. I think it makes sense to use PCRE2_EXTRA_ASCII_DIGIT for both decimal and hex digits. I don't think there's much likelihood of a use case where only one needs to  be forced to ASCII. There needs to be a minor update to pcre2api in the description of PCRE2_EXTRA_ASCII_DIGIT. I am happy to do that if you like, after this PR is undrafted and merged (but I'm about to go away for the weekend).

testdata/testinput5
```diff
@@ -2445,6 +2447,22 @@
...
+    d\x{ff10}
+    \x{ff26}8
+\= Expect no match
+    8g\=no_jit
+
+/^[[:xdigit:]]+$/utf,ucp,ascii_digit
+    f0
+    1A
+\= Expect no match
+    d\x{xfff0}
```
carenas, line 2462: typo on hex value

Issue #302, opened 2023-10-05
heap-buffer-overflow in PHP preg_replace on Mac M1 in Linux


morrisonlevi:
I have struggled to create a _simple_ reproducer for this, but I do have a Dockerfile below. Here is the sanitizer output:

```
=================================================================
==7558==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60d00011e40f at pc 0xaaaac6c6aaa8 bp 0xffffc13ae370 sp 0xffffc13ae368
READ of size 16 at 0x60d00011e40f thread T0
=================================================================
==7557==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60600007a87f at pc 0xaaaac5c5aaa8 bp 0xfffff3aea780 sp 0xfffff3aea778
READ of size 16 at 0x60600007a87f thread T0
    #0 0xaaaac6c6aaa4 in ffcps_0 /usr/local/src/php/ext/pcre/pcre2lib/pcre2_jit_neon_inc.h:267:22
    #1 0xffffa9914284  (<unknown module>)
    #2 0xaaaac6c28ffc in php_pcre2_jit_match /usr/local/src/php/ext/pcre/pcre2lib/pcre2_jit_match.c:165:8
    #3 0xaaaac6d25bac in php_pcre_replace_impl /usr/local/src/php/ext/pcre/php_pcre.c:1629:11
    #4 0xaaaac6d2741c in php_pcre_replace /usr/local/src/php/ext/pcre/php_pcre.c:1571:11
    #5 0xaaaac6d2741c in php_replace_in_subject /usr/local/src/php/ext/pcre/php_pcre.c:2141:12
    #6 0xaaaac6d2741c in preg_replace_common /usr/local/src/php/ext/pcre/php_pcre.c:2279:12
    #7 0xaaaac7820838 in ZEND_DO_ICALL_SPEC_RETVAL_USED_HANDLER /usr/local/src/php/Zend/zend_vm_execute.h:1337:2
    #8 0xaaaac772b0cc in execute_ex /usr/local/src/php/Zend/zend_vm_execute.h:56987:7
    #9 0xaaaac772b7ac in zend_execute /usr/local/src/php/Zend/zend_vm_execute.h:61584:2
    #10 0xaaaac76adcbc in zend_execute_scripts /usr/local/src/php/Zend/zend.c:1876:4
[...]

PR #306, opened 2023-10-06
jit: disable ASAN for arm64 SIMD helper

carenas:
Technically a false positive, but it might be easier to prevent it by going at the source this way.

Fixes: #302


zherczeg:
What is the effect? It disables something at compile time?

carenas:
Yes; ASAN works by adding "interceptor" routines in the code at build time, so this tells it not to do that for this function.

PR #307, opened 2023-10-10
Allow updating PCRE2_EXTRA_ASCII_DIGIT by (?aP) in pattern

carenas:
For consistency with PCRE2_EXTRA_ASCII_POSIX, allow the inline use of (?a) to affect the definition of [:digit:] and [:xdigit:]


carenas:
> I guess it will have to be (?aT)

if tried with (?aG) originally but it would seem more logical to reuse the one that applies to all POSIX classes instead as shown by the proposed changes, since that behaviour had to be modified anyway.

PhilipHazel:
I am confused. Setting (?aP) affects all POSIX classes, including digit and xdigit. When PCRE2_EXTRA_ASCII_POSIX (aka (?aP)) is set, PCRE2_EXTRA_ASCII_DIGIT has no effect. The use case for ...DIGIT is when you just want to restrain digits but let other classes be extended. What is lacking is a (?a setting that just sets/unsets ...DIGIT.

src/pcre2_compile.c
```diff
@@ -4233,7 +4233,7 @@ while (ptr < ptrend)
                 }
               if (*ptr == CHAR_P)
                 {
-                *xoptset |= PCRE2_EXTRA_ASCII_POSIX;
+                *xoptset |= (PCRE2_EXTRA_ASCII_POSIX|PCRE2_EXTRA_ASCII_DIGIT);
```
PhilipHazel, line 4236: I don't see why you want to set both flags here, since ASCII_POSIX affects all POSIX classes (thanks for adding emphasis to the docs, btw).

carenas: because without it, (?aP) was inconsistent as shown by:

```
  re> /(?-aP:[[:digit:]])[[:digit:]]/B,utf,ucp,ascii_digit
        Bra
        Bra
        [0-9]
        Ket
        [0-9]
        Ket
        End
```

PR #312, opened 2023-10-22
test: avoid Undefined Behaviour because of NULL pointer arithmetic in test17

carenas:
Not a regression, as it seems to be included with the 10.40 release.

Resolves the following error when building with [UBSAN](https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) in a recent compiler:

```
Test 17: JIT-specific features when JIT is available
src/pcre2_jit_match.c:124:25: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/pcre2_jit_match.c:124:25 in 
src/pcre2_jit_match.c:126:25: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/pcre2_jit_match.c:126:25 in 
```

### DIFF
--- a/testdata/testinput17
+++ b/testdata/testinput17
@@ -304,7 +304,7 @@
 /[aCz]/mg,firstline,newline=lf
     match\nmatch
 
-//jitfast
+//jit
     \=null_subject
 
 # End of testinput17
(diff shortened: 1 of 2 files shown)


zherczeg (approved):
LGTM

PR #315, opened 2023-11-01
update to latest zig

jimying:
Adapt to the latest zig ( 0.12.0-dev)


carenas:
I know zig is not stable, but care to explain why is a good idea to do this change now instead of when 0.12 gets eventually released?

also, could the known problems this have be fixed as well?, there are some details in the stale #285, but noticed that is also missing JIT (not even as an option).

PR #319, opened 2023-11-09
Implement vreverse in jit

zherczeg:
I have started working on vreverse. Unfortunately the task is very hard, because it adds another layer to the already complex bracket/assertion matching. Each new layer kind of duplicates the complexity, so I have no idea how many bugs I introduce with this patch. To tell the truth it was a long time ago when I developed the original code, and forgot many things since then.

The tests are passing now, although @PhilipHazel mentioned the vreverse tests are disabled for jit. I didn't see anything which disables them. Is there something I miss?

The patch also contains some unicode fixes, they can be moved to another patch if needed.

### DIFF
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -338,6 +338,12 @@ typedef struct recurse_backtrack {
   BOOL inlined_pattern;
 } recurse_backtrack;
 
+typedef struct vreverse_backtrack {
+  backtrack_common common;
+  /* Return to the matching path. */
+  struct sljit_label *matchingpath;
+} vreverse_backtrack;
+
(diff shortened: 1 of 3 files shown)


PhilipHazel:
There was no need to cut out JIT for vreverse tests because of the automatic fallback to the interpreter. The cut out tests are for the changes to \w and [:xdigit:]. I will merge and run tests on this patch shortly. That then leaves only the \w and [:xdigit:] changes that need an update to JIT.

PhilipHazel:
As you were. I see that this patch also contains the updates for \w and [:xdigit:]. Thanks.

zherczeg:
I am thinking about optimizing them further

PR #339, opened 2023-11-16
ci: add dev build to serve as canary

carenas:
Run a build with debugging enabled, link-size of 4 and sanitizers.


PhilipHazel:
Detecting things early is always good. I would have picked up the link size problem when running maint/ManyConfigTests, which I do before each release, and I always compile with maximum warnings enabled.

PR #342, opened 2023-11-17
Sanity checks for ctype functions

addisoncrump:
This addresses some rare crash conditions where PCRE2_UCHAR is larger than unsigned char. Additionally, ensures that arguments to ctype functions are not sign-extended.


carenas:
unless strictly needed, and since we follow a very linear development process (code was originally maintained in subversion) it is usually better to squash all commits together; specially when they have issues that are fixed later in the series.

addisoncrump:
Sure :slightly_smiling_face: I'll remember to squash the commits next time.

PR #344, opened 2023-11-22
ci: add dev job using clang

carenas:
A similar job to the one that was added originally and that used `gcc`.

The "fix" for the "Undefined Behaviour while shifting framesize" is a little more involved that I would like, but done this way to keep mostly the same logic.

### DIFF
--- a/maint/ManyConfigTests
+++ b/maint/ManyConfigTests
@@ -289,9 +289,7 @@ fi
 # throw up warnings that are not detected with -O0. Then run a second test with
 # -fsanitize=address, which also may throw up new warnings as well as checking
 # things at runtime. Finally, run another test using -fsanitize=undefined
-# -std-gnu99 to check for runtime actions that are not well defined. However,
-# we also use -fno-sanitize=shift to avoid warnings for shifts of negative
-# numbers, which occur in src/pcre2_jit_compile.c.
+# -std-gnu99 to check for runtime actions that are not well defined.
 
 if [ $ISGCC -ne 0 -a $usemain -ne 0 ]; then
   echo "---------- Maximally configured test with -O2 ----------"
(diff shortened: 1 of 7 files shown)


PhilipHazel:
The JIT code needs reviewing by @zherczeg.  As far as the non-JIT changes go, I am happy.

zherczeg:
Are these compiler warnings? I am ok to change them.

maint/ManyConfigTests
```diff
@@ -311,8 +309,8 @@ if [ $ISGCC -ne 0 -a $usemain -ne 0 ]; then
   fi
 # This also seems to be the case for sanitize undefined.
   if [ $useusan -ne 0 ]; then
-    echo "------- Maximally configured test with -fsanitize=undefined -fno-sanitize=shift -fno-sanitize=alignment -std=gnu99 -------"
-    CFLAGS="$OFLAGS $SAVECFLAGS -no-pie -fno-PIE -fsanitize=undefined -fno-sanitize=shift -fno-sanitize=alignment -std=gnu99"
+    echo "------- Maximally configured test with -fsanitize=undefined -fno-sanitize=alignment -std=gnu99 -------"
```
carenas, line 312: I couldn't find any explanation on why `-fno-sanitize=alignment` was originally added, but I don't think it might be needed anymore (sljit had several UB fixes in the last years).

src/pcre2_jit_compile.c
```diff
@@ -11206,10 +11206,15 @@ cc += 1 + LINK_SIZE;
 
 if (opcode == OP_ONCE)
   {
+  int data;
+  int framesize = BACKTRACK_AS(bracket_backtrack)->u.framesize;
+
+  SLJIT_ASSERT(SHRT_MIN <= framesize && framesize < SHRT_MAX/2);
   /* We temporarily encode the needs_control_head in the lowest bit.
-     Note: on the target architectures of SLJIT the ((x << 1) >> 1) returns
```
carenas, line 11210: while there is documentation to explain why `x << 1` is done here as "temporary", it is not clear where this is used and neither why it is ok to override "framesize" this way.

carenas:
Most of the changes are (by design) very safe as they don't change logic, but just make the implications of it more explicit.

I was hoping though that the initial version would be a bit less "hacky" by changing the logic and documenting better the code used on it.

Left a couple of "comments" to better clarify where feedback would be useful, but agree that it would likely be better than the current version so maybe they can be addressed later.

src/pcre2_jit_compile.c
```diff
@@ -11206,10 +11206,15 @@ cc += 1 + LINK_SIZE;
...
-     the same value for small signed numbers (including negative numbers). */
-  BACKTRACK_AS(bracket_backtrack)->u.framesize = (int)((unsigned)BACKTRACK_AS(bracket_backtrack)->u.framesize << 1) | (needs_control_head ? 1 : 0);
+     The real value should be short enough for this operation to work
+     without triggering Undefined Behaviour. */
+  data = (int)((short)((unsigned short)framesize << 1) | (needs_control_head ? 1 : 0));
```
carenas, line 11216: why is framesize overloaded this way?, wouldn't it be better to put "needs_control_head" on some independent flag member?

and if it is because we are not using the high bits of framesize anyway, and already using the sign bit as some sort of flag, why not use bit 30 instead of shifting a negative value which is UB for most versions of C?

PhilipHazel:
Again, I'm OK with this. Awaiting @zherczeg review.

PhilipHazel:
What is the status of this, please?

carenas:
It mostly just adds the job to CI with the "fixes" in the code, so it wouldn't fail, not changing logic.

Zoltan didn't have objections to the original code, and has mostly just been rebased so it should be fine to merge IMHO.

PR #356, opened 2023-12-08
Add vzeroupper support for x86

chen-hu-97:
Refer to "15.3 MIXING AVX CODE WITH SSE CODE" Intel Software Optimization manual, "if software inter-mixes AVX and SSE instructions without using VZEROUPPER properly, it can experience an AVX/SSE transition penalty."

This patch add support for vzeroupper so JIT can emit this instruction and place it in correct place.


carenas:
This seems to be something that should be added to [sljit](/zherczeg/sljit.git) first and therefore included in the next import of that code instead of here.

chen-hu-97:
Thanks @carenas for the info. Let me close this one and try to create PR in [sljit](https://github.com/zherczeg/sljit.git) at first.

PR #366, opened 2023-12-30
cmake: Install cmake files in a more common path

diizzyy:
Install cmake files in prefix/lib/cmake/pcre2 rather than prefix/cmake


CMakeLists.txt
```diff
@@ -1124,7 +1124,7 @@ configure_file(${PCRE2_CONFIG_IN} ${PCRE2_CONFIG_OUT} @ONLY)
 set(PCRE2_CONFIG_VERSION_IN  ${CMAKE_CURRENT_SOURCE_DIR}/cmake/pcre2-config-version.cmake.in)
 set(PCRE2_CONFIG_VERSION_OUT ${CMAKE_CURRENT_BINARY_DIR}/cmake/pcre2-config-version.cmake)
 configure_file(${PCRE2_CONFIG_VERSION_IN} ${PCRE2_CONFIG_VERSION_OUT} @ONLY)
-install(FILES ${PCRE2_CONFIG_OUT} ${PCRE2_CONFIG_VERSION_OUT} DESTINATION cmake)
+install(FILES ${PCRE2_CONFIG_OUT} ${PCRE2_CONFIG_VERSION_OUT} DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/pcre2)
```
carenas, line 1127: there used to be  time when CMAKE_INSTALL_LIBDIR was lib64 for 64bit builds, and I pressume avoiding that variance made using a "bitness" agnostic directory.

I also wonder how that would work in Windows, where the path is usually just "cmake" (usually with some funny case ifgnoring but preserving variant).

could we check with the rest of the "cmake" communitiy around #115 for advice before merging this backward incompatible change?

carenas (commented):
there was an unmerged proposal before to chnge this only for POSIX systems (where it makes more sense as well, as it sallows correctlu detecting the right bitness for PCRE2 in multilib systems)

CMakeLists.txt
diizzyy: https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure suggests that this is viable on both Windows and Unix

carenas: agree it is "viable", the pont I was making was how to minimize "backward compatibility" issues

ex: currently it "might" fail to allow locatiing the cmake files if ${CMAKE_INSTALL_LIBDIR} is (arguibly incorrectly) set to lib64 and the system is Windows.  Making this change ONLY if not Windows would "solve" that and IMHO was simple enough to do.

diizzyy: I'm not sure what you really want as several have suggested the same thing by now
https://github.com/PCRE2Project/pcre2/pull/260/files#r1225428249

carenas: funny enough you linked to my "ask", which was "what can we ut in the release notes to make sure whoever might be affected by this change fixes their setup".

I don't maintan any software that uses pcre2 and that would be using this code, so I am not the best person to provide that.

Indeed, I dont't even know how to create an application that could be affected, but assume one running Windows that uses cmake and linking with a PCRE2 that provides this file (meaning it was installed by running `cmake install` instead of the more common way to get PCRE2 installed which is calling `make install` as "root" while using a POSIX compatible shell, hence maybe mingw64 bash.

diizzyy: Not sure if this answers your question but vcpkg does this too
https://github.com/microsoft/vcpkg/blob/master/ports/pcre2/portfile.cmake#L54

carenas: some of the patches we are discosing were tested with vcpkg but had to be reverted before 10.34RC1 because they broke autotools bootstrapping.

it doesn't answer though how to document the hopefully small backward incompatibility that they introduce in their current form.

alexreinking:
As long as you're making a change like this, you might as well make the path configurable via a cache variable like `PCRE2_INSTALL_CMAKEDIR`. For example (sketch):

```cmake
include(GNUInstallDirs)

set(PCRE2_INSTALL_CMAKEDIR "${CMAKE_INSTALL_LIBDIR}/cmake/pcre2"
    CACHE STRING "Relative path to PCRE2's CMake config files from the installation root")

install(
    FILES ${PCRE2_CONFIG_OUT} ${PCRE2_CONFIG_VERSION_OUT}
    DESTINATION "${PCRE2_INSTALL_CMAKEDIR}"
)
```

I go into a lot more detail on my blog here: https://alexreinking.com/blog/building-a-dual-shared-and-static-library-with-cmake.html

zherczeg:
Is this still valid/needed?

alexreinking:
> Is this still valid/needed?

Yes, it looks like something along the lines of my suggestion above would be a good one.

https://github.com/PCRE2Project/pcre2/blob/db95416ee87a7abc49b7856f806a1b48f901756d/CMakeLists.txt#L1135

NWilson:
I am curious what you use-case is for "make install". Personally, it would never occur to me to use it for a library.

I mean, for a binary perhaps: you might want to check it out, build, and install it to `~/bin` or `/usr/local/bin`, but a library? Who installs a library like that? Libs will come packaged by the OS (on Linux) or bundled by an application that builds in the library using its own toolchain.

I have never done a "make install" for a library, so I just don't understand the purpose of it.

In principle though, I think we'd be happy to go along with any suggestions users have.

PhilipHazel:
@NWilson - More History! When PCRE was new, the _only_  way to install it was "make install". It was only after it started to get widely used that distros started packaging it. So I guess you have never wanted to install an uncommon library. :-)

NWilson:
If I were writing an Apache httpd plugin using PCRE (for example), and there wasn't a PCRE2 bundled by the OS, then I'd statically link it in.

That probably comes from my background shipping commercial software on Unix platforms. The apps are always shipped with dependencies included, so you don't get support requests from users on different Linux distros. I can see that the open-source way probably would instruct users to fetch and install the library globally, outside of the OS packaging system.

Things have moved on maybe. These days, open-source packages for Linux seem to be either bundled by distros, or the upstream project provides an apt/RPM feed of their own, or they provide a build with dependencies statically linked - precisely to avoid the hassle of incompatible library versions on the user's system.

carenas:
> If I were writing an Apache httpd plugin using PCRE (for example), and there wasn't a PCRE2 bundled by the OS, then I'd statically link it in.

That is very common in Windows but not so in Linux or the BSD.

Note that the OP is the FreeBSD packager for PCRE2, and they had (IMHO) mistakenly moved to use cmake instead of autotools to deploy those system wide.

PhilipHazel:
Early versions of PCRE were indeed bundled with Exim because PCRE was then an unknown library, but I stopped doing that once other software started to use PCRE and needed a properly installed separate library.

diizzyy:
> Is this still valid/needed?
Can also confirm that it is

NWilson:
I've added the excellent suggestion from @alexreinking to allow clients to override the new `lib/cmake/pcre2` location (for example, to go back to the old directory!).

This should address @carenas's concerns, by allowing affected users to go back to how it used to be.

I am satisfied with this change, and I think we should merge it for the 10.45-RC1. We can always make changes if there is any feedback.

Thank you @diizzyy for raising this!

NWilson (approved):

PR #381, opened 2024-02-16
Conditionally enable differential fuzzing

addisoncrump:
Unfortunately, there was an unexpected blocker on https://github.com/google/oss-fuzz/pull/11195 -- differential fuzzing support finding aborts made the build unacceptable for OSS-Fuzz, as these issues appear quite quickly.

This PR fixes it by making differential fuzzing conditional, while keeping fuzzing enabled for JIT and for the interpreter.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -395,6 +395,12 @@ AC_ARG_ENABLE(fuzz_support,
                              [enable fuzzer support]),
               , enable_fuzz_support=no)
 
+# Handle --enable-diff-fuzz-support
+AC_ARG_ENABLE(diff_fuzz_support,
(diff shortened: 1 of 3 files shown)


addisoncrump:
Should `config.h.in` be gitignored? It looks like we have two conflicting versions of autoconf.

PhilipHazel:
There are sure to be people who clone the repo and then expect config.h.in to be there for building, and if we gitignore it somebody will be caught out. No doubt either you or I will update our autotools at some point and this issue will go away. i am on Arch Linux, which I keep fairly up to date. I am currently on autoheader 2.72.

Luckily I had taken a break for lunch from making a new release, so I have managed to include this merge, but will now proceed.

PR #382, opened 2024-02-16
cmake: Make use of CMAKE_INSTALL_MANDIR

diizzyy:
Instead of hardcoding man path make use of CMAKE_INSTALL_MANDIR variable


carenas (approved):

PhilipHazel:
It is quite extraordinary how often patches arrive the day after a new release is out.

PR #389, opened 2024-02-23
fast_forward_first_n_chars should use while, not do-while

addisoncrump:
This addresses a potential issue where `.count` (i.e. length) is 0 but the first index is accessed nonetheless.


addisoncrump:
cc @zherczeg

addisoncrump:
@zherczeg, have you had time to review this?

zherczeg:
Could you add a test case? Not sure zero number of elements is valid, so there might be another bug here.

addisoncrump:
Hm; in the time that this has been around, my testcase no longer reproduces. I guess this was patched in the meantime somehow.

PR #398, opened 2024-03-18
Fixed edge case where entrysize == 0 for which the code is not well d…

jvanegue:
…efined

Consequence of missing check could be an infinite loop execution due to doing -= 0 on line 506 and += 0 n line 511 following incorrect arithmetic on line 501

The fix is a simple two-liner change to filter erroneous edge case, cutting off execution before getting into anything problematic.

API and data structure look exposed via external API call: pcre2_substring_copy_byname


PhilipHazel:
Your check is unnecessary because when entrysize == 0 (there are no named substrings in the pattern), both bot and top are also zero, and thus the loop is never entered.

PhilipHazel:
I'm closing this as unnecessary.

PR #416, opened 2024-06-08
update build.zig to support zig 0.13.0

player-two:
The cache director is now prefixed with a .[0]

Also, std.Build.LazyPath has changed, and now the recommended way to create one that is relative to the repo root is via std.Build.path().

[0]: https://ziglang.org/download/0.13.0/release-notes.html#codezig-cachecode-renamed-to-codezig-cachecode

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,5 +100,6 @@ src/stamp-h1
 
 zig-out/
 zig-cache/
+.zig-cache/
 
 # End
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -99,6 +99,6 @@ src/stamp-h1
 /bazel-*
 
 zig-out/
-zig-cache/
```
carenas, line 102: can we keep this so it also works for zig 0.12.1?

carenas (changes requested):
while 0.13.0 is the latest, it would be nice to keep this backward compatible

.gitignore
player-two: Yes of course :+1:

Issue #415, opened 2024-06-07
Test suite fails when targeting i686


dscho:
When I build and run the test suite in a MINGW32 build (i.e. targeting the i686 CPU architecture), I see that `RunTest` fails. Inspecting the log, I see a lot of mismatches in the "Memory allocation - compiled block" lines; They all seem to be off by 24, but the tests pass with the hard-coded numbers when targeting x86_64 or aarch64.

On i686, the tests pass after running this:

```sh
    for f in testdata/testoutput8-8-2 testdata/testoutput8-16-2 testdata/testoutput8-32-2
    do
      perl -pi -e 's{^(Memory allocation - compiled block : )(\d+)$}{$1 . ($2 - 24)}e' "$f"
    done
```

I _assume_ that the hard-coded values are correct for 64-bit CPU architectures only, and that the values may not want to be hard-coded like this?

PhilipHazel:
Back in the day I used to test on 32-bit architecture, but my old PC died. I am surprised it has taken so long for anyone to notice this. However, it is not easy to do anything about it. That particular test already has different outputs for each code unit size and each link size. There would have to be a way for RunTest to discover the 32/64 environment and use another set of test files for 32-bit. Or just skip this test - as it says in its comment: "This is just a doublecheck test to ensure the sizes don't go horribly wrong when something is changed."

thesamesam:
FWIW, you should be able to test on any 64-bit x86 CPU with a chroot and `linux32` if you want to.

Anyway, I hit this too, but I only hit it with the latest release.

PhilipHazel:
Do you mean 10.44? If so, that's odd because there aren't many changes from 10.43.

thesamesam:
Yes! Let me try bisect.

PhilipHazel:
Thank you.

carenas:
> Do you mean 10.44? If so, that's odd because there aren't many changes from 10.43.

the problematic code was added in 05aafb2e308e5529e3fd5fed45bae253eebed75b.

the 24 byte difference comes from the fact that the pcre2_cpde struct size includes also a memctl struct with 3 pointers, a couple of additional pointers and a size_t, all of which are bitwidth dependent.

dscho:
Thank you all!

PR #418, opened 2024-06-09
pcre2test: memory reports only compiled memory usage for code/data

carenas:
Since 05aafb2 (Implement pcre2_set_max_pattern_compiled_length() and set this limit in the fuzzer, 2024-04-24), the memory modifier has reported the full size of the allocated "code" returned by `pcre2_compile`.

Problem is that the size of the structure used to hold that in memory also depends on the platform ABI and even alignment by the compiler, and has been therefore fragile to compare.

Revert to reporting only the additional memory that `pcre2_compile()` will use for the compiled pattern (including any data tables) and make sure that the limit provided with `pcre2_set_max_pattern_compiled_length()` also avoid the internal struct overhead.

Fixes: #415


PhilipHazel:
Thanks, Carlo, for sorting this out. I clearly didn't think things through enough when I added that new limit.

PR #421, opened 2024-06-11
m4/pcre2_visibility.m4: add basic support for non GCC compilers

carenas:
Tested with the latest version of SunPro and TinyCC. It would also allow for other clang proprietary derivatives.

As usual, would fall cleanly for the cases where there is no support for `-fvisibility`


PhilipHazel:
This patch has broken the CIFuzz automatic check somehow.

Issue #423, opened 2024-06-11
Coverity defect: Illegal address computation


CrypticEcho:
PCRE2 version: 10.43
Our Coverity scan has reported a defect in PCRE2
**Illegal address computation**

If this address is later used for bounds checking another pointer before dereferencing, an overrun may occur due to the weak guard.

In pcre2_compile_8: An illegal address is computed, which either precedes a buffer or is more than just-past its end ([CWE-119](http://cwe.mitre.org/data/definitions/119.html))

Specific details:
pcre2_com[pile.c
```

10967 /* Errors discovered in parse_regex() set the offset value in the compile
10968 block. Errors discovered before it is called must compute it from the ptr
10969 value. After parse_regex() is called, the offset in the compile block is set to
10970 the end of the pattern, but certain errors in compile_regex() may reset it if
10971 an offset is available in the parsed pattern. */
10972
10973 HAD_CB_ERROR:
```
_**CID 1066215: (#5 of 5): Illegal address computation (OVERRUN)
44. illegal_address: pattern + cb.erroroffset evaluates to an address that is at byte offset 4294967295 of an array of 1 bytes.**_
```
10974ptr = pattern + cb.erroroffset;
10975
10976HAD_EARLY_ERROR:
10977*erroroffset = ptr - pattern;
10978
10979HAD_ERROR:
10980*errorptr = errorcode;
10981pcre2_code_free(re);
10982re = NULL;
        45. Jumping to label EXIT.
10983goto EXIT;
10984}
```

carenas:
> illegal_address: pattern + cb.erroroffset evaluates to an address that is at byte offset 4294967295 of an array of 1 bytes.

this seems suspiciously more like a bug in Coverity as `cb.erroroffset == UINT_MAX` and pattern is usually longer than 1 byte, specially if `parse_regex()` returned an error with that offset.

maybe it would be a good idea to trace all previous 43 steps to figure out where the logic went wrong.

CrypticEcho:
It is not uncommon for Coverity to report false positives.

Here is some additional context:
`PCRE2_CODE_UNIT_WIDTH=8`

pcre2_compile.c:
...
```
10103
10104 /*************************************************
10105 *     External function to compile a pattern     *
10106 *************************************************/
10107
10108 /* This function reads a regular expression in the form of a string and returns
10109 a pointer to a block of store holding a compiled version of the expression.
10110
10111 Arguments:
10112  pattern       the regular expression
[...]

PR #427, opened 2024-06-12
pcre2_compile: belt and suspenders for check_lookbehinds()

carenas:
Might address #423, but otherwise a good cleanup that shouldn't have any real impact whatsoever.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -293,7 +293,7 @@ versions. */
 #define META_LAST_QUANTIFIER  META_MINMAX_QUERY
 
 /* This is a special "meta code" that is used only to distinguish (*asr: from
-(*sr: in the table of aphabetic assertions. It is never stored in the parsed
+(*sr: in the table of alphabetic assertions. It is never stored in the parsed
 pattern because (*asr: is turned into (*sr:(*atomic: at that stage. There is
 therefore no need for it to have a length entry, so use a high value. */
 
@@ -9469,10 +9469,10 @@ for (;; pptr++)
   parsed_recurse_check *r;
   uint32_t *gptr, *gptrend;
   uint32_t escape;
+  uint32_t min, max;
   uint32_t group = 0;
   uint32_t itemlength = 0;
   uint32_t itemminlength = 0;
-  uint32_t min, max;
 
   if (*pptr < META_END)
     {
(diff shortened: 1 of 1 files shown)


src/pcre2_compile.c
```diff
@@ -10549,7 +10552,15 @@ if (has_lookbehind)
     }
   memset(cb.groupinfo, 0, (2 * cb.bracount + 1) * sizeof(uint32_t));
   errorcode = check_lookbehinds(cb.parsed_pattern, NULL, NULL, &cb, &loopcount);
-  if (errorcode != 0) goto HAD_CB_ERROR;
+  if (errorcode != 0)
+    {
+#ifdef PCRE2_DEBUG
+    /* BUG: check_lookbehinds() is missing code for a valid META */
+    if (errorcode == ERR70) cb.erroroffset = PCRE2_UNSET;
```
carenas, line 10559: it might be better to directly abort() here to help find this bug

src/pcre2_compile.c
```diff
@@ -9817,7 +9817,7 @@ for (;; pptr++)
   least one variable-length branch in the group, there is a test for the
   (smaller) variable-length branch length limit. */
 
-  if (INT_MAX - branchlength < (int)itemlength ||
```
PhilipHazel, line 9820: I don't think I understand this change. Note the comment that precedes it.

src/pcre2_compile.c
```diff
@@ -9877,7 +9877,7 @@ PCRE2_SIZE offset;
 uint32_t *bptr = *pptrptr;
 uint32_t *gbptr = bptr;
 int maxlength = 0;
-int minlength = INT_MAX;
+int minlength = LOOKBEHIND_MAX;
```
PhilipHazel, line 9880: This should probably also remain as INT_MAX.

PhilipHazel:
Thanks for putting the time into studying the code! I'm going to merge this now.

src/pcre2_compile.c
CrypticEcho: Can this commit be applied on top of rel 10.43 as patch? (or should I update the whole thing).

carenas: the critical part of the "fix" applies cleanly all the way to 10.30:

```
diff -ur pcre2-10.30/src/pcre2_compile.c pcre2-10.30.p/src/pcre2_compile.c
--- pcre2-10.30/src/pcre2_compile.c	2017-08-12 09:18:55
+++ pcre2-10.30.p/src/pcre2_compile.c	2024-06-14 15:49:14
@@ -8858,6 +8858,11 @@
   switch (META_CODE(*pptr))
     {
     default:
+
+    /* The following erroroffset is a bogus but safe value.
+    This branch should be avoided by providing a proper
+    implementation for all supported cases below. */
+    cb->erroroffset = 0;
     return ERR70;  /* Unrecognized meta code */
 
     case META_ESCAPE:
```

the full patch applies cleanly to 10.43 or any later versions.

PR #444, opened 2024-08-24
Fix bug in 'first code unit' and 'last code unit' optimization combined with lookahead assertion

alexdowad:
Looking forward to review on my first bug fix for PCRE2!

Thanks very much to @PhilipHazel for suggesting the simple strategy used here for avoiding the `firstcu` optimization in cases where it is invalid. Maybe later we can improve the analysis which is done to identify the first/last code units which are required for a match, but for now, this fixes a bug and still preserves the optimization in many cases.

I haven't followed the PCRE2 brace style, but am happy to adjust my code style if needed.


PhilipHazel:
I see that test 2 has failed (maybe others as well). Probably because there are changes to some of the tests. Please check the remaining tests - it may be you just have to update the output. I do also like to note changes in the ChangeLog file, partly because that's always been done (long before GitHub) but also because it's easier to search as a historical record compared with Git logs.

alexdowad:
@PhilipHazel Sorry about that, I will be more careful next time.

I do notice that for the one test case in test 2 which 'failed'... this PR actually *fixes* that case. As it is, the test expects to see "Subject length lower bound = 2", which is wrong.

alexdowad:
Hmm, now I need to figure out how to build the 16-bit version of PCRE2 so I can fix the 16-bit tests...

alexdowad:
> Hmm, now I need to figure out how to build the 16-bit version of PCRE2 so I can fix the 16-bit tests...

That is done, now just waiting for the test suite to run again...

PhilipHazel:
Oh how annoying. Usually once I've approved workflows to run for a given contributor, they do so automatically subsequently. Unfortunate timing (I was asleep). They are running now.

PR #445, opened 2024-08-26
Support a new experimental feature called scan substring

zherczeg:
This patch contains the core code of a new experimental feature called scan substring. It allows anchored matching the content of a capture group to a subpattern.

The parser (compiler) and interpreter is updated to support it. Dfa interpreter will not support it. JIT will support it in the future. Documenation is missing, I hope @PhilipHazel can help me with that. Several tests were added.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -75,6 +75,10 @@ with JIT was correct.
 12. Add a new error code (PCRE2_ERROR_JIT_UNSUPPORTED) which is yielded
 for unsupported jit features.
 
+13. Add a new experimental feature called scan substring. This feature
+is a new type of assertion which matches the content of a capturing block
+to a sub pattern.
+
 
 Version 10.44 07-June-2024
 --------------------------
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -231,63 +231,65 @@ code (meta_extra_lengths, just below) must be updated to remain in step. */
 #define META_COND_RNAME       0x80130000u  /* (?(R&name)... */
 #define META_COND_RNUMBER     0x80140000u  /* (?(Rdigits)... */
 #define META_COND_VERSION     0x80150000u  /* (?(VERSION<op>x.y)... */
-#define META_DOLLAR           0x80160000u  /* $ metacharacter */
-#define META_DOT              0x80170000u  /* . metacharacter */
-#define META_ESCAPE           0x80180000u  /* \d and friends */
-#define META_KET              0x80190000u  /* closing parenthesis */
-#define META_NOCAPTURE        0x801a0000u  /* no capture parens */
-#define META_OPTIONS          0x801b0000u  /* (?i) and friends */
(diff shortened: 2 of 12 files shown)


zherczeg:
Some comments besides the `\z`.

What is better: `scan_substring` or `scan_string`? I am not sure we need to emphasize the `sub` part, it does not add much value in this particular case.

The OP_CREF comment (`Cond Ref -> Capture Ref`) will be done in a different patch. I suspect it will affect a lot of lines, and I don't want to complicate this work with that.

PhilipHazel:
I prefer scan_substring because that is more precise.

zherczeg:
We could talk about the syntax as well. We could use:
`(*scs:1:...)` - this requires one less character, but looks unusal in regex world
`(*scs(1)...)` - this requires one less character, but could looks confusing

I prefer the current one, but it is worth to discuss the alternatives.

PhilipHazel:
I prefer parentheses.

PhilipHazel:
As you can see, I've done the merge. All looks good and I will work on the documentation - maybe try some tests of my own. There is one thing: around line 685 of pcre2_compile.c you have this:

```
   {  3, META_SCS_NUMBER        }, /* placeholder, updated later */
  { 14, META_SCS_NUMBER        }, /* placeholder, updated later */
```
 
It looks as if that is a typo and that the second line should be NAME, not NUMBER. I see from the later code that it doesn't matter, but I think this should be changed because it is a trap. It would need one extra "case" later on, but I think that's all. I'm happy to make the change if you agree.

PhilipHazel:
Two things: (1) Something needs to be done about \z vs \Z, as you pointed out. In the documentation for invalid UTF, where the subject is broken into substrings, it does state that these do not match the end of a substring. So I think this is a bug in \Z handling in this case and I will test for this and fix it. However, I do feel that these *should* match in the scs case, but that would require a "depth of scs" count in order to handle nested scs, but this ought to be straightforward. I will look at that too.

(2) While inventing some extra tests, I have found that an scs assertion doesn't work as a condition. I think it should, and I will work on that as well (should be easy, no need for both of us to be working on the same thing).

Then I think I have enough information to work on the docs!

zherczeg:
(1) Ok for me. I don't know why there was a difference, I just mentioned it.

(2) What does "condition" means here? Conditional block? It is not an atomic assertion, so it cannot work there.

PhilipHazel:
Oh, yes, of course, condition assertions must be atomic, but I got an odd
error when I tried it. I will check again and make sure the error is useful.

PhilipHazel:
I have committed my patches, which make \z, \Z, and $ match at the end of the captured string. However, this is anomalous because \A and ^ do *not* match at the start. For uniformity, \z etc should not match (as I think you originally suggested) unless it really is the end of the original subject but I do think it could be useful to match the end of the capture. The start doesn't matter because the match only takes place at the start - it's effectively anchored - so I propose leaving it as it is and documenting this carefully.

I also fixed a missing label in pcre2_study(). I will now start on the documentation.

PR #447, opened 2024-08-28
Support scan_substring in jit

zherczeg:
The first patch for jit support.

Notes:
- I have noticed that repeats are not supported for scan_substring assertions. This should work like other repeated assertions, so '*' is replaced by `?`. I hope I fixed it in the compiler, and added tests for it. Idea: maybe a byte-code dump should be added for them, to check the generated code.

- The subject start is not affected by scan_substring, so it is possible to use a lookbehind assertion to check the characters before the start. We should mention this is the documentation. It also revealed that I forgot to add the opcode to `check_lookbehinds()`. I am sure there are similar cases elsewhere :(

- I am sure I need to fix control verb related issues. I will do that in other patches.


zherczeg:
I have added tests for repetition

PR #449, opened 2024-08-29
Fix backtracking bugs in scan_substring

zherczeg:
I tried to figure out how to fix the bugs in the interpreter. Unfortunately the F,N,P, Flast_group_offset variables are a bit confusing for me, but at least the tests pass. I am thinking about whether keeping the current F instead of another RMATCH is a better idea. I am not even sure it is possible to do that.

### DIFF
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -2524,6 +2524,13 @@ while (cc < ccend)
     cc += 1 + LINK_SIZE;
     break;
 
+    case OP_ASSERT_SCS:
+    SLJIT_ASSERT(PRIVATE_DATA(cc) != 0);
+    if (recurse_check_bit(common, PRIVATE_DATA(cc)))
+      length += 2;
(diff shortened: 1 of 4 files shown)


PhilipHazel:
When I implemented the heap-based version of pcre2_match(), replacing the recursive function call version, I understood how the F, N, etc pointers worked, but now I have forgotten the detail. The last time I fixed a bug in this area (10.43/47 in ChangeLog) it was very difficult!

zherczeg:
I suspect F represents the last stack frame. However, to support OP_KET, the "chain list" of brackets is needed as well, represented by Flast_group_offset, which simply points to the latest bracketed group of any stack frame. P represents the frame of the last bracketed group, and N is the frame after that, although I don't know why it is important. I am not sure this is correct.

PhilipHazel:
Yes, F is the active frame, N is next and P is previous. At least, that's how I think I used those names, but I also suspect N and P are occasionally used for other purposes.

PR #457, opened 2024-09-03
Rename cond ref to capture ref

zherczeg:
I promised this renaming.

### DIFF
--- a/src/pcre2_internal.h
+++ b/src/pcre2_internal.h
@@ -1711,7 +1711,7 @@ some cases doesn't actually use these names at all). */
   "Cond",                                                         \
   "SBra", "SBraPos", "SCBra", "SCBraPos",                         \
   "SCond",                                                        \
-  "Cond ref", "Cond dnref", "Cond rec", "Cond dnrec",             \
+  "Capture ref", "Capture dnref", "Cond rec", "Cond dnrec",       \
   "Cond false", "Cond true",                                      \
   "Brazero", "Braminzero", "Braposzero",                          \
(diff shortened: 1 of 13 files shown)


zherczeg:
Btw there is no dnref in any of the test outputs.

carenas (changes requested):
need to also update `pcre2_printint`

zherczeg:
Where? It uses the OP_NAME_LIST from the internal header. The arguments are not changed.

PhilipHazel:
I think he means the phrase "cond ref" in the output should be "capture ref".

zherczeg:
Which output? I am sorry I cannot follow.

carenas:
Sorry; I updated my original comment so it is less vague.

I also think it might be worth adding some tests that exercise it:

```
diff --git a/testdata/testinput2 b/testdata/testinput2
index 9c59bb5..1324b4b 100644
--- a/testdata/testinput2
+++ b/testdata/testinput2
@@ -2862,6 +2862,13 @@
     b"11111
     a"11111
 
+/(?:a(?<digit>[0-5])|b(?<digit>[4-7]))c(?(<digit>)d|e)/B,dupnames
+    a4cd
+    b4cd
+\= Expect no match
+    a6cd
+    a6ce
+
 /^(?|(a)(b)(c)(?<D>d)|(?<D>e)) (?('D')X|Y)/IBx,dupnames
     abcdX
     eX
```

zherczeg:
Thank you for the explanation. I have updated the patch.

carenas (approved):

PR #458, opened 2024-09-03
Add missing 'expand' modifier to list in pcre2test manpage

alexdowad:
The 'expand' modifier is mentioned in later sections of the manpage, but is not included in any of the lists of possible modifiers.

### DIFF
--- a/doc/pcre2test.1
+++ b/doc/pcre2test.1
@@ -654,6 +654,7 @@ heavily used in the test files.
       convert_glob_separator=c  set glob separator character
       convert_length            set convert buffer length
       debug                     same as info,fullbincode
+      expand                    expand repetition syntax in pattern
       framesize                 show matching frame size
       fullbincode               show binary code with lengths
   /I  info                      show info about compiled pattern
--- a/doc/html/pcre2test.html
+++ b/doc/html/pcre2test.html
@@ -698,6 +698,7 @@ heavily used in the test files.
       convert_glob_separator=c  set glob separator character
       convert_length            set convert buffer length
       debug                     same as info,fullbincode
+      expand                    expand repetition syntax in pattern
       framesize                 show matching frame size
       fullbincode               show binary code with lengths
   /I  info                      show info about compiled pattern
--- a/doc/pcre2test.txt
+++ b/doc/pcre2test.txt
@@ -632,6 +632,7 @@ PATTERN MODIFIERS
              convert_glob_separator=c  set glob separator character
              convert_length            set convert buffer length
              debug                     same as info,fullbincode
+             expand                    expand repetition syntax in pattern
              framesize                 show matching frame size
              fullbincode               show binary code with lengths
          /I  info                      show info about compiled pattern


doc/pcre2test.1
```diff
@@ -654,6 +654,7 @@ heavily used in the test files.
       convert_glob_separator=c  set glob separator character
       convert_length            set convert buffer length
       debug                     same as info,fullbincode
+      expand                    enable repetition feature in subject
```
carenas, line 657: "subject" is the data that gets matched against, while `expand` affects the "pattern" or "needle".

maybe a description in imperative form like the ones used for `convert*` that says: `activate preprocessing of the pattern for expansion`?

carenas (changes requested):
Terminology used is confusing, note that while the syntax used is similar, the `expand` modifier is only needed to instruct `pcre2test` to process and create an "expanded" pattern before calling `pcre2_compile()`.

```
  re> /\[a]{3}/B,expand
        Bra
        aaa
        Ket
        End
```

The expansion of the data doesn't need that modifier:

```
  re> /aaa/B
        Bra
        aaa
        Ket
        End
data> \[a]{3}
 0: aaa

```

doc/pcre2test.1
alexdowad: Dear @carenas, thank you for the great comment! One thing seems confusing: you kindly explained that `expand`  affects only the pattern and not the subject, but the output you posted from `pcre2test` seems to show that expansion is done on *both* the pattern and also the subject. Is that incorrect? If so, why does `/aaa/` match `aaa` in `\[a]{3}`?

Once that is clarified, I'm sure we can find a simple but accurate way to express what `expand` does. Of course, readers will still need to look at the detailed description if they want to understand fully.

carenas: > show that expansion is done on _both_ the pattern and also the subject.

correct, and the syntax is similar, but note that there is no need for a modifier for the subject to be expanded, unlike what happens with the pattern when the modifier is missing:

```
  re> /\[a]{3}/B
        Bra
        [a
        ]{3}
        Ket
        End
```

PhilipHazel:
The bottom line is that expansion is always available in subject lines because there is no conflict with the syntax  - and in fact it was implemented long before expansion in the pattern. For the pattern, there has to be a switch to turn it on because the syntax is the same as pattern syntax, as explained in the doc. So the listing should say something like "enable string expansion in pattern".

alexdowad:
@carenas @PhilipHazel Thanks, this is clear now.

alexdowad:
Updated. Please kindly see if this looks better now.

carenas (approved):
since the autogenerated `pcre2test.txt` was updated why not also do the same  for `doc/html/pcre2test.html'?

see `PrepareRelease`

alexdowad:
Dear @carenas, thanks for pointing me to `PrepareRelease`. I have updated `pcre2test.html` as advised.

carenas (approved):

PR #459, opened 2024-09-03
Support multiple group references for scan substring

zherczeg:
The idea is extending the scan substring argument to a list of capture group references. The brackets are scanned in definition order, and the first bracket in the list, which value is set is used for the substring scanning.

Unfortunately the implementation required code reorganizations, which made the patch more complex than it should be.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -233,63 +233,65 @@ code (meta_extra_lengths, just below) must be updated to remain in step. */
 #define META_COND_VERSION     0x80150000u  /* (?(VERSION<op>x.y)... */
 #define META_SCS_NAME         0x80160000u  /* (*scan_substring:(<name>)... */
 #define META_SCS_NUMBER       0x80170000u  /* (*scan_substring:(digits)... */
-#define META_DOLLAR           0x80180000u  /* $ metacharacter */
-#define META_DOT              0x80190000u  /* . metacharacter */
-#define META_ESCAPE           0x801a0000u  /* \d and friends */
-#define META_KET              0x801b0000u  /* closing parenthesis */
-#define META_NOCAPTURE        0x801c0000u  /* no capture parens */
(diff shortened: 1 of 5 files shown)


zherczeg:
The patch is ready for review, although I have a question. In the past the parser (compiler) run twice, first to detect the size, and second to generate code. However, it seems that it just overallocate a target buffer now. In the latter case, I don't have to check the remaining parsed_pattern buffer size in the loop. Am I right?

zherczeg:
Note: in the future, we could remove duplications from the capture group list as an optimization, but I don't think this is worth it.

PhilipHazel:
The main compiler part of pcre2_compile.c still runs twice, as before. It is just the parsing into a sequence of META items that only runs once.

zherczeg:
Is this true: The META generation phase has a large buffer, and the byte code generation phase already knows the precise size?

@PhilipHazel may I ask you to update the documentation? The new feature is that multiple capture references can be listed for scan substring, and the first one in declaration order which value is set is used for the range of scan substring. As for duplicated names, it follows the declaration order in the pattern.

Actually it was dupnames which raised the checking order problem. As for conditional blocks, the order doesn't matter, since the set/unset status is checked only. As for scan substring, the checking order is pre defined. First I thought that users should be able to set the checking order. Then I realized it is better if we would allow multiple capture groups. This way users can specify the checking order.

PhilipHazel:
The META generation phase estimates the size of the buffer by requiring each item to use no more meta words than the number of characters in the item. See the comment around line 10671 in pcre2_compile.c and the following code. The point is, the memory used for the META data is released at the end of compilation, so if it is oversized it doesn't matter. The byte code generation phase does not know the size needed, which is why it is run twice, and this memory is returned to the user so should not be over-estimated. 

I have already updated the documentation.

zherczeg:
Thank you!

PR #460, opened 2024-09-03
Use assertion to ensure erroroffset return from pcre2_compile is valid

alexdowad:
When testing the new pattern rewriting phase for regex compilation using a fuzzer, I had a scary experience. Due to a bug in my pattern rewriting code, pcre2_compile() could return a totally invalid erroroffset. If a library user tried to do something with the erroroffset without checking it for validity, in the worst case, this had the potential to lead to an RCE vulnerability.

In case something similar ever happens again, I've added an assertion which will make it easier to notice the problem.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -11147,6 +11147,8 @@ HAD_CB_ERROR:
 ptr = pattern + cb.erroroffset;
 
 HAD_EARLY_ERROR:
+PCRE2_ASSERT(ptr >= pattern); /* Ensure we don't return invalid erroroffset */
+PCRE2_ASSERT(ptr <= (pattern + patlen));
 *erroroffset = ptr - pattern;
 
 HAD_ERROR:


alexdowad:
I'm surprised to see that one of these assertions fails when `RunTest` executes in CI. I haven't been able to duplicate this locally yet (all the tests pass for me) and am still trying to figure out the reason.

src/pcre2_compile.c
```diff
@@ -11147,6 +11147,8 @@ an offset is available in the parsed pattern. */
 ptr = pattern + cb.erroroffset;
 
 HAD_EARLY_ERROR:
+PCRE2_ASSERT(ptr >= pattern); /* Ensure we don't return invalid erroroffset */
+PCRE2_ASSERT(ptr < (pattern + patlen));
```
carenas, line 11151: `s/</<=/`, since the NUL at the end of pattern is a perfectly valid erroroffset

carenas (changes requested):
the output of the steps from CI could help you replicate locally, maybe didn't build with `--enable-debug`?

src/pcre2_compile.c
alexdowad: @carenas Thanks for mentioning that. I actually flip-flopped between `<` and `<=` when preparing this PR. However, I noticed that (**if** I have understood well) it seems that patterns which are passed to `pcre2_compile` are not always NUL-terminated. (When the pattern is NUL-terminated, one can pass `PCRE2_ZERO_TERMINATED` as the `length` argument to `pcre2_compile`.)

Is that correct? If so, should the assertion be predicated on `length == PCRE2_ZERO_TERMINATED`?

carenas: > should the assertion be predicated on `length == PCRE2_ZERO_TERMINATED`?

no; and it is unlikely to work as the length value is reset earlier AFAIK.

the [documentation](https://pcre2project.github.io/pcre2/doc/html/pcre2api.html) mentions:

> Some errors are not detected until the whole pattern has been scanned; in these cases, the offset passed back is the length of the pattern

alexdowad: > > should the assertion be predicated on `length == PCRE2_ZERO_TERMINATED`?
> 
> no; and it is unlikely to work as the length value is reset earlier AFAIK.

I think I didn't express myself well. I was referring to the value of `length` *on function entry*, not at the point where the assertion occurs.

> the [documentation](https://pcre2project.github.io/pcre2/doc/html/pcre2api.html) mentions:
> 
> > Some errors are not detected until the whole pattern has been scanned; in these cases, the offset passed back is the length of the pattern

Hmm. If that is so, then I guess `pattern + erroroffset` is not guaranteed to be a valid pointer. Since PCRE2 users may presumably wish to use `pattern + erroroffset` to display the part of the pattern which caused an error, this feels like a defect in the API to me.

I guess if users read the documentation, then they should know to check for the special value `erroroffset == patlength`.

alexdowad: Anyways, regardless of the issues mentioned, it is apparent that the assertion should indeed use `<=` and not `<`. PR updated.

carenas: > I guess if users read the documentation, then they should know to check for the special value `erroroffset == patlength`.

sure hope they do, specially if trying to print the place where the pattern error was found, as they then would realize it might not be printable, if it happens to be in the middle of an UTF8 character, for example.

carenas: > > > should the assertion be predicated on `length == PCRE2_ZERO_TERMINATED`?
> > 
> > 
> > no; and it is unlikely to work as the length value is reset earlier AFAIK.
> 
> I think I didn't express myself well. I was referring to the value of `length` _on function entry_, not at the point where the assertion occurs.

What I meant was that there is no way for PCRE2 to know if a pattern is NUL terminated or not (PCRE2_ZERO_TERMINATED is only a hint and an "optimization" so callers of this function can rely on PCRE2 doing `strlen` for them).

For additional context, '\0' in patterns used to be impossible before PCRE2.

carenas:
Note that this assertions is buggy, since `patlen` was never updated when the pattern was rewritten

alexdowad:
@carenas Good point. That would be true *if* pattern rewriting is done, which is not the case in `master` yet.

carenas:
> That would be true

if being pedantic, you are correct that this assert isn't buggy yet in master, but obviously the comment was meant to address it within the context of this change being part of that series (as explained above) and which was also linked.

I did it this way to avoid polluting the conversation in the other PR, not to make a false statement, but would follow whichever preferences you have to avoid further confusion.

PS: FWIW I prefer not to be "`@`" to avoid notification spam, but if you haven't seen it yet, the discussion where this PR was linked also addresses this point and the fact that it might require API changes.

alexdowad:
Understood, I won't keep sending you "@" notifications then.

This PR is not simply part of a series with the other PR; it is a worthwhile change on its own, which is why I submitted it separately.

alexdowad:
Please see added comment in #464 on why the assertion added in this PR is still valid, even if pattern rewriting optimizations are later merged into `master`.

carenas:
> hy the assertion added in this PR is still valid

Glad to be wrong about needing to update `patlen`, since that was also a major sticking point as there is no way to communicate that back to the user by the current API.

This means though that ALL `erroroffset`  (except for the few ones that were cached in `parsed_pattern`) are likely incorrect, eventhough you are right that they are also likely to be lower than `patlen` and therefore not trigger this assert.

alexdowad:
> ALL `erroroffset` (except for the few ones that were cached in `parsed_pattern`) are likely incorrect

I don't believe that is the case, for the reasons explained in #464. However, it is good to be sure, so I will check this area over again.

carenas:
> don't believe that is the case

`cb.erroroffset` is set by all functions that participate in the compilation with code like:

```
if (bad)
  {
  *errorcodeptr = ERR
  cb->erroroffset = offset; /* or an expression and sometimes even a ptrdiff */
  return FALSE;
  }
 ```

how can that work when offset is based on the rewritten pattern?

alexdowad:
> how can that work when offset is based on the rewritten pattern?

The point is that if it is set to a value other than `patlen`, `erroroffset` *must* be set to an offset within the pattern string, by code which is analyzing the pattern string. The pattern rewriter does not modify the pattern string. It only modifies the `parsed_pattern` vector. Hence, it has no effect on any code which analyzes the pattern string.

If there is any code in PCRE2 which currently sets `erroroffset` based on the offset of a `parsed_pattern` item within the `parsed_pattern` vector, that is very wrong and must be fixed. I don't believe that such code exists, but I would like to check and make sure.

We definitely do have code which sets `erroroffset` to 0, to `patlen`, and to `PCRE2_UNSET`. Setting it to `PCRE2_UNSET` seems wrong to me and I would like to look into that. However, it is a different issue.

carenas:
I stand corrected, apologies for the confusion.

PR #462, opened 2024-09-06
build: improve portability of assertions

carenas:
Fix a build failure in AIX and prevent warnings in Windows.


zherczeg (commented):
Nice patch!

PR #463, opened 2024-09-06
Guard against out-of-bounds memory access when parsing LIMIT_HEAP et al

alexdowad:
While fuzzing PCRE2, I found a case in which `pcre2_compile` may incorrectly read past the end of the pattern string which it is given. Here are the details:

Patterns passed to `pcre2_compile` are not guaranteed to be null-terminated. Also, it can happen that there is an invalid pattern like this:

```
(*LIMIT_HEAP=123
```
If the next byte of memory after the end of the pattern happens to be a digit, it will be parsed as part of the limit value. Or, if the next byte is a right parenthesis character, it will be taken as the end of the `(*LIMIT_HEAP=nnn)` construct.

This will result in `skipatstart` being larger than `patlen`, which will result in underflow and an erroneous call to malloc requesting a huge number of bytes.

### DIFF
--- a/testdata/testoutput15
+++ b/testdata/testoutput15
@@ -111,10 +111,10 @@ Minimum depth limit = 10
  3: ee
 
 /(*LIMIT_MATCH=12bc)abc/
-Failed: error 160 at offset 17: (*VERB) not recognized or malformed
+Failed: error 160 at offset 16: (*VERB) not recognized or malformed
 
 /(*LIMIT_MATCH=4294967290)abc/
-Failed: error 160 at offset 24: (*VERB) not recognized or malformed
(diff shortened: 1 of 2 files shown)


testdata/testoutput15
```diff
@@ -111,10 +111,10 @@ Minimum depth limit = 10
  3: ee
 
 /(*LIMIT_MATCH=12bc)abc/
-Failed: error 160 at offset 17: (*VERB) not recognized or malformed
+Failed: error 160 at offset 16: (*VERB) not recognized or malformed
```
alexdowad, line 114: This is more logical now: the first bad character in the LIMIT_MATCH construct is at offset 16, not offset 17.

zherczeg (approved):
LGTM

Issue #411, opened 2024-06-04
Alternative branch should match shared prefix only once


mvorisek:
This is a feature request for engine optimization.

Repro: https://3v4l.org/JKn6D/rfc#vgit.master_jit

Such dummy regexes like `(a|b|...|z)` are very common for language grammars or routing.

The repro shows massive speedup (from 3.4 ms to 0.82 ms). According regex101.com the speedup is 865 vs. 93 steps only.

The engine should detect shared prefix/subpattern in alternative branch like `(ab?c|ab?d)` and optimize it into `(ab?(?:c|d))`.

Challenges:

- non-consequent shared prefix - ex. `(ab?c|x|ab?d)`
- subpatterns with capturing groups - ex. `((a)b|((a)c)`

But I belive they can be easily handled internally.

zherczeg:
Optimizations are not that easy in general. Pattern to pattern optimizations are usually done by other projects such as https://github.com/zherczeg/repan . The general workflow is optimizing patterns at compile time, and the application uses the optimized variants. This saves time and energy, since the costly steps are done once.

mvorisek:
> Pattern to pattern optimizations are usually done by other projects such as https://github.com/zherczeg/repan .

The repan project impl. the feature as "merge alternatives" - https://github.com/zherczeg/repan/blob/25d6e3aabf/tests/opt/test_merge_alternatives_expected.txt.

Such optimization is very powerful with a little extra CPU cycles needed when compiling the pattern and with zero to minimal extra CPU cycles needed when evaluating the compiled pattern. Such optimization should be done by PCRE library itself as many typical usages do not involve compile phase (interpreted languages, user search inputs, ...) and requiring this to be done by intermediate library would require parsing the regex twice and requiring all vendors to maintain extra dependency.

I have updated the description to stress there are cases which can and should be optimized internally but cannot be done by "pattern to pattern optimizations" anyway.

PhilipHazel:
Now that it's recorded as an issue, somebody might pick this up in the future. It won't be me, because after I manage to get 10.44 released (soon, I hope) I want to start looking for someone to take over the PCRE2 project because I am getting too old. :-(

PhilipHazel:
[...]

Issue #13, opened 2021-08-24
Revise and extend character classes


PhilipHazel:
This issue records several potential upgrades to the handling of character classes in PCRE2. This could be a lot of work in both the interpreters and the JIT.

1. The current code in the compiler has been hacked into an untidy mess and the compiled code is also messy. A revised implementation is needed that is more uniform and can better handle Unicode characters so as to make matching more efficient. For example, bitmaps could be used for runs of characters other than just 0-0xFF. Or some better coding scheme could be devised.

2.  Perl has an experimental extended class feature as in this example:
 
  /(?[ ( \p{Thai} + \p{Lao} ) & \p{Digit} ])/

Any new compiled format should be able to handle such things.

3. There was a request for a way of re-defining \w (and therefore \W, \b, and \B). An in-pattern sequence such as (?w=[...]) was suggested. Easiest way would be simply to inline the class, with lookarounds for \b and \B. Ideally the setting should last till the end of the group, which means remembering all previous settings; maybe a fixed amount of stack would do - how deep would anyone want to nest these things? Of course, this idea also suggests redefining \d and \s. Is this worth doing, given that named groups can be used? It would be more efficient, because it can be processed at compile time.

michaeltang4829:
Hi, I'm curious if there's any new interesting in this effort? As mention above, Perl already has experimental behavior with extended character classes. I believe ECMA JavaScript and Rust engines already support character class set operations e.g. subtraction, intersection, union, etc... And Python is issuing FutureWarnings hinting this feature is coming to their engine soon. It's without a doubt this could be a huge effort to implement.

PhilipHazel:
Quite a long time ago I had some ideas as to how to do this, but I didn't write them down and I got distracted with other things. It would, as you say, be a big effort, affecting the compiler, the interpreters, and the JIT coding. I won't be getting involved because I'm trying to get other folk to take over PCRE2 - see #426.

PhilipHazel:
I am closing this because  #553 implements extended character classes.

PR #464, opened 2024-09-06
Rewrite regexes where common prefix can be pulled out from alternation branches

alexdowad:
Thanks to @mvorisek for suggesting this optimization.

A new "rewrite" pass has been added to the regex compilation process. For now, the rewrite pass only optimized one type of regex: those where every branch of an alternation construct has a common prefix.

In such cases, we rewrite the regex like so (for example):

```
    (abc|abd|abe) ⇒ (ab(?:c|d|e))
```

An extra non-capturing group is not introduced if the alternation is within a non-capturing group (which is not quantified using ?, *, or a similar suffix). In that case we simply do something like:

```
    (?:abc|abd|abe) ⇒ ab(?:c|d|e)
```

In some edge cases, it is possible that rewriting a group with common alternation prefix might open up the opportunity to pull out more common prefixes. For example:

```
    (a(b|c)d|(ab|ac)e)
```

In that case, if the group '(ab|ac)' was rewritten to pull out the common prefix, it would then become possible to pull out a common prefix from the top-level group. However, we do not take advantage of that opportunity. Further, we do not perform the rewrite in cases where the prefixes are semantically equivalent, but parse to a different `parsed_pattern` sequence.

Groups which the regex engine might need to backtrack into are never pulled out, since this could change the order in which the regex engine considers possible ways of matching the pattern against the subject string, and could thus change the returned match. For example, this pattern will not be rewritten:

```
    ((?:a|b)c|(?:a|b)d)
```

Also, callouts are never extracted even if they form a common prefix to an alternation. Some backtracking control verbs, like (*SKIP) and (*COMMIT), are never extracted either. Further, alternation is not rewritten in lookbehind assertions, since this can easily convert fixed-length lookbehind to variable lookbehind.

A different type of rewrite is performed if an alternation construct matches only single, literal characters:

```
    (a|b|c) ⇒ ([a-c])
```

A new compile option, `PCRE2_NO_PATTERN_REWRITE`, has been added to skip the pattern rewrite phase when compiling a pattern.

I have extensively fuzzed this patch. Fuzzing revealed many bugs, which have been fixed.


alexdowad:
I've noticed one thing here... `RunTest.bat` only runs test suites 1-23 on Windows (24, 25, and 26) are not run.

Is this intentional? Should I add the new suite 27 to RunTest.bat?

src/pcre2_compile.c
```diff
@@ -10695,9 +11421,9 @@ if (!utf)
 is set we have to assume a numerical callout (4 elements) for each character
 plus one at the end. This is overkill, but memory is plentiful these days. For
 many smaller patterns the vector on the stack (which was set up above) can be
-used. */
+used. Include space for wrapping pattern in non-capturing parentheses. */
```
carenas, line 11424: A comment that explains better why the magic number 2 is needed, might be better IMHO

src/pcre2_compile.c
```diff
@@ -5178,7 +5187,722 @@ errorcode = ERR79;
...
+    if (group_end != cb->parsed_pattern_end-1) {
+      break;
+    }
+
+    size_t pattern_length = cb->parsed_pattern_end - cb->parsed_pattern;
+    /* Copy down everything except for unneeded BRA, unneeded KET, and META_END
+     * We will leave some junk in the high end of the parsed_pattern buffer, but
+     * this doesn't matter; `parsed_pattern_end` will point to the end of the used
+     * portion, and we will ensure the used portion ends with META_END */
+    memmove(&cb->parsed_pattern[0], &cb->parsed_pattern[1], (pattern_length - 3) * sizeof(uint32_t));
```
carenas, line 5901: wouldn't be easier to override the last record with META_END and move the pointer to the first forward?

src/pcre2_compile.c
```diff
@@ -5178,7 +5187,722 @@ errorcode = ERR79;
...
+{
+  rewrite_buf buf;
+  rewrite_buf_init(&buf);
+
+  rewrite_alternation(cb->parsed_pattern, cb->parsed_pattern_end-1, NULL, &buf, cb->parsed_pattern, cb->start_pattern, &ccontext->memctl);
+  rewrite_finish(&buf, cb, &ccontext->memctl, heap_parsed_pattern);
+}
+
+/* Strip non-capturing parentheses which may be wrapping the entire regex */
+static void strip_enclosing_nocaptures(compile_block *cb)
```
carenas, line 5888: this is a fairly unique "optimization", care to explain the rationale of why is it needed/desired in the comment instead of the "what" which can be inferred by the code?

carenas (commented):
Didn't fully look at the code (or even pulled it), as the difference on style/design and white space changes keeps distracting me though, but if this code hasn't been tested yet in Windows it might be worth marking it as "draft".
Notice there are some warnings triggered by the CI, and there is at least one place where `-Wdeclaration-after-statement` should trigger.

RunTest
```diff
@@ -898,6 +903,18 @@ for bmode in "$test8" "$test16" "$test32"; do
     fi
   fi
 
+  # Pattern rewriter tests
+
+  if [ $do27 = yes ] ; then
+    echo $title27
+    if [ $utf -eq 0 ] ; then
```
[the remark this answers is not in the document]
alexdowad: Excellent point, will do that.

testdata/testinput17
```diff
@@ -282,7 +282,7 @@
 
 /[axm]{7}/
 
-/(.|.)*?bx/
+/(a|.)*?bx/
```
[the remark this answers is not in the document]
alexdowad: Good point.

src/pcre2_compile.c
alexdowad: True, that is very true.

src/pcre2_compile.c
alexdowad: Will do. The point is basically that the new code adds an extra `(?:` and `)` wrapping every pattern at the top level, in order to make recursive traversal of the pattern easier. So that those do not affect anything in the following stages, it strips them off after pattern rewriting is complete.

src/pcre2_compile.c
alexdowad: Ah, wait a moment, I answered too soon. `cb->parsed_pattern` may, in some cases, be a malloc'd pointer. We have to keep it so that we can `free` it later. So if we want to be able to remove a prefix from the pattern by bumping `cb->parsed_pattern` forward, we would need to add another member to `compile_block` to record the original value.

Do you think that would be worthwhile? Personally, I have my doubts.

Even so, at the very least, I could ensure that `memmove` is only done once even if the pattern is wrapped in multiple layers of redundant parens.

src/pcre2_compile.c
carenas: got it, so the comment should be using "are" instead of "may be" and that also explains the "magic 2" and other oddities like that renamed pointer name and the addition of a clone of it in "intops".

anyway, thanks for doing this and kudos on what looks like a sound first major contribution.

src/pcre2_compile.c
carenas: > we would need to add another member to `compile_block` to record the original value.

or just keep it in the stack as a local variable, no?; worst case you might need keeping track of the depth and I am sure there is a variable in there that could be reused.

it is not only the performance impact (remember that `compile_block` is in the stack to avoid doing malloc/free and the performance impact that has) of doing a `memmove`, but the fact that it is (most of the time) writing over the stack memory and that is also risky.

src/pcre2_compile.c
zherczeg: As far as I remember all patterns are enclosed in (?: ) by default.

doc/pcre2_compile.3
```diff
@@ -65,6 +65,7 @@ The primary option bits are:
                             theses (named ones available)
   PCRE2_NO_AUTO_POSSESS    Disable auto-possessification
   PCRE2_NO_DOTSTAR_ANCHOR  Disable automatic anchoring for .*
+  PCRE2_NO_PATTERN_REWRITE Disable pattern rewriting optimizations
```
mvorisek, line 68: In the future, there might be more rewrite optimizations. I would name this specifically for https://github.com/PCRE2Project/pcre2/issues/411 optimization only.

carenas: Disagree (at least in principle). As described in the discussion for that feature request this is indeed adding a completely new mode of operation for PCRE which was NEVER an optimizing compiler and where the normal use cases expected some other way to optimize the pattern (the same applies to JIT BTW which mostly translates the interpreter opcodes into native CPU code)

I "agree" that long term other "rewrites" might be added and would be a good idea to identify them here, but the way to do so would be to add flags for each case as extended options IMHO.  Indeed I would argue that this option SHOULDN'T be using one of the few options bits left and should be instead enabled through an extended option itself.

zherczeg:
Similar to other analysis, I would move this out of the pcre_compile. As a first step, I would introduce a `pcre_compile.h` which contains the internal definitions, and only included by similar optimizations. We could organize optimizations better in the future.

doc/pcre2_compile.3
zherczeg: I would use an O0-O3 system similar to gcc. The O1 is the default, and the behaviour is same as now. Then we could introduce optimizations, which costs more time, but improves performace more. We could also disable optimizations, such as auto-possession with O0. Disabling first character optimizations would be deprecated as well.

PhilipHazel: My view when Alex and I discussed this was that NO_PATTERN_REWRITE would disable *all* rewriting optimizations, present and future. If there is a need to disable individual ones, then additional extended options can be used.

PhilipHazel:
Are we ready to merge this yet?

alexdowad:
> Are we ready to merge this yet?

Absolutely not. :smile: There has been some great feedback from @carenas and others which I need to implement.

src/pcre2_compile.c
alexdowad: @zherczeg You are right, but that is done later in the compilation process. So if I don't strip redundant parens here, then it will mean every pattern will be enclosed in *two* sets of (? :).

src/pcre2_compile.c
alexdowad: > or just keep it in the stack as a local variable, no?; worst case you might need keeping track of the depth and I am sure there is a variable in there that could be reused.

That would also work, but we would need to keep that stack variable around all the way to the very end of `pcre2_compile` (currently line 11919-11920). If we do that, I think it's not any different from adding another member to `compile_block`.

testdata/testinput17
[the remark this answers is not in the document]
alexdowad: This is done. I will push an update soon.

doc/pcre2_compile.3
carenas: I think I wasn't clear enough and indeed might had confused the discussion further by mentioning the extended options, which I was proposing should be used to ENABLE each optimization, and therefore the `#define` and bit in the compile options that this thread is attached two could be reused.

Agree that the `(*NO_PATTERN_REWRITE)`  pattern option setting should disable all optimizations.

PhilipHazel:
I have marked this PR as draft.

src/pcre2_compile.c
carenas: > I think it's not any different from adding another member to `compile_block`

AFAIK `compile_block` is by design a "context" that gets passed around between possibly recursive calls, and therefore gets updated by them.

Unless I am not understanding your explanation correctly, it would seem that the original pointer to `cb->parsed_pattern` is constant and only would need to be updated in the parent function, no? so wouldn't it make more sense (by design) to keep it as a simple local variable?

alexdowad: I'll have to look at the code in detail and see exactly where the original pointer to `cb->parsed_pattern` is needed. Then I can give a better answer.

PhilipHazel: The "compile block", pointed at by the variable cb, contains data that is needed by more than one function in the compiling process, and is more than would be sensible to pass as multiple arguments. In a non-thread-safe world these would probably be static variables, but the PCRE2 functions (with the exception of the POSIX wrapper stuff) are thread-safe and therefore cannot use static variables. If a value is needed only in one function, it can just be on the local stack.

doc/pcre2_compile.3
zherczeg: What do you think about `(*OPT0)` - `(*OPT3)`? People in general don't really understand the effect of individual optimizations. O0-O3 is a simple way to say I don't mind a longer compilation time for faster execution.

We could add a new api to set the optimization level in compile context.

zherczeg:
Could you share your results about the optimization? Compile time overhead vs speed gain. Interpreter and jit is also interesting.

zherczeg:
I realized we are not just adding a new function here, we are also adding an entirely new subsystem. So we should discuss how similar optimizations should be handled in the future, because changing the api after a release is difficult.

My api proposal:
`pcre2_compile_set_opt(compile_context, optimization_level, disabled_optimizations);`

Most people don't know the effect of various optimizations, so they can simply set `disabled_optimizations` to 0. However, if a certain optimization cause issues in a certain system, they can disable that particular one.

The `optimization_level` can be 0-3, and performs a `disabled_optimizations |=` operation to add disabled optimizations internally. Hence the compile context does not store `optimization_level` directly.

It is also a good question how patterns could override the optimization level. Setting it to 0 may hurt system performance, and not necessary desired in all cases. Shall we add an option to `disabled_optimizations` which prevents overriding the optimization level? Or can only increase or decrease the optimization level.

src/pcre2_compile.c
alexdowad: This has been done, update not yet pushed.

src/pcre2_compile.c
alexdowad: Added comment here.

src/pcre2_compile.c
alexdowad: @carenas is right 😆 
Updated my PR accordingly. It's better this way.

alexdowad:
Although not all issues raised by reviewers have been resolved, I am pushing an update so any comments can refer to the latest version of the PR.

RunTest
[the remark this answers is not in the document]
alexdowad: OK, this is done.

doc/pcre2_compile.3
carenas: I guess that setting an optimization level in pattern with `(*OPT=n)` might be useful to end users that have a new enough library and no other way to enable them, but still think it would be better to enable them through an API call instead (or an extended option) based on the use cases spelled in the original request.

I also think it will be better to be disabled by default (your idea of level 1 being the default and including "autoposification" as a "rewrite like" optimization if you squint hard enough, might work) but I suspect it might be cleared to start with 0 and had this part of 1.

zherczeg: Could you describe your suggestion for the whole API? Including C and pattern string API. I think this is the most important part to decide, the rest can be changed internally between releases.

PhilipHazel: My hope was that this new optimization would be cheap enough that it could happen by default, just like pcre2_study(). There are two reasons for having a disabling option: (1) As an emergency safety net in case there are broken cases and (2) so that it's easy to make comparisons with and without the optimization. I am not really keen on optimization levels.

carenas: We had yet to see any performance numbers though, and I am also a little concerned by the increase on code size memory from the test output, but it might be just bugs and easy to nail down once we are back to having code that builds mostly warning free and without leaks.

Going back to Zoltan's API question, are you suggesting we punt the need for an API or extra flags until there is another optimization? or even until there are problems with one of them specifically?

alexdowad: > We [have] yet to see any performance numbers though...

That will be addressed soon enough. I am dealing with the issues raised one by one.

alexdowad:
Tomorrow I hope to look into the issue @carenas mentioned whereby the 'code size memory' has increased for some patterns. Will post my findings here.

doc/pcre2_compile.3
zherczeg: I would like all current and future optimization handled in the same way. There are only a few compile flags remained, and I would prefer reserving them for features, rather than optimizations. If we add 2-3 optimizations / year, that will be 20-30 in 10 years. Would be good if their generic interface is decided now, rather than each optimization is handled by different ways.

carenas: > My hope was that this new optimization would be cheap enough that it could happen by default, just like pcre2_study()

The problem though is that unlike `pcre_study()` the pattern itself could change and therefore basic things like `erroroffset` become unreliable and as pointed out in #460 could be even a source of crashes (another reason why I think it might not be a good idea to allow it by default).

Eitherway I agree with Zoltan that nailing the API first is worth doing with the first release (even better if with the first commit)

src/pcre2_compile.c
carenas: > If a value is needed only in one function, it can just be on the local stack.

That is what I meant by "design", and indeed why I was expecting it to be set as a local variable for pcre2_compile() where its scope is simpler to understand instead of something that _could_ be updated by any function that participates in the process of compiling a pattern.

It seems I wasn't "right" enough though as the implementation went in a different direction.

alexdowad: @carenas Yes, I discovered that it wasn't that easy. The problem was that if a rewrite is performed, the buffer which the rewritten pattern goes into *also* needs to be reliably freed.

doc/pcre2_compile.3
alexdowad: > The problem though is that unlike the `pcre*_study()` optimizations, the pattern itself could change and therefore basic things like `erroroffset` become unreliable and as pointed out in #460 could be even a source of crashes (another reason why I think it might not be a good idea to allow it by default).

What you have mentioned here is a very serious issue. I may put this PR 'on hold' to allow time to think about this. It could be a fatal flaw.

alexdowad: With regard to the `erroroffset` issue raised by @carenas: I think that with the existing PCRE2 API, pattern rewriting optimizations of the kind in this PR can be done *only* if all possible causes of errors (or at least, errors which require a value to be returned in `erroroffset`) are checked first. Pattern rewriting can only be done once we are sure that there is no possible reason to return an `erroroffset`.

alexdowad: Ah... 🤦🏻‍♂️ Sorry, scratch the above comment.

I've just remembered: the pattern rewriting phase added in this PR does not rewrite the *pattern string*; it rewrites the `parsed_pattern` vector which comes from the pattern string.

Some of the `parsed_pattern` items have associated data which includes offsets into the original pattern string; when adjusting the `parsed_pattern` vector, the pattern rewriter maintains all these offsets "as is", so that if any error is detected later in the compilation process, `erroroffset` will point to the appropriate part of the original pattern string.

(This might be a good debug assertion: run through the `parsed_pattern` vector and check that all pattern string offsets are within bounds.)

zherczeg: I thought the rewriting happens after the parsing is successful, so no error (besides out of memory) can happen.

alexdowad: > I thought the rewriting happens after the parsing is successful, so no error (besides out of memory) can happen.

You are right that rewriting happens after a successful parse... but there are just a few types of errors which are not detected during parsing, but rather when processing the `parsed_pattern` vector. For all such cases, offsets into the original pattern string are stored in the `parsed_pattern` vector, and the pattern rewriter makes sure not to destroy these.

carenas: The erroroffset issue was likely a red herring, but it is offtopic for this conversation which I think needs a resolution.

It would seem that @PhilipHazel likes the current one as it is sufficient and doesn't make a difference when there is only one rewriting optimization anyway and might want this discussion punted until later.

Note that the option flag mentioned here is to disable all rewriting.

zherczeg: Whatever we do, it should happen before the next release. There are only a few compile flags left, and this flag feel out of scope for me. What about making this always enabled in this patch, and make it conditional when we decide how to do it.

alexdowad: Definitely, this issue needs to be resolved. But even just for testing purposes, it's vital that we can disable the optimization. I think making it "always enabled" is not an option.

My next priority on this PR is to investigate why the 'code memory' size has increased for some patterns. After that will be to eliminate all compiler warnings. Other issues to be worked on include: enabling tests on Windows if appropriate, coming to a resolution on whether PCRE2 will move to the C99 standard or not, questions of brace style, and performance benchmarking. (Philip has already done a bit of benchmarking, and from his limited results, it appeared that this optimization is both 1) worthwhile and 2) has relatively small impact on compile time. But more thorough and extensive benchmarking is needed. It's possible that the benchmarking might cause me to do more performance work to reduce the impact on compile time.)

As all those matters fall into place one by one, at some point, the issue of how to disable the optimization must be worked out. I agree with Zoltan that the number of bits left in the compile options is truly small. But otherwise I haven't thought hard enough to weigh in on the issue. I just followed what has been done for previous optimizations.

I consider it extremely important that Zoltan, Carlo, and other key contributors who have helped to make PCRE2 what it is now are all happy with whatever is decided.

PhilipHazel:
Two comments:  (1) I was about to comment about erroroffset not being a problem, but that seems to have been explained by others. (2) How about adding a new 32-bit field to the compile context called, say, OPTIMIZE_SWITCHES, which contains a bit for each optimization that can be enabled/disabled. A function could then set them individually, or "levels" such as 0, 1, 2, etc could set a selection of them. Existing NO_xxx options would have to be kept, of course, for backward compatibility, but we could move the new NO_REWRITE into the new field.

alexdowad:
> How about adding a new 32-bit field to the compile context called, say, OPTIMIZE_SWITCHES, which contains a bit for each optimization that can be enabled/disabled...

👏🏻 

Nice. Very nice.

alexdowad:
> ...I would think of possibly using SIMD to test for 8 different byte values at once...

For x86, perhaps something like:

```c
    __m64 class = _mm_set_pi8(char1, char2, char3, char4, char5, char6, char7, char8);
    __m64 test = _mm_set1_pi8(subject_char);
    if (_mm_movemask_pi8(_mm_cmpeq_pi8(class, test)) {
        /* subject_char matches char class */
    }
```

In JIT'd code, hopefully it might be possible to load the MM register just once and keep the character class values there for the duration of the match.

zherczeg:
JIT has many optimizations for classes. E.g. `[Aa]` is converted to `(char | 0x20) == a`, or `[a-z]` is converted to `(char - a) <= z - a`.

PhilipHazel:
Prompted by the discussion with the Excel people I have actually had a few thoughts about refactoring how classes are handled, so I would suggest not doing anything drastic at this stage. I may write some ideas down - perhaps add them to issue #13.

alexdowad:
I have just fixed a couple of compiler warnings in my code (haven't yet pushed an update).

Carlo mentioned that "there are warnings triggered by CI"; I looked at the output of some of the CI build jobs but don't see where these warnings are. May I ask where is the right place to look?

carenas:
> For char classes which only have a small number of character values (say up to 8), I wouldn't be surprised if it would be faster _and_ make the regex bytecode smaller to represent the character values as literal bytes.

indeed in some cases the pattern writer might had optimized it by using alternations (usually atomic) and the pattern rewriting is undoing that work.

noticed that the original feature request didn't request for what it is described as a "different type of rewrite" above:

```
(a|b|c) ⇒ ([a-c])
```

where did that came from?; while there is definitely a lot that can be done to improve performance for classes I would expect this to be a pessimization (specially with UTF)

carenas:
> where these warnings are

next to the PR there is a mark that lists all the jobs, they are also visible from the "actions" menu on top.

The "Build" step of the windows job which might be accessible from this [link](https://github.com/PCRE2Project/pcre2/actions/runs/10758641124/job/29834287900) then shows stuff like:

```
D:\a\pcre2\pcre2\src\pcre2_compile.c(5350,1): warning C4715: 'find_group_end': not all control paths return a value [D:\a\pcre2\pcre2\build\pcre2-16-static.vcxproj]
D:\a\pcre2\pcre2\src\pcre2_compile.c(5363,1): warning C4715: 'find_class_end': not all control paths return a value [D:\a\pcre2\pcre2\build\pcre2-16-static.vcxproj]
```

alexdowad:
> where did that came from?; while there is definitely a lot that can be done to improve performance for classes I would expect this to be a pessimization (specially with UTF)

Philip asked me to add that feature because he felt that char classes should be faster than alternation. However, when thorough benchmarking is done, we will find out if that is true or not.

alexdowad:
> The "Build" step of the windows job which might be accessible from this [link](https://github.com/PCRE2Project/pcre2/actions/runs/10758641124/job/29834287900) then shows stuff like:

Ah, I see. I was just looking at the jobs for Linux. Thanks! :smile:

It looks like the Windows compiler (Visual C++, I guess) doesn't support `__builtin_unreachable` and that is the reason for the warning. Just doing some more research...

PhilipHazel:
I did some experiments with (a|b|c) => ([abc]) and in my case the latter was much faster, which was what I expected. It is just testing one character against a class, as opposed to setting up the testing of several branches with backtracking, and that involves initializing stack frames, etc, etc.

alexdowad:
Tomorrow I plan to enable some of the new tests on Windows. Because I don't own a Windows machine to test on, I might test the changes by pushing updates to this PR (and seeing what happens in CI). So... sorry if the reviewers get lots of push notifications.

PhilipHazel:
Here are some timings on my quite old Linux box:

```
/x(a|b|c)/
    xdxexfxc
Match time  0.3103 microseconds
 0: xc
 1: c

/x([abc])/
    xdxexfxc
Match time  0.2328 microseconds
 0: xc
 1: c

/x(?:a|b|c)/
    xdxexfxc
Match time  0.2730 microseconds
 0: xc

/x[abc]/
    xdxexfxc
Match time  0.1820 microseconds
 0: xc

```

carenas:
> Here are some timings on my quite old Linux box:

on the topic of classes being inefficient with UTF (another red herring), from my probably older MBP:

```
  re> /x[áàãäa]/BI,utf 
        Bra
        x
        [a\xe0\xe1\xe3\xe4]
        Ket
        End
Capture group count = 0
Options: utf
First code unit = 'x'
Subject length lower bound = 2
data> xä
Match time  0.0724 microseconds
 0: x\x{e4}
data> xa
Match time  0.0690 microseconds
 0: xa
```

And more importantly, to try to keep on topic, how the pattern rewriter actually helps:

```
  re> /x(?:a|b|c|d|e)/BI,utf
        Bra
        x
        [a-e]
        Ket
        End
Capture group count = 0
Options: utf
First code unit = 'x'
Subject length lower bound = 2
data> xa
Match time  0.0686 microseconds
 0: xa
data> xe
Match time  0.0686 microseconds
 0: xe
```

carenas:
> Because I don't own a Windows machine to test on, I might test the changes by pushing updates to this PR (and seeing what happens in CI)

The windows job was indeed added to help with that, and also test the code in 32-bit (which also isn't that common nowadays)

There is no need to spam this PR with your tests though, as the actions can be also enabled in your fork and run independently, on that line I also added a couple of PRs there that would be nice to see included in the next release.

A couple more things that would be nice to see are:

1. It would seem that the detection for not supported cases could be done earlier (in `parse_regex()`) and therefore `rewrite_regex()` could return early or not even be invoked in those cases.  Although a layering violation, it might be worth keeping track of how much of the pattern could be skipped and simply copied over to the new `pattern` buffer.
2. Changes to the API that were discussed earlier.

Let us know how to help more effectively with those, otherwise.

alexdowad:
> There is no need to spam this PR with your tests though, as the actions can be also enabled in your fork and run independently,

That's nice to hear. What do I need to do so that any new branches I push to my fork (even if no PR is open) will invoke a CI run?

carenas:
I did mine long ago, and so I might be misremembering, but there used to be a "big red button" under the "actions" menu in my repository that asked for actions to be enabled.

Which branches the configuration affects is defined in the "code" as part of what is inside the `.github` directory, and ours have indeed no restrictions at all to make it easier.

alexdowad:
Since this PR, as currently written, requires C99:

Philip sent an e-mail to several contributors earlier, asking whether anyone has comments on whether it would be a good idea to bump the minimum required version of the C language to the C99 standard. The gentleman who packages PCRE2 for Debian replied that he likes the idea, but others did not respond. Can I please ask for further comments on the issue?

If some contributors do not like this, or if there are too few responses, I can still rework the PR so that the code still builds with pre-C99 compilers.

PhilipHazel:
Actually, I think this train has already left the station. C99 features already exist in a number of files. I've just done a grep for "for \(int" (which is a C99 thing) and that usage is widespread. Nobody has complained, as far as I know.

alexdowad:
> Actually, I think this train has already left the station. C99 features already exist in a number of files.

Great, that is clear.

Next issue which needs some consensus: Should I update this PR to follow the existing brace style in PCRE2?

carenas:
> If some contributors do not like this, or if there are too few responses, I can still rework the PR so that the code still builds with pre-C99 compilers.

The issue was never about C99 but instead coding style and consistency of the code base, which is relevant for future maintenance.

Mixing declarations and code (what I was referring before as `-Wdeclaration-after-statement` issues) makes understanding the scope of variables harder since they are no longer all defined at the beginning of it, but can be useful sometimes, for example in the `for (int` example Philip referred to and that is used already and does indeed help with making the scope of that variable shorter and clearer. Some of the ones I see in the last version of the code are difficult to spot and could use extra empty lines either before or after them but there are also several that could be move up to the beginning of the scope to get rid of the warning just as much, for example this is IMHO clearer than the original and might have even uncovered a bug:

```
static inline void rewrite_buf_realloc(rewrite_buf *buf, size_t new_size, pcre2_memctl *memctl)
{
  uint32_t *expanded;
  ptr_diff current_size;

  PCRE2_ASSERT(buf->start != NULL && buf->p >= buf->start);
  expanded = memctl->malloc(new_size * sizeof(uint32_t), memctl->memory_data);
  current_size = buf->p - buf->start;
  memcpy(expanded, buf->start, current_size * sizeof(uint32_t))
  memctl->free(buf->start, memctl->memory_data);
  buf->start = expanded;
  buf->p = expanded + current_size;
  buf->limit = expanded + new_size;
} 
```

Adding functions with a different brace style (and trust me, everyone but Philip is a fan of the current one) also distracts reviewers and it should be avoided IMHO (I also added functions with a different style before, thinking it didn't matter and that was ok because it was self contained) hence with hindsight would suggest against doing it.

alexdowad:
> for example this is IMHO clearer than the original and might have even uncovered a bug:

Thanks for mentioning this, but what is the bug?

alexdowad:
I may create a separate PR to implement Philip's idea of adding a new `optimization_options` field to the compile context, and an API function for manipulating it.

One thing I am thinking about: other existing API functions use `uint32_t` for options arguments. (Such as the `extra_options` argument to `pcre2_set_compile_extra_options`.) It's unfortunate that the caller has to be explicitly aware of the size of the options field; it means that if the number of optimizations which can be enabled/disabled keeps growing, it won't be possible to extend the field to 64 bits without breaking callers.

Is there any way to design the API so that it *would* be possible to adjust the size of the field without breaking callers? (This also means maintaining binary compatibility, so that binaries which were compiled against one version of PCRE2 can be dynamically linked with a newer version without problems.)

alexdowad:
Hmm... one thought...
What if the new API function which sets optimization flags is declared to take

alexdowad:
Whoops... accidentally closed the PR when trying to delete a comment... sorry! 😊

zherczeg:
You have more options for a generic api:
pass an array of `uint32_t`, and a size. Since most people will use O0-O3, they simply pass `NULL, 0`. This is the by far the simplest approach.

We could also add a flag manipulation api.
set(context, index, value) -> also returns the previous value
get(context, index)
set_level(context, level) -> sets a default set of optimizations, useful as first step.

This way optimizations can be individually switched. Can be useful for complex systems.

PhilipHazel:
I agree with Carlo in disliking declarations in the middle of computational statements because they get lost. A declaration inside a "for" is different because it is sort of at the top of a block. As for the current brace style, I am of course a fan of it as I have been using it since the 1960s when I first discovered block structured languages (BCPL) but as PCRE2 goes forward with much less contribution from me, I am not going to argue either way.

alexdowad:
OK, I have adjusted this PR to use the "1960s BCPL" brace style. :wink: It also builds without `-Wdeclaration-after-statement` warnings.

alexdowad:
If my other PR re: Windows tests can be merged, that will help to keep pushing this one forward.

alexdowad:
OK, tests for this PR have also been added to `RunTest.bat` so they will run on Windows.

All the feedback and ideas on the new API for enabling/disabling optimizations have been much appreciated. At this point, if anyone is willing to offer more specific suggestions (i.e. function signatures), that will be great.

I do like Zoltan's suggestion that the new API function could accept an array/length pair, where each value in the array represents an optimization (or group of optimizations) which is switched on or off. These values would not be individual bits, but rather successive integers, so the namespace is very large. (Perhaps even declare them as an `enum`?) Internally, they could still be converted to a bitfield.

This should preserve both source-level and binary compatibility for code written against an older version of PCRE2, when users upgrade to a newer version.

How does that sound?

Can we call it something like `pcre2_set_optimization`?

zherczeg:
Most optimizations are a simple boolean, so I would prefer a boolean array. The other alternative is multiple call of the API function, where each call sets one optimization. We can also add more api functions later if needed. We have a context, so everything is backward compatible.

src/pcre2_compile.c
carenas: @alexdowad: I think the argument Zoltan was making was that there is a potential for tighter integration here that is being missed; could `parse_regex()` just add this so your code doesn't add them and then remove them?, could `compile_regex()` be told how to generate the original OP_BRA using those wrapping META instead?

zherczeg: Removing unnecessary bracekts are a good idea, and it should improve performance:
```
  re> /a(?s:.)b/B
        Bra
        a
        Bra
        AllAny
        Ket
        b
        Ket
        End
```

The nested Bra / Ket is totally unnecessary. The jit actually optimize them out. However, the interpreter still spends some time to process it, so removing them should be good.

alexdowad: > The nested Bra / Ket is totally unnecessary. The jit actually optimize them out. However, the interpreter still spends some time to process it, so removing them should be good.

I guess another option would be to strip out redundant OP_BRA/OP_KET *after* the `parsed_pattern` is converted to bytecode, rather than before.

Anyways, for the moment, I'm happy with this aspect of this PR.

alexdowad:
> A function such as pcre2_set_optimization() should have as its arguments (1) a pointer to the compile context, (2) the index into an array of 32-bit words (so that we can extend ad infinitum), (3) two 32-bit words, the first being bits to unset followed by the bits to set.

Dear Philip, I'm just trying to figure out how this interface will work. Why is there both an index into an array of 32-bit words, and also two 32-bit words with bits to set/clear? When you say there should be an index into an array of 32-bit words, how will this array work? What will the values in the array be?

To make this discussion a bit more concrete, I may open a PR today with a proposed interface.

carenas:
> if we were to make it a 64-bit field, we would ever run out? That makes for an easier interface.

frankly, I think that a 63-bit field must be enough (even maybe stick with `uint32_t`, although I can't think of any relevant architecture were a `int64_t` wouldn't be as available), after all if an optimization NEEDS to be disabled then it probably shouldn't exist anyway and so these bits will be mostly allocated when the optimization is developed and just to make sure that it can be disabled if found buggy or to compare with the unoptimized code, and hopefully retired before being made permanent.

With that the inteface will be instead:

```
#define PCRE2_COMPILE_OPTIMIZE_ALTERNATE (1l << 0)
void pcre2_disable_optimizations(pcre2_compile_context *, int64_t);
```

zherczeg:
Some optimizations may take more time, and less useful for simple patterns, and disabling it is ok. I would prefer a O0-O3 system for that, even if we don't use all levels at the moment. Enabling optimizations individually is more of a private system, because users unlikely know the details of them. Hence a bitset (uint32_t) array could be enough in long term.

carenas:
> Why is there both an index into an array of 32-bit words

Because since you wanted an infinitely large bitset vector, then the mask to enable/disable has to know at which offset on that vector to apply the passed values.

> and also two 32-bit words with bits to set/clear? 

because the API is called `_set_`, so doesn't assume either that all optimizations are enabled by default unlike my "simplified" suggestion.

I agree with Philip though that doing  `_set_` would be better for this API but wanted to avoid the confusion this comment hopefully clarifies.

alexdowad:
Thanks to all for the suggestions; it is helpful.

I've opened a (draft) PR adding an API which I believe is better than the kind suggestions which the reviewers have shared. It currently has Zoltan's -O0 and -O3.

alexdowad:
I have rebased this PR on the latest `master`. Before pushing an update, I would like to hear opinions from others: Do you agree that `(*NO_PATTERN_REWRITE)` can be removed?

For all (or basically all) of the current compile option flags, it is possible to set them using syntax like `(*NO_JIT)`, etc. at the beginning of a regex. This syntax is converted to META_OPTIONS by the parser.

If more optimization flags are added in the future, I doubt if we should keep adding more and more of such options syntax. Personally, it doesn't seem necessary that users should be able to set any possible optimization flag using such syntax. So I'm thinking there may be no need to support an inline `(*NO_PATTERN_REWRITE)` option.

carenas:
> Do you agree that `(*NO_PATTERN_REWRITE)` can be removed?

No; the in pattern VERB are there to allow disabling the option for final users of the library who might not have access to the C API to do so, and even for the ones that do, it might need to be disabled for specific patterns because it is not a good option (compilation time can be significantly increased for an already optimized pattern) or has bugs that are triggered by it.

PhilipHazel:
The original motivation for introducing (*NO_JIT) etc was to allow users who have no access to the C code to change these options. In other words, users of some application that allows them to provide patterns might want or need to set such things. In particular (*UTF) and (*UCP) might be necessary - though these can be locked out by the application. I can see that there those that relate to optimizations are somewhat different, but it's not really much code in pcre2_compile.c. We cannot remove the old ones without provoking complaints about backwards compatibility, and if we don't add NO_PATTERN_REWRITE somebody is sure to grumble. So I would add it, for an easy life, but what do others think?

alexdowad:
> One important thing to clarify here too, is WHAT is being disabled

I would say that for now, it's disabling all rewrites. If there is need to add more fine-grained options, that can be done later.

alexdowad:
Looks like I need to understand Zoltan's new code for handling character classes. The code in this PR for rewriting char classes worked on the previous `master`, but doesn't look like it works now.

zherczeg:
Am I understand that rewriting happens using META code? My new code is running in META -> BYTE CODE translation phase, and should not interact with rewriting, since it happens later. Also, I have introduced a `pcre_compile.h`, and added new `.c` file for doing range optimization, and I would like that rewriting optimization also goes to another `.c` file. The `pcre_compile.c` is big enough.

alexdowad:
> Am I understand that rewriting happens using META code?

Indeed, it does. So as you said, it *seems* that the new code should not have an effect here. But I'll investigate just to make sure.

> I would like that rewriting optimization also goes to another `.c` file. The `pcre_compile.c` is big enough.

No objection, though it means that it will not be possible to make the entry point `static`. I guess it will be more in line with the existing style of the project as well (since DFA matching is in a separate file, the interpreter is in a separate file, etc).

carenas:
> I guess it will be more in line with the existing style of the project as well

Not really, and I think this point is worth clarifying, because unlike most codebases we do have a lot of static functions in C files that are included into other C files.

All that Zoltan is suggesting is to move your code to a different file, if you MUST keep your entrypoint static (and I can't see why, since it is not an exported symbol anyway, it is called only once and is probably too big to inline) then you could by following our existing style.

alexdowad:
> Not really, and I think this point is worth clarifying, because unlike most codebases we do have a lot of static functions in C files that are included into other C files.

Ooh, this is interesting. Thanks for the tip.

alexdowad:
Just updated this PR to work on the latest `master`, though it needs to be fuzzed again.

Tests are passing, except for "check autogenerated file freshness" task which is complaining about "noise" changes, like changes in the spaces between words in `pcre2.txt`, which are not caused by this PR. (It would be helpful if the `PrepareRelease` script could be made to generate text files with consistent spacing, so that diffs are not cluttered with irrelevant changes.)

NWilson:
Yes, I apologise. It's due to the `nroff` tool using different justification rules at different versions. I've standardised on the version that's in a Ubuntu 24.04LTS just for convenience in the CI. It is a pain though. I'll see what I can do.

alexdowad:
> Yes, I apologise. It's due to the `nroff` tool using different justification rules at different versions. I've standardised on the version that's in a Ubuntu 24.04LTS just for convenience in the CI. It is a pain though. I'll see what I can do.

Dear @NWilson, this is quite enlightening. Do you happen to know what version of `nroff` that is? The version I have locally is GNU nroff 1.22.4, but I can build another version if it helps to generate files which match those generated  by CI.

NWilson:
Apparently it's 1.23.0: https://packages.ubuntu.com/search?keywords=groff&searchon=names&suite=all&section=all

(I'm travelling at the moment) 

I have a set of Linux VMs for testing features of different distributions, with my home directory mounted and shared across them. I use the Ubuntu 24.04 one for running PCRE2 commands. 

It shouldn't be needed though. We really ought to make it less sensitive to tool versions.

alexdowad:
I just cloned the repo for GNU nroff and built version 1.23.0, but its output is quite different from the documentation files which are currently checked into the PCRE2 repository.

For now, I am considering the above failure of the "Check autogenerated file freshness" CI task to be spurious.

alexdowad:
Closing due to inactivity.

PhilipHazel:
Alex, I would like to put on record my thanks to you for all the work you put into this. As you probably know, I am no longer active in PCRE2 development, so it's up to the current team whether to pursue the original idea further. What I learnt from the exercise is that "just pull out the common prefixes" isn't nearly as straightforward as it sounds, and is quite limited in scope, which begs the question "is it worth it?". A question that is no longer mine to answer!

NWilson:
Yes, thank you Alex for this contribution! I won't forget about this work, although I haven't had time to spend on it recently. I regret that we weren't able to provide more input to help it along.

The issue #411 remains open on our backlog, so that will serve as a reminder for us (me?) to revisit this at some point in the future.

I will consider if there's an easy an effective way to apply this optimisation. As Philip notes, it may ultimately not be useful in general situations, but if it is, we would gladly complete it.

PR #465, opened 2024-09-10
Implement PCRE2_UNREACHABLE assertion for MS Visual C++

alexdowad:
Carlo pointed out that my code in a different PR was causing compiler warnings on Windows. The root cause is that `PCRE2_UNREACHABLE()` was not implemented on Windows.

Have I missed any other files which need to be updated?

### DIFF
--- a/src/pcre2_convert.c
+++ b/src/pcre2_convert.c
@@ -1159,7 +1159,6 @@ for (i = 0; i < 2; i++)
   }
 
 PCRE2_UNREACHABLE(); /* Control never reaches here */
-return PCRE2_ERROR_INTERNAL;
 }
 
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,6 +187,11 @@ CHECK_C_SOURCE_COMPILES([=[
 set(CMAKE_REQUIRED_FLAGS ${ORIG_CMAKE_REQUIRED_FLAGS})
 
 CHECK_C_SOURCE_COMPILES(
+  "int main(void) { __assume(1); return 0; }"
+  HAVE_BUILTIN_ASSUME
+)
+
+CHECK_C_SOURCE_COMPILES(
   "int main(void) { if (__builtin_expect(1+1, 2)) return 1; return 0; }"
   HAVE_BUILTIN_EXPECT
 )
--- a/config-cmake.h.in
+++ b/config-cmake.h.in
@@ -1,6 +1,7 @@
 /* config.h for CMake builds */
 
 #cmakedefine HAVE_ASSERT_H 1
+#cmakedefine HAVE_BUILTIN_ASSUME 1
 #cmakedefine HAVE_BUILTIN_EXPECT 1
 #cmakedefine HAVE_BUILTIN_MUL_OVERFLOW 1
 #cmakedefine HAVE_BUILTIN_UNREACHABLE 1
--- a/configure.ac
+++ b/configure.ac
@@ -93,6 +93,20 @@ fi
 CFLAGS=$tmp_CFLAGS
 AC_LANG_POP([C])
 
+# Check for the assume() builtin
+
+AC_MSG_CHECKING([for __assume()])
+AC_LANG_PUSH([C])
+AC_LINK_IFELSE([AC_LANG_PROGRAM([[]], [[__assume(1)]])],
+  [pcre2_cc_cv_builtin_assume=yes],
+  [pcre2_cc_cv_builtin_assume=no])
+AC_MSG_RESULT([$pcre2_cc_cv_builtin_assume])
+if test "$pcre2_cc_cv_builtin_assume" = yes; then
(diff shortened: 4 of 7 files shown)


carenas (approved):

src/pcre2_convert.c
```diff
@@ -1159,7 +1159,6 @@ for (i = 0; i < 2; i++)
   }
 
 PCRE2_UNREACHABLE(); /* Control never reaches here */
-return PCRE2_ERROR_INTERNAL;
```
carenas, line 1162: why was this return call removed?

this is a public API function and while this codepath shouldn't be normally reached, returning a random value instead of an error is likely to cause more problems and indeed it mostly reverts a previously added fix without explanation.

alexdowad: I think the reasoning was something like: this is an unreachable point in the code. So having the `return` statement there cannot help anything, since it will never be executed.

carenas: PCRE2_UNREACHABLE() could be defined as a NOP as well (indeed it was for MSVC, which is why I added the return back in the previous "fix"), and therefore if this point is ever reached, code execution will just exit the function with a random return value.

Please see #465 for a real fix, and sorry I didn't catch this (not mentioned in the commit message or PR) change before it got merged, but next time it will be IMHO better to ask why the previous fix that you reverted was there to begin with.

alexdowad: > PCRE2_UNREACHABLE() could be defined as a NOP

That is true.

I don't think this point is worth debating, but before 'fixing' anything, I think a good first question to consider might be: Is this point in the code actually unreachable, or not? If it is truly unreachable, then it would be best not to merge #465. If it is reachable, then the next question would be: Why is it reachable? Is there nothing we can do to prevent that?

Just a thought.

carenas: I don't think that returning a random value because of a bug that might make it reachable is a valid way to build an API.

Please, take your time to read that PR and the proposed changes, and if you still think that returning a random value instead of PCRE2_ERROR_INTERNAL (which is documented and used by other APIs to report those kind of errors already) lets discuss it there.

alexdowad: > Please, take your time to read that PR and the proposed changes

Yes, I did that before commenting.

> Please, take your time to read that PR and the proposed changes, and if you still think that returning a random value instead of PCRE2_ERROR_INTERNAL...

Definitely, there is no way that I could advocate for returning a random value. What I would advocate for is understanding the code well enough, and testing it well enough, that we are certain that our assertions, invariants, etc. are true.

This isn't just about functions where the end of the body is unreachable. In PCRE2, there are probably thousands of pointer dereferences. Every one of those *could* conceivably become a NULL dereference **if** there is a bug somewhere. But we don't guard every pointer dereference with `if (ptr == NULL) return PCRE2_ERROR_INTERNAL;` Rather, both by design and by testing, we try to ensure that the code works as intended and we don't have surprise NULL pointers.

But as I said, the issue of whether to include `return PCRE2_ERROR_INTERNAL` or not is not big enough to be worth debating. If you prefer to have it there, I agree 100% :+1:

What I do not personally agree with, and encourage you to reconsider, is deleting the assertion in that function. If you do so, it will be more likely that bugs will creep in and then users will actually receive `PCRE2_ERROR_INTERNAL`, rather than the bugs being caught and fixed before they get into a public release.

alexdowad: Ah, Carlo, here's another (tangentially related) thought...

You mentioned that before I added the implementation of `PCRE_UNREACHABLE` which is based on `__assume`, it was a no-op with VC++.

How about this: If the target host doesn't have `__builtin_unreachable`, and it doesn't have `__assume`, and if PCRE2_DEBUG is defined, how about we implement `PCRE2_UNREACHABLE` using `abort()`?

That would ensure that *all* debug builds of PCRE2, regardless of compiler, will abort with an error if "unreachable" code is reached.

carenas: > That would ensure that _all_ debug builds of PCRE2, regardless of compiler, will abort with an error if "unreachable" code is reached.

I am finally understanding why those asserts looked so strange to begin with; you never understood the implementation well and only meant them as a debug tool.

And now I am glad I didn't catch this and had to propose a different solution instead.

Will propose another PR to try to address that, and hopefully this time that would be tested better before merging.

alexdowad: > Although I am starting to believe you really considered worth exposing production PCRE2 users to the additional crashes it would trigger?

In answer to this question, of course it's not my intention to expose users of PCRE2, or any other OSS project, to crashes. Rather, the intention is that by adding debug assertions to this project, bugs can be more readily found and fixed *before* they go into production. To achieve that goal, the assertions also need to be combined with thorough fuzzing (or use of other powerful techniques, like concolic testing, which find code paths which would lead to an assertion failure).

PR #468, opened 2024-09-11
Updates to Windows tests (RunTest.bat)

alexdowad:
The comments in RunTest.bat state that @zherczeg added 16-bit and 32-bit support... could he confirm that 544dfb00 reflects his intentions correctly?

I was able to run the CI jobs in my forked repo and the Windows 32-bit job passed. I guess that's a good thing... though since I have never seen it fail, I don't really know for sure that it is testing anything 😅

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -67,6 +67,28 @@ jobs:
       - name: Test (pcre2posix program)
         run: ./pcre2posix_test -v
 
+  bat:
+    name: MSVC
+    runs-on: windows-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - name: Configure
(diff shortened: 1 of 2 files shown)


carenas (approved):
To help test (through the CI) I am suggesting to merge alexdowad#3 on top.

you might want to at least squash the RunTest.bat changes on your commit though if want to keep them all as independent commits

PhilipHazel:
Is this ready now?

carenas:
Tested locally as well.

alexdowad:
> Is this ready now?

Yes!

carenas:
> > Is this ready now?
> 
> Tested locally as well using the cmdline but it seems to fail with MSVC, not sure if it is a regression though.

Indeed, not a regression, but fixing tests for the 32-bit library when the 16-bit library was not enabled

PhilipHazel:
I merged; there now seems to be a test failure...

alexdowad:
> I merged; there now seems to be a test failure

I'm looking into it. Thanks for mentioning.

alexdowad:
Hmm. I can't see any failure in the GitHub Actions tab. Is there somewhere else where I can see it?

carenas:
> Hmm. I can't see any failure in the GitHub Actions tab. Is there somewhere else where I can see it?

Should be visible [here](https://github.com/PCRE2Project/pcre2/actions/runs/10844916826/job/30094701046) (the red X next to the last commit in the "code menu")

The failure wasn't caused by this code though but just a badly timed deprecation.

PR #472, opened 2024-09-16
Add test and fix for too-large octals starting with \8 or \9

NWilson:
For backreferences of the form `\8000000` (larger than GROUP_MAX but smaller than INT_MAX) the regex was rejected as invalid, due to an out-of-range backreference.

However, for those of the form `\80000000000` (larger than INT_MAX) it would treat it as a escape for the literal character '8', which is inconsistent with Perl (confirmed by test).

This appears to be just an oversight in the code, and there's no test case so this condition wasn't covered.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -1349,7 +1349,10 @@ Arguments:
   ptrptr      points to the character pointer variable
   ptrend      points to the end of the input string
   allow_sign  if < 0, sign not allowed; if >= 0, sign is relative to this
-  max_value   the largest number allowed
+  max_value   the largest number allowed;
+              you must not pass a value for max_value larger than
+              INT_MAX/10 - 1 because read_number() relies on max_value to
+              avoid integer overflow
   max_error   the error to give for an over-large number
   intptr      where to put the result
   errcodeptr  where to put an error code
@@ -1901,17 +1904,19 @@ else
 
       /* As we know we are at a digit, the only possible error from
       read_number() is a number that is too large to be a group number. In this
-      case we fall through handle this as not a group reference. If we have
-      read a small enough number, check for a back reference.
+      case we treat the group number as too-large (since it may be larger than
+      INT_MAX we cannot return it for the caller to check).
 
       \1 to \9 are always back references. \8x and \9x are too; \1x to \7x
       are octal escapes if there are not that many previous captures. */
 
-      if (read_number(&ptr, ptrend, -1, INT_MAX/10 - 1, 0, &s, errorcodeptr) &&
-          (s < 10 || oldptr[-1] >= CHAR_8 || (unsigned)s <= bracount))
+      if (!read_number(&ptr, ptrend, -1, MAX_GROUP_NUMBER, 0, &s, errorcodeptr))
+        s = INT_MAX;
+
+      if (s < 10 || oldptr[-1] >= CHAR_8 || (unsigned)s <= bracount)
         {
         if (s > (int)MAX_GROUP_NUMBER) *errorcodeptr = ERR61;
-          else escape = -s;     /* Indicates a back reference */
+        else escape = -s;     /* Indicates a back reference */
         break;
         }
 
--- a/testdata/testoutput2
+++ b/testdata/testoutput2
@@ -14723,7 +14723,7 @@ Failed: error 161 at offset 9: subpattern number is too big
 No match
 
 /(?(1)()\983040\2)/
-Failed: error 161 at offset 14: subpattern number is too big
+Failed: error 161 at offset 13: subpattern number is too big
 
 /(*LIMIT_MATCH=)abc/
 Failed: error 160 at offset 14: (*VERB) not recognized or malformed
--- a/testdata/testinput1
+++ b/testdata/testinput1
@@ -6421,7 +6421,28 @@ ef) x/x,mark
 
 /\214748364/
     >\x{8c}748364<
-    
+
+# larger than GROUP_MAX, smaller than INT_MAX
+/a\800000b/
+
+# coming up on INT_MAX... (used to succeed with \8 being literal 8)
+/a\800000000b/
+
+# over INT_MAX (used to succeed with \8 being literal 8)
+/a\8000000000b/
+
+# smaller than GROUP_MAX
+/\21300/
+    \x8b00
+
+# larger than GROUP_MAX
+/\213000/
+    \x8b000
+
+# larger than INT_MAX
+/\21300000000/
+    \x8b00000000
+
 /a{65536/
     >a{65536<
 
--- a/testdata/testoutput1
+++ b/testdata/testoutput1
@@ -10167,7 +10167,34 @@ No match
 /\214748364/
     >\x{8c}748364<
  0: \x8c748364
-    
+
+# larger than GROUP_MAX, smaller than INT_MAX
+/a\800000b/
+Failed: error 161 at offset 7: subpattern number is too big
+
+# coming up on INT_MAX... (used to succeed with \8 being literal 8)
+/a\800000000b/
+Failed: error 161 at offset 7: subpattern number is too big
+
+# over INT_MAX (used to succeed with \8 being literal 8)
+/a\8000000000b/
+Failed: error 161 at offset 7: subpattern number is too big
+
+# smaller than GROUP_MAX
+/\21300/
+    \x8b00
+ 0: \x8b00
+
+# larger than GROUP_MAX
+/\213000/
+    \x8b000
+ 0: \x8b000
+
+# larger than INT_MAX
+/\21300000000/
+    \x8b00000000
+ 0: \x8b00000000
+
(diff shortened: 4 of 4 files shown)


src/pcre2_compile.c
```diff
@@ -1900,17 +1903,20 @@ else
...
-      read a small enough number, check for a back reference.
+      case we treat the group number as too-large (since it may be larger than
+      INT_MAX we cannot return it for the caller to check).
 
       \1 to \9 are always back references. \8x and \9x are too; \1x to \7x
       are octal escapes if there are not that many previous captures. */
 
-      if (read_number(&ptr, ptrend, -1, INT_MAX/10 - 1, 0, &s, errorcodeptr) &&
-          (s < 10 || oldptr[-1] >= CHAR_8 || s <= (int)cb->bracount))
+      if (!read_number(&ptr, ptrend, -1, MAX_GROUP_NUMBER, 0, &s, errorcodeptr))
```
NWilson, line 1912: I took the opportunity to reduce the limit from INT_MAX/10 to MAX_GROUP_NUMBER, although it makes no difference really.

The only (minor) impact is that the error is now reported on the location of the first character exceeding MAX_GROUP_NUMBER, which makes the location of the error more consistent with the other places `read_number()` is used. But arguably anyway, `read_number()` should perhaps advance the pointer to end of the run of digits when reporting a "too many digits" error, instead of just leaving pointer in the middle of the digit sequence.

src/pcre2_compile.c
```diff
@@ -1900,17 +1903,20 @@ else
...
+        {
+        s = INT_MAX;
+        }
+      if (s < 10 || oldptr[-1] >= CHAR_8 || s <= (int)cb->bracount)
```
NWilson, line 1916: Values with `s < 10` and `s <= bracount` are unaffected. This fix only affects the escapes `\8xxxx` and `\9xxxx` which must be interpreted as backreferences, regardless of whether read_number succeeds or fails.

testdata/testoutput2
```diff
@@ -14716,7 +14716,19 @@ Failed: error 161 at offset 9: subpattern number is too big
 No match
 
 /(?(1)()\983040\2)/
-Failed: error 161 at offset 14: subpattern number is too big
+Failed: error 161 at offset 13: subpattern number is too big
+
+# larger than GROUP_MAX, smaller than INT_MAX
+/a\800000b/
```
NWilson, line 14722: No change - test passes with no difference to expected

src/pcre2_compile.c
```diff
@@ -1900,17 +1903,20 @@ else
...
 
       \1 to \9 are always back references. \8x and \9x are too; \1x to \7x
       are octal escapes if there are not that many previous captures. */
 
-      if (read_number(&ptr, ptrend, -1, INT_MAX/10 - 1, 0, &s, errorcodeptr) &&
-          (s < 10 || oldptr[-1] >= CHAR_8 || s <= (int)cb->bracount))
+      if (!read_number(&ptr, ptrend, -1, MAX_GROUP_NUMBER, 0, &s, errorcodeptr))
+        {
+        s = INT_MAX;
+        }
```
zherczeg, line 1915: No need braces here.

zherczeg (approved):
LGTM with minor fix

testdata/testoutput2
carenas: the offset is incorrect though, should be 8 no?

src/pcre2_compile.c
PhilipHazel: What is the problem with ALT_BSUX here? That option is supposed to apply only to \U, \u and \x.

carenas (approved):
after thinking it further, the bad offsets should probably be addressed independently, as the change here is just a "sideeffect" of the limit change anyway and they are already incorrect.

src/pcre2_compile.c
carenas: After looking at it further, I think the limit should be reverted to INT_MAX/10 - 1.

We would get into this code also if the octal we are reading is next to other numbers and will incorrectly fail (ex: /\\214748364/), also the check after for the returned value being larger than MAX_GROUP_NUMBER makes no sense otherwise.

carenas (changes requested):
revert `max_value` change in `read_number` to make result consistent with further check for GROUP_MAX (when needed)

testdata/testoutput2
carenas: The changes required weren't as intrusive as I feared so they are included in the alternative PR to solve this issue.

src/pcre2_compile.c
NWilson: > After looking at it further, I think the limit should be reverted to INT_MAX/10 - 1.

I'm not convinced. INT_MAX is a completely arbitrary non-semantic limit, that's just a random number unrelated to the true limit of MAX_GROUP_NUMBER. Exposing the size of `int` (so that 5,000,000,000 is treated differently to 1,000,000,000) is strange - the regex grammar doesn't in any way depend on the size of int!

The original code was extremely close to being correct: it used `read_number` to peek for a decimal, without touching the `errorcodeptr`. We don't need to chage that.

NWilson:
> revert max_value change in read_number to make result consistent with further check for GROUP_MAX (when needed), the following is a valid test that would be broken otherwise:

You test passes on my branch. I've added your test case to the PR.

There's no need to reference INT_MAX anywhere in this code, because the behaviour does change on reaching 2,000,000,000 (values below and above this should be treated the same).

PhilipHazel:
So how do we decide between #472 and #476 if they both do the same job?

NWilson:
That's your job as project leader, I guess :) :)

Personally, I prefer my fix for the reasons I've commented - smaller diff compared to the original code; one less duplicated copy of the "char >= 8" logic; and thirdly avoids behaviour around INT_MAX which isn't relevant to the user.

But, of course I won't be offended if you prefer Carlo's version for any reason. I don't really care, as long as it works the same.

carenas:
> Personally, I prefer my fix for the reasons I've commented - smaller diff compared to the original code; one less duplicated copy of the "char >= 8" logic; and thirdly avoids behaviour around INT_MAX which isn't relevant to the user.

My fix is indeed much smaller and complete (as it also includes optionally fixing the strange erroroffset).

I was originally working on top of this but soon realized by looking at the history of the changes, that the bug was probably introduced while working around previous bugs and so minimizing the amount of change was wiser.

Agree that duplicating the 'char >= 8' is "unclean" but THAT is the bugfix. Resetting the ptr was done for a fallback so we can read the number as an octal, but obviously and early exit was needed if the number wasn't and that combination was fatal.

Agree that moving the limit out of INT_MAX will be better as well which is why that change is included as an additional optional commit, which can't be cleanly backported though as asserts are only available in master.

One additional concern I have about this proposal is that setting `s` to INT_MAX on error instead of just stop ignoring the error is a little more risky long term.

src/pcre2_compile.c
carenas: > The original code was extremely close to being correct: it used read_number to peek for a decimal, without touching the errorcodeptr. We don't need to chage that.

The bug was likely introduced in 8144ae04, without realizing that because `errorcodeptr` was no longer being set, the break when `c == CHAR_8` was no longer propagating that upstream.

carenas:
> > revert max_value change in read_number to make result consistent with further check for GROUP_MAX (when needed), the following is a valid test that would be broken otherwise:
> 
> You test passes on my branch.

Apologies, was testing a modified version of your branch which failed because of my own changes and the fact that s was getting the bogus INT_MAX value instead.

> I've added your test case to the PR.

Perl compatible changes should go to testinput1.

carenas (approved):
Third commit should be moved to the Perl compatible test or deleted, squashing all commits preferred as well.

@PhilipHazel: if merging this instead of #476, will need additional changes to fix the bogus erroroffset though

NWilson:
Thanks @carenas, I've rebased, squashed, and moved the tests to `test{input/output}1`.

@PhilipHazel We've spent (perhaps) too long on this simple bugfix. Can we merge both PRs - this one first, then Carlo's PR second with just bugfix for the `read_number()` ptr offset.

Thank you Philip, by the way, for approving and documenting all these PRs so quickly! I'll be working on PCRE2 for maybe 1-2 weeks, and it will all quieten down again.

carenas:
> We've spent (perhaps) too long on this simple bugfix. Can we merge both PRs - this one first, then Carlo's PR second with just bugfix for the `read_number()` ptr offset.

There is no cherry-pick of commits either and of course the other PR will have conflicts if this one gets merged, but I will rebase it on top and repurpose it as a "fixup" PR so Philip can apply it cleanly then, as needed.

carenas:
Thanks and sorry for the misunderstandings and extra work

NWilson:
Thank you @PhilipHazel and @carenas for the fixups! I'm doing my best to match the style & idioms as I observe them.

At least we're allowed C99 :)

PR #473, opened 2024-09-16
Add Python-style \1 backreferences in pcre2_substitute()

NWilson:
Previously, using `\1` inside a pcre2_substitute string would cause a bad-replacement error.

Now, we support this syntax in the same way as Python, to mean the same as `${1}`.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -1576,7 +1576,8 @@ Arguments:
   errorcodeptr   points to the errorcode variable (containing zero)
   options        the current options bits
   xoptions       the current extra options bits
-  isclassorsub   TRUE if in a character class or called from pcre2_substitute()
+  bracount       the number of capturing parentheses encountered so far
+  isclass        TRUE if in a character class
   cb             compile data block or NULL when called from pcre2_substitute()
 
 Returns:         zero => a data character
@@ -1587,8 +1588,8 @@ Returns:         zero => a data character
 
 int
 PRIV(check_escape)(PCRE2_SPTR *ptrptr, PCRE2_SPTR ptrend, uint32_t *chptr,
-  int *errorcodeptr, uint32_t options, uint32_t xoptions, BOOL isclassorsub,
-  compile_block *cb)
+  int *errorcodeptr, uint32_t options, uint32_t xoptions, uint32_t bracount,
+  BOOL isclass, compile_block *cb)
 {
 BOOL utf = (options & PCRE2_UTF) != 0;
 BOOL alt_bsux =
@@ -1817,7 +1818,7 @@ else
     */
 
     case CHAR_g:
-    if (isclassorsub) break;
+    if (isclass) break;
 
     if (ptr >= ptrend)
       {
@@ -1839,7 +1840,7 @@ else
       PCRE2_SPTR p = ptr + 1;
 
       while (p < ptrend && (*p == CHAR_SPACE || *p == CHAR_HT)) p++;
-      if (!read_number(&p, ptrend, cb->bracount, MAX_GROUP_NUMBER, ERR61, &s,
+      if (!read_number(&p, ptrend, bracount, MAX_GROUP_NUMBER, ERR61, &s,
           errorcodeptr))
         {
         if (*errorcodeptr == 0) escape = ESC_k;  /* No number found */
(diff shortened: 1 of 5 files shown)


src/pcre2_substitute.c
```diff
@@ -875,6 +879,12 @@ do
         goto LITERAL;
 
         default:
+            if (rc < 0)
+              {
+              special = 0;
+              group = -rc;
+              goto GROUP_SUBSTITUTE;
```
NWilson, line 886: This sort of goto-based code leaves me terrified. Surely all the common blocks of code should be pulled out into functions instead of gotos! Ouch.

But for a minimal fix, I added another goto. The danger is that we need to be very, very sure that we are jumping past the initialisation of any important stack variables.

I have checked as carefully as I can, that only `special` and `group` need to be set for the jump (and `name` would be required if `group < 0`).

src/pcre2_compile.c
```diff
@@ -1587,8 +1588,8 @@ Returns:         zero => a data character
 
 int
 PRIV(check_escape)(PCRE2_SPTR *ptrptr, PCRE2_SPTR ptrend, uint32_t *chptr,
-  int *errorcodeptr, uint32_t options, uint32_t xoptions, BOOL isclassorsub,
-  compile_block *cb)
+  int *errorcodeptr, uint32_t options, uint32_t xoptions, uint32_t bracount,
+  BOOL isclass, compile_block *cb)
```
NWilson, line 1592: We need `bracount` now in the case when `cb == NULL`

PhilipHazel:
I see that there is a compiler warning which has prevented the Dev action from completing. Otherwise, I think this is uncontentious.

src/pcre2_substitute.c
NWilson: The gotos in PCRE2 would mostly be fine by 2024 standards, you're right. Certainly the error-handling ones at the end of the functions are OK in "modern C" (if that's even a thing...).

I'm not judging any existing code, really. It's my own goto that I've just added that frightens me the most, although I am confident, after staring at it for a while.

Tweaked code to silence the warning, which was due to the poor compiler not being able to reason about which variables were initialised before use (due to the new goto!)

carenas:
Since you are also a new contributor might not be aware that since this code was originally managed with subversion there are really no merges, so if you want to have a coherent commit message it will be better if you squash your changes (specially considering that the second commit is just a "fixup").

src/pcre2_compile.c
```diff
@@ -1907,7 +1908,7 @@ else
       are octal escapes if there are not that many previous captures. */
 
       if (read_number(&ptr, ptrend, -1, INT_MAX/10 - 1, 0, &s, errorcodeptr) &&
-          (s < 10 || oldptr[-1] >= CHAR_8 || s <= (int)cb->bracount))
+          (s < 10 || oldptr[-1] >= CHAR_8 || (unsigned)s <= bracount))
```
carenas, line 1911: minor nitpick: use uint32_t instead of unsigned

testdata/testoutput2
```diff
@@ -14948,7 +14948,19 @@ Failed: error -55 at offset 3 in replacement: requested value is not set
 
 /abc/substitute_extended,replace=>\845<
     abc
- 1: >845<
+Failed: error -49 at offset 5 in replacement: unknown substring
```
carenas, line 14951: isn't this a regression?

src/pcre2_compile.c
NWilson: I think `unsigned` is better. The variable `s` is declared as `int`, and it is known to be non-negative, so the correct conversion to widen it to the corresponding unsigned type is `unsigned`. On a (hypothetical) architecture with 64-bit `int`, casting to `uint32_t` would be incorrect.

NWilson:
> Since you are also a new contributor might not be aware that since this code was originally managed with subversion there are really no merges, so if you want to have a coherent commit message it will be better if you squash your changes (specially considering that the second commit is just a "fixup").

Thanks that's helpful, I've rebased & squashed the commits. I was assuming that the PR would be merged by squash merge, with the commit comment taken from the title of the PR.

carenas (approved):

PhilipHazel:
I had a worry that introducing this would be backwards incompatible because of the octal ambiguity, but then I remembered that octal in substitution strings is broken in the current release, so nobody out there can actually be using it. I note that currently it has the Perl ambiguity behaviour but presumably when the #470 is applied that will apply here too so that the behaviour in patterns and in substitutions is the same.I will merge this and then document it.

NWilson:
> I remembered that octal in substitution strings is broken in the current release, so nobody out there can actually be using it. I note that currently it has the Perl ambiguity behaviour but presumably when the https://github.com/PCRE2Project/pcre2/pull/470 is applied that will apply here too so that the behaviour in patterns and in substitutions is the same.

Agreed on both. There is no regression. And, I intend for the other PR to apply the Perl/Python ambiguity option to both patterns and replacements.

PR #476, opened 2024-09-17
pcre2_compile: make erroroffsets returned from read_number() more accurate

carenas:
A "fixup" to the bug addressed by #472, and that corrects a small regression introduced there
where group numbers larger than 65535 will report and even less accurate erroroffset.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -1351,7 +1351,7 @@ Arguments:
   allow_sign  if < 0, sign not allowed; if >= 0, sign is relative to this
   max_value   the largest number allowed;
               you must not pass a value for max_value larger than
-              INT_MAX/10 - 1 because read_number() relies on max_value to
+              INT_MAX/10 - 1 because this function relies on max_value to
               avoid integer overflow
   max_error   the error to give for an over-large number
   intptr      where to put the result
@@ -1370,6 +1370,8 @@ int sign = 0;
 uint32_t n = 0;
 PCRE2_SPTR ptr = *ptrptr;
 BOOL yield = FALSE;
+
+PCRE2_ASSERT(max_value <= INT_MAX/10 - 1);
 
 *errorcodeptr = 0;
 
@@ -1395,6 +1397,7 @@ while (ptr < ptrend && IS_DIGIT(*ptr))
   if (n > max_value)
     {
     *errorcodeptr = max_error;
+    while (ptr < ptrend && IS_DIGIT(*ptr)) ptr++;
     goto EXIT;
     }
   }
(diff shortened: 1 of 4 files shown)


src/pcre2_compile.c
```diff
@@ -1906,15 +1907,15 @@ else
...
+      if (read_number(&ptr, ptrend, -1, INT_MAX/10 - 1, ERR61, &s, errorcodeptr) &&
           (s < 10 || oldptr[-1] >= CHAR_8 || s <= (int)cb->bracount))
         {
         if (s > (int)MAX_GROUP_NUMBER) *errorcodeptr = ERR61;
           else escape = -s;     /* Indicates a back reference */
         break;
         }
 
-      ptr = oldptr;      /* Put the pointer back and fall through */
+      if (c <= CHAR_7) ptr = oldptr; /* Put the pointer back for fall through */
```
NWilson, line 1918: Okay... this seems convoluted to me. The check for `oldptr[-1] >= CHAR_8` was previously done in one single place. Now you've effectively duplicated that check in two locations (and inverted the logic here).

And, you've changed it so that `errorcodeptr` state is now polluted, but is later conditionally un-polluted, with a third check on `if (c >= CHAR_8) break`.

Personally, that seems less clear.

But, I have tested the code with all the possible scenarios:
* Single digit, starting with 1, or with 8
* Two-digit, starting with 1 or with 8
* < bracount
* > bracount
* read_number() overflow, starting with 1, or with 8

Everything... seems to work, I think? So this PR is equally correct as mine, it just adds some extra conditional logic.

src/pcre2_compile.c
```diff
@@ -1392,6 +1392,7 @@ while (ptr < ptrend && IS_DIGIT(*ptr))
   if (n > max_value)
     {
     *errorcodeptr = max_error;
+    while (ptr < ptrend && IS_DIGIT(*ptr)) ptr++;
```
NWilson, line 1395: I approve. I think we should do this change, but perhaps in a separate PR.

NWilson: Here, we're advancing `ptr`. We could equally well rewind the `ptr` back to the beginning of the digits, which might actually be better.

But really, it's an almost irrelevant nitpick.

carenas: Perl does report the bad offset at the end of the number (hence the preference to do the same).

Agree that in this specific case, where the error is recoverable and we are planning to rewind and read the number again, it might make sense to do the opposite though, but it might be a surprise to future users that expect `ptr` to point past the error from other similar functions.

src/pcre2_compile.c
carenas: Was trying to minimize the change, but hopefully the new version (which does a little more refactoring to clean the currently messy code) is clearer.

Historically, resetting of `ptr` was meant to be only done in preparation for the fallback which is why I changed the comment, but could had used a little more verbosity to the change.

src/pcre2_compile.c
NWilson: Great! Let's advance the pointer then, for consistency.

NWilson (approved):
A nice set of changes! Cool.

zherczeg (approved):
LGTM

src/pcre2_compile.c
```diff
@@ -1940,19 +1943,19 @@ else
...
-      case we treat the group number as too-large (since it may be larger than
-      INT_MAX we cannot return it for the caller to check).
+      read_number() is a number that is too large to be a group number. Because
+      that number might be still valid if read as an octal, errorcodeptr is not
+      set on failure, and therefore we rely on its value being slightly
+      bigger than MAX_GROUP_NUMBER to set it, if not falling through. */
 
-      \1 to \9 are always back references. \8x and \9x are too; \1x to \7x
-      are octal escapes if there are not that many previous captures. */
+      read_number(&ptr, ptrend, -1, MAX_GROUP_NUMBER, 0, &s, errorcodeptr);
```
NWilson, line 1951: You changed the code since I reviewed. You're now ignoring the return value from `read_number`, and baking in the assumption that when `read_number` returns false, `s` will be populated with the partially-parsed value.

That seems nasty. You should keep the check on its return value.

NWilson (changes requested):

src/pcre2_compile.c
carenas: > `s` will be populated with the partially-parsed value.

the partially parsed value, but one that would be slightly larger than  MAX_GROUP_NUMBER which will be returned by the function in those cases.

> That seems nasty. You should keep the check on its return value

OK, but then I would also document better how the IMHO equally bogus value we are going to use is relevant to the code below.

zherczeg (approved):
LGTM

src/pcre2_compile.c
NWilson: I'd prefer to think of it as a "sentinel value" than a "bogus value". 😉 

But thank you for updating the comments.

I'm happy knowing that the code can't be broken accidentally by refactoring, especially with the assertion you added.

PR #477, opened 2024-09-17
Improve caseless character range processing when utf is enabled

zherczeg:
Currently when utf caseless matching is requested, each character in a class range is checked one-by-one to find their other cases. I always thought this is inefficient, even if it is only done by the parser.

To speed things up, I have added a data structure, which contains ranges where characters have no other cases. The ranges have a minimum size. The size in uint32_t units with different minimum sizes (the number of ranges is half of that size):

```
minimum size 4: 124; /* 1111114 characters */
minimum size 8: 74; /* 1110983 characters */
minimum size 16: 60; /* 1110911 characters */
```

I choose 8 at the end, because only 74 values are needed, and these ranges cover 1110983 characters, which is 99.7% of all utf characters.

Performance improvement by the patch:

Original (-O3):
```
./pcre2test -t 100
  re> /[\x{100}-\x{100000}]/iB,utf
Compile time 8957.8600 microseconds
        Bra
        [KSks\xb5\xc5\xdf\xe5\xff\x{100}-\x{100000}]
        Ket
        End
```

New (-O3)
```
./pcre2test -t 100
  re> /[\x{100}-\x{100000}]/iB,utf
Compile time  83.0600 microseconds
        Bra
        [KSks\xb5\xc5\xdf\xe5\xff\x{100}-\x{100000}]
        Ket
        End
```

The new one is 107 times faster than the old, I think this is a nice speedup.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -5340,6 +5340,44 @@ for (++c; c <= d; c++)
 *cptr = c;             /* Rest of input range */
 return 0;
 }
+
+
+
+/*************************************************
+*             Get nocase ranges                  *
+*************************************************/
+
+/* This function returns the next nocase range after a character
+using binary search. The character might be included in the range.
+
+Arguments:
+  c           current character
+
+Yield:        range (start/end pair)
+*/
+
+static const uint32_t*
+get_nocase_range(uint32_t c)
+{
+uint32_t left = 0;
+uint32_t right = PRIV(ucd_nocase_ranges_size);
+uint32_t middle;
+
+if (c > MAX_UTF_CODE_POINT) return PRIV(ucd_nocase_ranges) + right;
+
+while (TRUE)
+  {
+  /* Range end of the middle element. */
+  middle = ((left + right) >> 1) | 0x1;
+
+  if (PRIV(ucd_nocase_ranges)[middle] <= c)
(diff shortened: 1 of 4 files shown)


zherczeg:
I have some ideas to further improve this patch.

src/pcre2_compile.c
```diff
@@ -5388,14 +5426,45 @@ if ((options & PCRE2_CASELESS) != 0)
...
+      {
+      c = skip_range[1];
+      skip_range += 2;
+      skip_start = skip_range[0];
+      }
 
     while ((rc = get_othercase_range(&c, end, &oc, &od,
              (xoptions & PCRE2_EXTRA_CASELESS_RESTRICT) != 0)) >= 0)
       {
+      if (c > skip_start)
```
zherczeg, line 5446: I thought get_othercase_range returns the `c` after the character with othercase, but sometimes it moves even further. I am not fully understand this.

zherczeg:
I have updated the code. @PhilipHazel what do you think?

PhilipHazel:
Very nice idea! I am happy with it. Is it ready for merging now?

zherczeg:
It is ready. The code is fail safe, although it would be great if I understand the purpose of the extra increase.

PR #478, opened 2024-09-17
Add fix and tests for \x with no hex digits

NWilson:
Why mess around with something that's broken, and which no users are likely to want?

Grasp the nettle, and simply disallow `\x` with no following hex digits (following non-hex character, or at end of pattern).

As documented in the code comments added, this diverges from Perl's default behaviour, but aligns with Perl+warnings, so is consistent with best-practices Perl usage.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -2093,10 +2093,29 @@ else
 
       else
         {
-        c = 0;
-        if (ptr >= ptrend || (cc = XDIGIT(*ptr)) == 0xff) break;  /* Not a hex digit */
+        /* Perl has the surprising/broken behaviour that \x without following
+        hex digits is treated as an escape for NUL. Their source code laments
+        this but keeps it for backwards compatibility. A warning is printed
+        when "use warnings" is enabled, and it's forbidden when "use strict"
+        is enabled. Because we don't have warnings, we simply forbid it. */
+        if (ptr >= ptrend || (cc = XDIGIT(*ptr)) == 0xff)
+          {
+          /* Not a hex digit */
+          *errorcodeptr = ERR78;
+          break;
+          }
         ptr++;
         c = cc;
+
+        /* With "use re 'strict'" Perl actually requires exactly two digits (error
+        for both \xA and \xAAA). This seems overly strict, and there seems
+        little incentive to align with that, given the backwards-compatibility
+        cost.
+
+        For comparison, note that other engines disagree. For example:
+          - Java allows 1 or 2 hex digits. Error if 0 digits. No error if >2 digits
+          - .NET requires 2 hex digits. Error if 0, 1 digits. No error if >2 digits.
+        */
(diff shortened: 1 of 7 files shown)


PhilipHazel:
I hadn't known about Perl's warning. Good point. I have merged and will edit the docs.

NWilson:
I found this helpful site:

https://perlbanjo.com/bca01b0a27

The `use re 'strict'` error was introduced in Perl 5.22.4 (9 years ago).

https://perlbanjo.com/394c21e056

The `use warnings` logging (for this `\x` issue) has a warning going back as far as 5.8.9 (and probably older). The warning message was made more detailed in 5.32.1 (3 years ago).

---

Finally, the trailing `\x` at end of pattern generates no warning with `use warnings` on the latest Perl, but with `use re 'strict'` it is forbidden since 5.22.4 (9 years ago).

src/pcre2_compile.c
```diff
@@ -2093,10 +2093,29 @@ else
 
       else
         {
-        c = 0;
-        if (ptr >= ptrend || (cc = XDIGIT(*ptr)) == 0xff) break;  /* Not a hex digit */
+        /* Perl has the surprising/broken behaviour that \x without following
+        hex digits is treated as an escape for NUL. Their source code laments
+        this but keeps it for backwards compatibility. A warning is printed
+        when "use warnings" is enabled, and it's forbidden when "use strict"
+        is enabled. Because we don't have warnings, we simply forbid it. */
```
carenas, line 2100: this comment needs updating so it correctly states `use re "strict"`

carenas:
> use re 'strict'

is still experimental though even in the unreleased version of Perl

the comments referred by mistake to "use strict" though

NWilson:
The `use warnings` logging has been around since forever. PCRE2 has precedent of following the dialect of "warning-free Perl" ie we follow what they document as being correct Perl, rather than more permissively follow what they implement. I personally don't feel guilty about fixing this.

carenas:
I wasn't arguing that we shouldn't fix it, just got tripped by the comment typo and the fact that my test was also unlucky to have `\x` at the end of the pattern.

Thanks for the helpful site tip BTW;  '\x' at the end of the pattern not triggering a warning is not only happening in the last version and it is likely because Perl is triggering the warning when it find a character that can't identify as hexadecimal after `\x` and reading up to additional characters.

src/pcre2_compile.c
carenas: oh dear, didn't realize that the "strict" confusion was also reflected in the documentation.

To clarify, Perl also shows warnings when using less than hexadecimal digits (ex:`/\xa/`), and the warning we are referring to also apply to `/\x/`.

That is documented below, but in the case of the single digit hexadecimals we are ignoring those warnings for backward compatibility.

Again, I am not arguing there should be any changes on behaviour to this code, just pointing out that there is some slight confusion on the rationale for doing them. Let me know if I should instead provide a PR, which might be also a less confusing way to make that argument, it would seem.

Issue #163, opened 2022-11-23
Support of old C89 standard


mikir:
Not sure if support of old C89 standard has been cancelled but if this is not that case, the last release has introduced some small violations in modules `pcre2_compile.c` and `pcre2_script_run.c`. Examples:

```
pcre2_compile.c: In function 'compile_branch':
pcre2_compile.c:5857:13: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
             for (int i = 0; i < 32; i++) pbits[i] |= cbits[(int)i + taboffset];
             ^
pcre2_compile.c:5857:13: note: use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code
pcre2_compile.c:5859:22: error: redefinition of 'i'
             for (int i = 0; i < 32; i++) pbits[i] &= ~cbits[(int)i + taboffset];
                      ^
pcre2_compile.c:5857:22: note: previous definition of 'i' was here
             for (int i = 0; i < 32; i++) pbits[i] |= cbits[(int)i + taboffset];
                      ^
pcre2_compile.c:5859:13: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
             for (int i = 0; i < 32; i++) pbits[i] &= ~cbits[(int)i + taboffset];
             ^

pcre2_script_run.c: In function '_pcre2_script_run_8':
pcre2_script_run.c:109:1: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
 for (int i = 0; i < FULL_MAPSIZE; i++) require_map[i] = 0;
 ^
pcre2_script_run.c:109:1: note: use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code
pcre2_script_run.c:248:7: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
       for (int i = 0; i < FULL_MAPSIZE; i++)
       ^
pcre2_script_run.c:285:9: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
         for (int i = 0; i < FULL_MAPSIZE; i++) require_map[i] &= map[i];
```

Please consider to change it. Or feel free to close it without any actions if C89 violations are not an issue any more.

PhilipHazel:
Declarations in "for" statements are in C99, so have been standardized for over 20 years. I am happy to move to C99, unless there are very strong objections.

mikir:
OK, I see. No objections. :-)

PR #480, opened 2024-09-18
ci: add very basic 64bit Windows DEV job

carenas:
This job should at least ensure the library builds cleanly (although without JIT) and creates a DLL.

Its not as verbose as the original MSVC job (which has been renamed so it is clear it is 32bit) so that any warnings are clearly visible if building `pcre2grep`, `pcre2test` or JIT is enabled.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -8794,10 +8794,10 @@ for (;;)
     {
     if (lengthptr == NULL)
       {
-      PCRE2_SIZE branch_length = code - last_branch;
+      uint32_t branch_length = (uint32_t)(code - last_branch);
       do
         {
-        PCRE2_SIZE prev_length = GET(last_branch, 1);
+        uint32_t prev_length = GET(last_branch, 1);
         PUT(last_branch, 1, branch_length);
         branch_length = prev_length;
         last_branch -= branch_length;
@@ -8808,7 +8808,7 @@ for (;;)
     /* Fill in the ket */
 
     *code = OP_KET;
-    PUT(code, 1, (int)(code - start_bracket));
+    PUT(code, 1, (uint32_t)(code - start_bracket));
     code += 1 + LINK_SIZE;
 
     /* Set values to pass back */
@@ -11006,7 +11006,7 @@ if (errorcode == 0 && cb.had_recurse)
         }
       }
 
-    PUT(rcode, 1, rgroup - codestart);
+    PUT(rcode, 1, (uint32_t)(rgroup - codestart));
     }
   }
 
--- a/src/pcre2test.c
+++ b/src/pcre2test.c
@@ -5071,15 +5071,14 @@ switch(cmd)
     {
     while (isspace(*argptr)) argptr++;
     if (*argptr == 0) break;
-    for (i = 1; i < sizeof(newlines)/sizeof(char *); i++)
+    for (uint16_t j = 1; j < sizeof(newlines)/sizeof(char *); j++)
       {
-      size_t nlen = strlen(newlines[i]);
-      if (strncmpic(argptr, (const uint8_t *)newlines[i], nlen) == 0 &&
+      size_t nlen = strlen(newlines[j]);
+      if (strncmpic(argptr, (const uint8_t *)newlines[j], nlen) == 0 &&
           isspace(argptr[nlen]))
         {
-        if (i == NEWLINE_DEFAULT) return PR_OK;  /* Default is valid */
-        PCRE2_ASSERT(i <= UINT16_MAX);
-        if (first_listed_newline == 0) first_listed_newline = (uint16_t)i;
+        if (j == NEWLINE_DEFAULT) return PR_OK;  /* Default is valid */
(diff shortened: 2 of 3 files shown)


src/pcre2_compile.c
```diff
@@ -8794,10 +8794,10 @@ for (;;)
     {
     if (lengthptr == NULL)
       {
-      PCRE2_SIZE branch_length = code - last_branch;
+      uint32_t branch_length = (uint32_t)(code - last_branch);
```
zherczeg, line 8797: Don't change the type. Just add the casting.

src/pcre2test.c
```diff
@@ -5071,15 +5071,14 @@ switch(cmd)
     {
     while (isspace(*argptr)) argptr++;
     if (*argptr == 0) break;
-    for (i = 1; i < sizeof(newlines)/sizeof(char *); i++)
+    for (uint16_t j = 1; j < sizeof(newlines)/sizeof(char *); j++)
```
zherczeg, line 5074: Do we use such declaration? Btw is this a bug?

zherczeg (changes requested):

src/pcre2test.c
carenas: C99 style declarations in loops are in the codebase for almost two years (see #163), not sure what you refer as a bug though.

src/pcre2_compile.c
carenas: there is no point having a PCRE2_SIZE variable here if the data we are going to handle is always going to be less than UINT32_MAX.

Additionally, we have to `PUT` this value below and that would also require a casting in that case.

src/pcre2test.c
zherczeg: The variable 'i' is declared as it seems. If its value needs to be preserved, then this code is wrong.

src/pcre2_compile.c
zherczeg: `PUT` needs uint32_t? So far no compiler is complained about this before.

src/pcre2test.c
carenas: The variable `i` is declared as a `size_t` at the beginning of `process_command()` and used in multiple places (mainly as a loop variable) but in none of those places it was meant to be preserved AFAIK.

src/pcre2_compile.c
carenas: An example [run](https://github.com/carenas/pcre2/actions/runs/10916668606/job/30298529257) with the proposed change and without the cast in PUT.

zherczeg: https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_intmodedep.h#L105

The macro has an explicit type cast. Even with `uint32_t`, it is often shortened to 8/16 bit. Why the compiler complains only here, and not in other similar cases?

src/pcre2test.c
zherczeg: It looks like an unsigned variable is perfect here. `sizeof(newlines)/sizeof(char *)` is already an unsigned value. The patch also removes a `PCRE2_ASSERT`

src/pcre2_compile.c
carenas: My guess (as explained before with clang) is that that specific truncation is common enough with the migration to 64bit CPUs to warrant a warning.

zherczeg: But why only here? With explicit type casts? I suspect something is wrong with the macro definition, and should be fixed elsewhere. That is why I pushing the topic.

zherczeg: Maybe clang could be used to get a more detailed info about the implicit type cast which we don't see. If it shows as detailed info as `gcc` these days.

PhilipHazel:
Is this ready for merging? I'm not sure whether the discussion is still active or not.

src/pcre2_compile.c
carenas: Again, I think that the difference comes from the fact that truncating a 64bit integer into 32bits (which is what PCRE2_UCHAR is for the 32bit library) triggers these warnings (by default in MSVC and with extra help in clang).

That doesn't mean that truncating it to 16 o 8bit isn't as problematic, but the compiler writers probably wanted to focus on the ones that will be more likely in current codebases.

FWIW, there are several PUT invocations already with casts, which might had been added because of similar warnings, but at least in this case, the best will be to avoid the cast and use a type of the right size, so it will be IMHO better to resolve this specific thread.

zherczeg:
Ok, lets merge it.

src/pcre2test.c
carenas: > The patch also removes a `PCRE2_ASSERT`

yes, as it is mentioned in the commit message, since it is not necessary and was introduced by mistake (instead of an static assert that would break at build time, and that was suggested instead) in #466.

note that any truncation should be also visible in the clang DEV job so implementing an static assert macro (which I plan to do anyway) and use it here is not that urgent.

PR #482, opened 2024-09-18
Fix pcre2_substitute() to allow all named captures to be referenced

NWilson:
* Capture group names can be up to 128 characters, but `pcre2_substitute()` was only allowing 32-character names
* With `/utf`, capture group names can contain non-ASCII characters, but `pcre2_substitute()` was allowing ASCII names only

The new function `read_name()` takes a silly number of lines of code given what it does, but I wanted it to match visually as closely as possible with the corresponding function in `pcre2_compile.c`.

Note that we don't allow SPACE and HT inside `${name}` replacements, even though Perl does. I have kept this unchanged. I have no opinion or desire to add more whitespace skipping.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -2574,6 +2574,9 @@ if (ptr >= ptrend)                 /* No characters in name */
 *nameptr = ptr;
 *offsetptr = (PCRE2_SIZE)(ptr - cb->start_pattern);
 
+/* If this logic were ever to change, the matching function in pcre2_substitute.c
+ought to be updated to match. */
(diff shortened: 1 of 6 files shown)


PhilipHazel:
Thanks for picking this up. The increase of group names from 32 to 128 happened for release 10.44, and obviously I forgot to do it in the substitute function.

PhilipHazel:
I've made one minor change, to get rid of a "mixed code and declaration" warning. The style of PCRE2 code is not to mix code and declarations, except for a declaration inside an "if".

PR #483, opened 2024-09-18
Add implementation of \g<...> and $<...> in substitute

NWilson:
These are the standard syntaxes in Python and JavaScript for capture group references.

Python's `\g<name_or_num>` permits either a capture group name or number. We don't allow whitespace inside the `<>` nor do we allow `\g<*MARK>` syntax.

JavaScript's `$<name>` syntax only allows a capture group name; otherwise it's the same as `\g<name>`.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -1703,7 +1703,7 @@ else
   if (cb == NULL)
     {
     if (c < CHAR_0 || 
-       (c > CHAR_9 && (c != CHAR_c && c != CHAR_o && c != CHAR_x)))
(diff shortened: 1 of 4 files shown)


NWilson:
PR #482 (fix MAX_NAME_LENGTH & unicode limit) would need to be merged first to avoid conflicts.

PhilipHazel:
This PR is now giving a build warning (sign differences) and also says there is a conflict.

NWilson:
Thanks Philip, it's OK - I was planning to rebase this PR. It should be ready now, assuming I haven't made any mistakes.

PhilipHazel:
There's another mixed declaration and code, which I will fix. I will also document this, but not today. I think there is now sufficient syntax in substitute strings that it will be worth adding a section to the pcre2syntax page.

Issue #484, opened 2024-09-19
Make PCRE2_ASSERT work


zherczeg:
Even with `--enable-debug` this code does not work:

https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_util.h#L56

As far as I know, `__builtin_unreachable` has no effect, only informs the compiler. A failed assertion should stop execution. Is there a flag needed to be passed to the compiler?

Debugging should not involve complicated methods, the purpose is catching errors not high performance.

zherczeg:
@carenas can you help me?

carenas:
I couldn't see a problem with my compilers, but they are all fairly recent, which one did you have problems with?

zherczeg:
gcc 11.4.0. What happens for unreachable in your system when an assert fails? The description of unreachable says what happens is random, since no code is generated, and the next code is simply executed. As for asserts, the compiler considers it as a dead code, since the assert has no side effect.

zherczeg:
https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html

"If control flow reaches the point of the __builtin_unreachable, the program is undefined." Nothing happens is a perfect effect.

carenas:
> gcc 11.4.0. What happens for unreachable in your system when an assert fails?

the compiler introduces an `ud2` instruction that kills the program:

```
illegal hardware instruction  ./pcre2test
```

zherczeg:
Do you need a compiler option for that? Maybe it is better to move back to more reliable methods, especially if undefined becomes defined in the future.

carenas:
No additional compiler option needed, agree it will be better not to rely on "undefined behaviour" for this, but was waiting on @alexdowad input since this was originally his code.

Lets give him some time, specially considering you have a "workaround" as well AFAIK,

alexdowad:
Ah, for MS Visual C++, there is also `__assume`, which might be more appropriate than `__debugbreak`.

zherczeg:
This is debug only, so the most reliable is the best for me. I also prefer if some info is printed. What shall we do?

carenas:
> What shall we do?

Just using assert() by default might be better IMHO, will prepare a PR

Note that assert() is part of ANSI C, so it shouldn't be much of a concern, portability wise, to do so.

MatthewVernon:
[...]

PR #485, opened 2024-09-20
Do not use __builtin_unreachable() in PCRE2_ASSERT()

carenas:
Simplify PCRE2_ASSERT() and while at it remove no longer needed dependency on __builtin_expect()

Fixes: #484

### DIFF
--- a/src/pcre2_util.h
+++ b/src/pcre2_util.h
@@ -52,9 +52,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 #ifdef PCRE2_DEBUG
-#if defined(HAVE_BUILTIN_EXPECT) && defined(HAVE_BUILTIN_UNREACHABLE)
-#define PCRE2_ASSERT(x) do { if (__builtin_expect(!(x), 0)) __builtin_unreachable(); } while (0)
-#elif defined(HAVE_ASSERT_H)
+#if defined(HAVE_ASSERT_H) && !defined(NDEBUG)
 #include <assert.h>
 #define PCRE2_ASSERT(x) assert(x)
 #elif defined(HAVE_STDLIB_H) && defined(HAVE_STDIO_H)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -192,11 +192,6 @@ CHECK_C_SOURCE_COMPILES(
 )
 
 CHECK_C_SOURCE_COMPILES(
-  "int main(void) { if (__builtin_expect(1+1, 2)) return 1; return 0; }"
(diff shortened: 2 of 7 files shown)


zherczeg (approved):
LGTM

src/pcre2_util.h
```diff
@@ -52,9 +52,7 @@ POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 #ifdef PCRE2_DEBUG
-#if defined(HAVE_BUILTIN_EXPECT) && defined(HAVE_BUILTIN_UNREACHABLE)
-#define PCRE2_ASSERT(x) do { if (__builtin_expect(!(x), 0)) __builtin_unreachable(); } while (0)
-#elif defined(HAVE_ASSERT_H)
+#if defined(HAVE_ASSERT_H) && !defined(NDEBUG)
```
thesamesam, line 55: Would it make sense for the non-`PCRE2_DEBUG` case to do use the old (removed in this commit) version of `PCRE2_ASSERT` there?

carenas: I don't think so, what is that you would like to accomplish by doing so?

thesamesam: I'm not really bothered about it, I just wanted to note that it would let you keep the optimisation benefits in non-debug builds and also get a trap from UBSAN when asserts are disabled. If it's not desired, no problem.

carenas: I see; `PCRE2_ASSERT()` doesn't generate any code right now in non debug builds, so it is already optimized for speed AFAIK.

zherczeg: I prefer no code generated by asserts in release builds. There could be release asserts if it is needed, but I have no plans for that.

thesamesam: Yeah, it just means you don't get the optimisation benefit of saying "this shouldn't be reachable".

carenas: > it just means you don't get the optimisation benefit of saying "this shouldn't be reachable"

for that there is a different macro  `PCRE2_UNREACHABLE()`, but their use is also likely to change (see #490), `PCRE2_ASSERT()` was never meant to be used to signal that, and the fact that it accidentally did because of the implementation is what is being corrected here.

zherczeg: I suspect he mentions a different thing. When `(!(a < 5))` goes to unreachable, the compiler can tell that `a < 5`, and it can use it for optimization. However, I have doubts about the usefulness of such knowledge, and the condition part of the assertion may also have side effects (forcing extra memory accesses), if the compiler cannot fully eliminate them.

PR #486, opened 2024-09-21
improve usability of pcre2_set_optimize() documentation

carenas:
Align it better to the style used in other similar pages (ex: pcre2_set_compile_extra_options.3) by including
enough information so it can be used independently of other pages.


zherczeg (approved):
LGTM

PR #487, opened 2024-09-21
pcre2_compile: avoid 1 byte buffer overread parsing VERBs

carenas:
Complete the bugfix merged with #463 and update documentation (including ChangeLog)

The documentation itself might need further tweaks to better clarify when parsing of the
affected `pso` VERBs is meant to happen though.


zherczeg (approved):
LGTM

PR #490, opened 2024-09-22
reimplement PCRE2_UNREACHABLE() assertions with a slightly safer approach

carenas:
Reimplement the original assertions to use different versions for debug and non debug builds.

Non debug asserts use compiler specific builtins that are not safe to invoke, and therefore their use has been audited to paths that should NEVER be executed.

Co-Authored-by: Alex Dowad <alexinbeijing@gmail.com>


alexdowad:
One comment: the `PCRE2_UNREACHABLE()` macro was intended to serve 2 different purposes:

1. In debug builds, it's a debugging tool which can help fuzzers, etc. to find bugs where control flow goes somewhere it shouldn't;
2. In release builds, it's an optimization hint for the compiler.

I think the implementation in this PR only meets point 1 but not point 2; but if that is not correct, please mention.

carenas:
You are correct, there is no production implementation and therefore there is no optimization hint either.

Interestingly enough from all the places where it was used, the only place where it would make a difference was THIS function, and the difference wouldn't be good for our users.

If the assertion was ever reached, and their library was built with a recent enough compiler in an x86 system, they will get a crash, instead of getting an error, because as you clearly pointed out, the "return" call itself would be removed at compile time; was that the what was expected from that "optimization"?

alexdowad:
> If the assertion was ever reached, and their library was built with a recent enough compiler in an x86 system, they will get a crash, instead of getting an error, because as you clearly pointed out, the "return" call itself would be removed at compile time; was that behaviour what was expected from that "optimization"?

In answer to that question, what was expected was that if we assert a certain path is unreachable, that path should really be unreachable, and because the compiler is told that is unreachable, it can (in some cases) generate better code. For example, if you tell the compiler that the `default:` label in a `switch` statement is unreachable, it can assume that there will always be a matching label and doesn't need to generate code to handle the "default" case.

If the unreachable path is really unreachable, then there will be no crashes in production (because it will never be reached). On the other hand, if we are not able to guarantee, through whatever method, that our "unreachable" code is really unreachable... that is a problem. 😦 Since PCRE2 has significant installed base, it would be nice if the codebase was at a high enough level of quality that we could have full confidence that the "unreachable" paths are really unreachable. In that case, they could safely benefit from `__builtin_unreachable`'s optimization hint.

However, if the codebase is not at that level of quality yet, then you are very right that `PCRE2_UNREACHABLE` must not do anything unsafe in production builds.

In any case, thanks very much for looking into this issue.

Another thought: It would be nice if, in production builds, *all* instances of `PCRE2_ASSERT` and `PCRE2_UNREACHABLE` would make public API functions return `PCRE2_INTERNAL_ERROR` if an assertion fails. That could even mitigate possible vulnerabilities by turning them into error returns instead. However, it would probably require some trickery with `longjmp`, which seems inadvisable.

alexdowad:
Just one suggestion: I think the assertion failure message in this PR would be easier to understand if it said something like "Execution must not reach this point" or "Execution reached unexpected point".

zherczeg:
Reaching an unreachable code is always a big problem, since you cannot write tests to see what happens. Hence whatever we do, we probably have issues. I don't mind unreachable in matching code, because its purpose is high performance. For pattern compilation, we could use errors, although the problem is the same: we cannot test them.

carenas:
I would prefer no unreachables in production, as they are additional code, their behaviour changes from compiler to compiler, and even between versions of the same compiler and have bugs, but agree that since they are meant to be really unreachable (assuming the original selection was done carefully enough) they could be used to maybe improve performance.

Anyway, as requested by Alex updated the series for further discussion.

zherczeg:
Btw I would prefer an assertion before all internal errors. When I develop code, and get those errors, it is too hard to find their sources.

.github/workflows/build.yml
```diff
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   linux:
     name: Linux
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
zherczeg, line 8: What is the problem with latest?

src/pcre2_util.h
```diff
@@ -44,26 +44,52 @@ POSSIBILITY OF SUCH DAMAGE.
...
-#define PCRE2_UNREACHABLE() do {} while(0)
+#warning "missing implementation of PCRE2_UNREACHABLE()"
 #endif
 
 #ifdef PCRE2_DEBUG
+
 #if defined(HAVE_ASSERT_H) && !defined(NDEBUG)
 #include <assert.h>
 #define PCRE2_ASSERT(x) assert(x)
 #elif defined(HAVE_STDLIB_H) && defined(HAVE_STDIO_H)
```
carenas, line 59: this additional checks were apparently introduced by mistake, which is why the final version might also remove them and why they are not being used in the new debug assert for PCRE2_UNREACHABLE() below.

the core was just reformatted so it can be easier to read and also compare with the proposed similar implementation.

.github/workflows/build.yml
carenas: latest is currently 22.04 and has a buggy version of gcc, BTW GitHub is in the process to move the "latest" label to 24.04 so this could happen eitherway.

alexdowad:
> agree that since they are meant to be really unreachable (assuming the original selection was done carefully enough) they could be used to maybe improve performance.

Carlo, how would this be:

- We agree that an UNREACHABLE assertion should only be used when 1) the code has been carefully audited to make sure that the "unreachable" path is really unreachable, and 2) the code in question is heavily tested by the existing test suite.
- UNREACHABLE assertions compile to an optimization hint in release builds, a failing assertion (i.e. abort) in debug builds.
- For now, all calls to PCRE2_UNREACHABLE can be converted to a comment saying "/* We should not get here */" and PCRE2_ERROR_INTERNAL return. I will go through one by one and check each site carefully before converting it back to an assertion.

Just a suggestion.

alexdowad:
> Btw I would prefer an assertion before all internal errors. When I develop code, and get those errors, it is too hard to find their sources.

Suggestion: Aside from PCRE2_ASSERT, can we add another macro which expands to a failing assertion (abort) in debug builds, but `return PCRE2_ERROR_INTERNAL;` in release builds?

This macro must only be used in functions which return `int` status code (zero for success, non-zero for error code).

src/pcre2_util.h
```diff
@@ -44,26 +44,52 @@ POSSIBILITY OF SUCH DAMAGE.
...
+fprintf(stderr, "Execution reached unexpected point " __FILE__      \
+                ":%d\n", __LINE__);                                 \
+abort();                                                            \
+} while(0)
+
+#endif /* PCRE2_DEBUG */
+
+#ifndef PCRE2_UNREACHABLE
+#ifdef PCRE2_DEADTRAP
+#define PCRE2_UNREACHABLE(d) if (d == 0) PCRE2_DEADTRAP()
```
alexdowad, line 83: Ah, I see. So the parameter to PCRE2_UNREACHABLE indicates whether we have confidence that the code is "really unreachable" or not.

Interesting.

carenas: Guess that is a way to put it, but what it really indicates to me is, "this unreachable should NEVER abort in non debug builds and should NEVER be considered as a hint for code elimination either, it is there ONLY to be used in debug builds to catch bugs, as all assertions should be.

carenas:
> can we add another macro

sure, fork away and go ahead, I promise not to rebase this until all discussion is settled and review your code to suggest improvements.

note though that having different assertion per type of function is going to be messy, since each function has different ways to report those errors, so IMHO might be easier to open code each as needed using `PCRE2_ASSERT()`

src/pcre2_util.h
```diff
@@ -44,26 +44,52 @@ POSSIBILITY OF SUCH DAMAGE.
...
 #ifdef HAVE_BUILTIN_UNREACHABLE
-#define PCRE2_UNREACHABLE() __builtin_unreachable()
-#elif HAVE_BUILTIN_ASSUME
-#define PCRE2_UNREACHABLE() __assume(0)
+#define PCRE2_DEADTRAP() __builtin_unreachable()
+#elif defined(HAVE_BUILTIN_ASSUME)
+#define PCRE2_DEADTRAP() __assume(0)
 #else
-#define PCRE2_UNREACHABLE() do {} while(0)
+#warning "missing implementation of PCRE2_UNREACHABLE()"
```
zherczeg, line 51: This looks like a strong requirement

src/pcre2_util.h
zherczeg: Instead of a parameter, I would introduce two unreachable variants. PCRE2_UNREACHABLE / PCRE2_SOFT_UNREACHABLE (or any better name).

src/pcre2_util.h
carenas: I know and I didn't like it either, but how else can we let people know that their compiler might silently continue running code after reaching something labeled as `PCRE2_UNREACHABLE()`.

This whole funny business with #489 where we could reasonable think it was ok to delete the fail safe code twice in a month, makes me think this is too confusing and needs at least a stronger "heads up".

src/pcre2_util.h
zherczeg: The names are good. If I understand correctly any unreachable is a bug, just we have a fallback in the latter case.

src/pcre2_util.h
zherczeg: well ok

carenas:
> * We agree that an UNREACHABLE assertion should only be used when 1) the code has been carefully audited to make sure that the "unreachable" path is really unreachable, and 2) the code in question is heavily tested by the existing test suite.

The part we seem to be having difficult understanding is that there are cases were some code that would be normally unreachable and might be at a specific point when audited, could be still affected by future changes and in those cases, removing apparent dead code that was there to be a safety net for those cases (either explicitly or even worst, implicitly through dead code elimination by a compiler) is not a good idea and need to be handled differently (at least in non debug builds)

> * UNREACHABLE assertions compile to an optimization hint in release builds, a failing assertion (i.e. abort) in debug builds.

this was implemented in the last revision, but with the caveat that we need the exception from the previous point in at least 1 case for the non debug builds.

> * For now, all calls to PCRE2_UNREACHABLE can be converted to a comment saying `/* We should not get here */` and PCRE2_ERROR_INTERNAL return. I will go through one by one and check each site carefully before converting it back to an assertion.

returning PCRE2_ERROR_INTERNAL might not be what is required on all places were this assert was introduced originally, it would also seem like a lot of work for something that could be done without the intermediate step and accomplished at the end of this series if you just did the promised audit for each entry over it.

as stated in the subject, my end objective is not to remove all those assertions but just make sure that it is clear how to use them going forward (even by changing their names so there is no longer confusion) and their use is safe from unexpected side effects in all the places were they are currently being used.

src/pcre2_study.c
```diff
@@ -759,7 +759,7 @@ for (;;)
     }
   }
 
-  PCRE2_UNREACHABLE(); /* Control never reaches here */
+PCRE2_UNREACHABLE(); /* Control never reaches here */
```
carenas, line 762: debated with myself about kicking these changes into a post cleanup patch, but I thought that since they were included since the first version and I am not clear yet on how the "audit" is going to be made or by whom it was safer to keep them for now to avoid anyone else having unnecessary conflicts or distracting noise.

src/pcre2_auto_possess.c
```diff
@@ -245,7 +245,7 @@ switch(ptype)
     default:
     return (PRIV(ucp_gentype)[prop->chartype] == ucp_Z) == negated;
     }
-  PCRE2_UNREACHABLE(); /* Control never reaches here */
+  PCRE2_UNREACHABLE(0); /* Control never reaches here */
```
carenas: this an the next are just leftover from v2, as the PCRE2_UNREACHABLE() doesn't have a parameter anymore and the one that is meant to be used in places were "fall back" code should be kept in non debug build use a different macro instead as suggested by Zoltan

src/pcre2_study.c
alexdowad: Yes, it's better to include them. Sorry that I was not used to the brace style for this project (still not used to it, but am getting there) when I included that extra indent.

alexdowad:
> IMHO might be easier to open code each as needed using `PCRE2_ASSERT()`

Sure, let's do that... it was just an idea.

alexdowad:
2 occurrences of `PCRE2_UNREACHABLE` in pcre2grep.c are safe. Same for pcre2test.c.

Recommend that `PCRE2_MAYBE_UNREACHABLE` should be used instead in pcre2_study.c; this one *should* be safe, but the function in question is long, complicated, and it's hard to see at a glance everything that is happening there. Likewise for pcre2_auto_possess.c.

Some occurrences in pcre2_compile.c are definitely safe. But as Zoltan mentioned in this thread, compilation is less performance-sensitive than matching... so maybe it's better to use `PCRE2_MAYBE_UNREACHABLE` anyways.

I do think that the name `PCRE2_MAYBE_UNREACHABLE` will confuse people who are reading the code for the first time. The natural question which will come to a reader's mind is: "This code *may be* unreachable? What does that mean?"

Other options include: `PCRE2_ASSERT_UNREACHABLE`, `PCRE2_UNREACHABLE_SAFE`, `PCRE2_DEBUG_UNREACHABLE`.

carenas:
> I do think that the name `PCRE2_MAYBE_UNREACHABLE` will confuse people who are reading the code for the first time

Fair enough, then will post a v4 that would use `PCRE2_DEBUG_UNREACHABLE` instead as it was the one you proposed and that I suggested as well before but discarded.

FWIW as per the "audit", I was expecting commits not comments, which might be better to hold until that update is pushed, so no unnecessary conflicts are created; presume you are ok with my "POC" audit in patch 4?, will hold othewise pushing an update if you would rather see changes so I can include them there.

alexdowad:
> FWIW as per the "audit", I was expecting commits not comments, which might be better to hold until that update is pushed, so no unnecessary conflicts are created; presume you are ok with my "POC" audit in patch 4?

Sure, I will provide some commits then. Yes, patch 4 looks good.

carenas:
> prefer one patch

I know, and I promise to squash them all as soon as we get a working audited list of assertions from Alex, lets give him some time, as that might be a little tricky, since as you pointed out it is "impossible" to debug assertions.

on that line, wonder what would the following code print on your system with gcc 10.4?

```
#include <stdio.h>

int main(void)
{
  int i = 1;
  int r;

  switch(i) {
    case 1:
      r = 2;
      __builtin_unreachable();
      break;
    case 2:
      r = 1;
      break;
  }

  printf("%d\n", r);
  return 0;
}
```

zherczeg:
prints `1`. According to the description anything is possible, and that is true.

zherczeg:
With:
```
    case 2:
      r = 3;
```
prints '3'. The `break` is considered dead code.

alexdowad:
Carlo and Zoltan, please see `auditasserts` branch in my fork.

https://github.com/alexdowad/pcre2/commit/0c290a4e681876a80401804a088402a8aa9fee71#r147115420

zherczeg:
This is an assert for me:
```
          if (s != INT_MAX)
            PCRE2_DEBUG_UNREACHABLE();
```

carenas:
> please see `auditasserts`

to be able to review the code, it will be easier if making a PR against this [branch](https://github.com/carenas/pcre2/tree/debugass) instead, of course once your patches are ready, I'll include them as part of this PR

alexdowad:
Pushed an update based on correction from Zoltan.

zherczeg:
The changes are ok. This looks like a nice improvement in the code.

alexdowad:
> to be able to review the code, it will be easier if making a PR against this [branch](https://github.com/carenas/pcre2/tree/debugass) instead, of course once your patches are ready, I'll include them as part of this PR

Thanks for that tip. I am just going out shortly but should be able to do this after I get back.

carenas:
> This looks like a nice improvement in the code

Agree, but I have to admit that I was hoping to see more of an "audit" of the current asserts instead, which I thought we all agree were risky to have (specially in non debug builds) without the 3 assumptions that were posted [before](https://github.com/PCRE2Project/pcre2/pull/490#issuecomment-2367257484) and that I summarize as:

1. Only add non debug asserts in code that has high test coverage and that has been audited to be completely safe and not trigger risky sideeffects (some of which were described in the commit message)
2. In places where internal errors are generated and there is dead code, add DEBUG assertions to make sure it stays dead and make sure there is test coverage around it.

The lack of added tests, and the fact we are looking again at another massive code change that seem to had been driven by search of comments "hinting" of dead code, and that seem to had been what got us in this conundrum to begin with, worries me.

alexdowad:
> Agree, but I have to admit that I was hoping to see more of an "audit" of the current asserts instead

That's what was done, and in my commit message, I explained the basis for using PCRE2_UNREACHABLE in selected places. Please let me know what else I can do to communicate the results of the 'audit' more clearly.

carenas:
> Please let me know what else I can do to communicate the results of the 'audit' more clearly.

You misunderstood my comment, I am not arguing you didn't do the audit (will take your word for it, although documenting the code at each point instead is what is preferred), but that you added a lot more assertions (which again, I am not arguing again, but just not in THIS PR).

Don't forget how we got here. A well intentioned change to make some comments all around the codebase about dead code verifiable to allow for better automated testing with fuzzers, but that had the sideeffect of removing failsafes and increasing the risk of crashes in non debug builds.

alexdowad:
> you added a lot more assertions (which again, I am not arguing against, and indeed I also think like Zoltan that make the code better, but just not in THIS PR).

Understood. I can break those out into a separate PR.

carenas:
> > you added a lot more assertions (which again, I am not arguing against, and indeed I also think like Zoltan that make the code better, but just not in THIS PR).
> 
> Understood. I can break those out into a separate PR.

On second though, and since  you already had us review them (even if comments might be eventually lost without a PR), I think that it will be better if I can pull it here instead so we can review everything in one single place after squashing and that way avoid losing more threads and be more efficient.

Note that the last posted update has known issues (and probably a mistaken update) but I will squash it AS-IS to avoid further delays.

src/pcre2_auto_possess.c
```diff
@@ -1121,7 +1122,7 @@ for(;;)
   if (list[1] == 0) return TRUE;
   }
 
-  PCRE2_UNREACHABLE(); /* Control never reaches here */
+PCRE2_DEBUG_UNREACHABLE(); /* Control should never reach here */
 }
```
carenas, line 1128: the `PCRE2_UNREACHABLE()` below was fixed in a previous version of this patchset but regressed, a fix is forthcoming as [discussed](https://github.com/alexdowad/pcre2/commit/bb5cc085fd5d448e0e7b467fb25c958308fa883c#r147116369)

src/pcre2_compile.c
```diff
@@ -1863,7 +1863,12 @@ else
         but we keep it just to be safe and because it will also catch the bogus
         value set on failure of that function. */
 
-        if ((unsigned)s > MAX_GROUP_NUMBER) *errorcodeptr = ERR61;
+        if ((unsigned)s > MAX_GROUP_NUMBER)
+          {
+          if (s != INT_MAX)
```
carenas, line 1868: this could be written instead as `PCRE2_ASSERT(s == INT_MAX)`

alexdowad: Yep, Zoltan already pointed that out and I pushed updated code accordingly.

alexdowad: Hmm, looks like an earlier version of my commit was squashed into this PR. Definitely it is better to use `PCRE2_ASSERT(s == INT_MAX)`; could you update that?

alexdowad:
This looks good to me.

src/pcre2_compile.c
carenas: That is what happens when you `commit --amend` and `git push --force` without a PR, while providing commit references through comments.

zherczeg:
Very nice patch! Shall we land it?

src/pcre2_substring.c
```diff
@@ -343,7 +343,10 @@ else  /* Matched using pcre2_dfa_match() */
 left = match_data->ovector[stringnumber*2];
 right = match_data->ovector[stringnumber*2+1];
 if (left > match_data->subject_length || right > match_data->subject_length)
+  {
+  PCRE2_DEBUG_UNREACHABLE();
```
carenas, line 347: is this really a debugable failure?, I think this should be reverted and we probably need to add some tests instead

src/pcre2_compile.c
```diff
@@ -10528,7 +10536,9 @@ if ((options & PCRE2_LITERAL) == 0)
           break;
 
           default:
-          PCRE2_UNREACHABLE();
+          /* Fail assertion if a new type was added and someone forgot to
+          update this switch */
+          PCRE2_DEBUG_UNREACHABLE();
```
carenas, line 10541: Funny enough this one was one of the few new asserts that were commented, an after reading the comment think it should had been kept as `PCRE2_UNREACHABLE()`

1. In debug mode will trigger and assert, and assuming the unnamed developer that made the mistake tests locally with a debugging build then it should never reach a release. Note that we do debug builds for all the DEV CI jobs that even if them were careless enough it might be caught there.
2. While the potential for performance improvement here is irrelevant, the whole reason why this macro is even enabled in production is to optimize cases like this, so why it should have been done IMHO.

I am curious why this was changed?, and if (considering we should be well aware on how to use this macros) maybe what we are missing is documentation in HACKING (or somewhere else) that would clarify how to use them so the developer that eventually hits this should had known better?, maybe we can reuse some of the information from Alex's commit that was not used?

src/pcre2_match.c
```diff
@@ -6119,7 +6122,11 @@ fprintf(stderr, "++ %2ld op=%3d %s\n", Fecode - mb->start_code, *Fecode,
       /* It is the end of whole-pattern recursion. */
 
       offset = Flast_group_offset;
-      if (offset == PCRE2_UNSET) return PCRE2_ERROR_INTERNAL;
+      if (offset == PCRE2_UNSET)
+        {
+        PCRE2_DEBUG_UNREACHABLE();
```
carenas, line 6127: maybe `PCRE2_ASSERT(0)`?

src/pcre2_compile.c
```diff
@@ -9382,6 +9388,7 @@ for (;; pptr++)
     /* This should never occur. */
 
     case META_END:
+    PCRE2_DEBUG_UNREACHABLE();
```
carenas, line 9391: assert?

src/pcre2_compile.c
```diff
@@ -5525,6 +5530,7 @@ have duplicate names. Give an internal error. */
 
 if (i >= cb->names_found)
   {
+  PCRE2_DEBUG_UNREACHABLE();
```
carenas, line 5533: assert?

src/pcre2_auto_possess.c
```diff
@@ -456,6 +456,7 @@ switch(c)
        {
        /* Early return if there is not enough space. This should never
        happen, since all clists are shorter than 5 character now. */
+       PCRE2_DEBUG_UNREACHABLE();
```
carenas, line 459: definitely assert

carenas:
This whole time, there was (apparently) a bug in `pcre2_convert()` and [fixing](https://github.com/carenas/pcre2/pull/2) it will also might get rid olf the UNREACHABLE asserts.

I am using a `PCRE2_ASSERT(0)` instead of `PCRE2_DEBUG_ASSERT()` and that I think might be less confusing overall and are equivalent, so, should we keep the new macro then?

src/pcre2_compile.c
carenas: FWIW added some documentation on the code (like we do for other macros/functions) so we have at least a starting point

src/pcre2_util.h
carenas: BTW removed it from the latest revisions, because it was too painful to have it when building on anything outside the 3 compilers we have implementations for, but that means we also have to be extra careful in the opposite direction, because something like:

```c
  case THIS:
     do_good();
     PCRE2_UNREACHABLE();
  case THAT:
     do_bad();
```

is perfectly valid code and will build cleanly on those 3 compilers we have in CI, but result in a dangerous fall through when built by a different compiler.

src/pcre2_auto_possess.c
carenas: in any case, included a little more information that could be used to start a fix for this and that could be also used IMHO as a good example on how to use thess macros.

carenas:
@alexdowad: do you have any comments on the additional points raised?, I think it would be nice to at least document them a little better if you want to keep them in this PR; otherwise I am tempted to (as you suggested) remove them and then you could pick them up and reintroduce them one by one after more careful consideration.

FWIW, I am trusting all the original ones (and maybe some of the new ones that got lost in the sea of changes) were audited and are safe, otherwise.

alexdowad:
> @alexdowad: do you have any comments on the additional points raised?, I think it would be nice to at least document them a little better if you want to keep them in this PR; otherwise I am tempted to (as you suggested) remove them and then you could pick them up and reintroduce them one by one after more careful consideration.

I'll be happy to go through and add more comments. Apologies, but my life has become a bit busy in the last couple of days, so it will be appreciated if you can allow 48 hours or so.

> FWIW, I am trusting all the original ones (and maybe some of the new ones that got lost in the sea of changes) were audited and are safe, otherwise.

I would be happy to make a full pass through again and double-check all occurrences of PCRE2_{,DEBUG_}UNREACHABLE.

carenas:
> Apologies, but my life has become a bit busy in the last couple of days

Sorry, there is no need to worry  or apologize about that, we are volunteers anyway and you can take as much time as needed.

src/pcre2_util.h
alexdowad: Carlo, just a thought... Aside from gcc, clang, and MS Visual C++, do you know any other C compilers which are commonly used to build PCRE2? If so, I can do a bit of research and see if they have a similar built-in.

I understand that even so, the issue will still remain that *some* users, *somewhere* may use a compiler for which PCRE2_UNREACHABLE is not implemented. But it wouldn't be bad to have implementations for as many compilers as possible.

carenas: > do you know any other C compilers which are commonly used to build PCRE2?

Oracle's (AKA Sun) Studio, the latest release doesn't have this (with the C++ version implemented as a NOP internally, so it will even be detected and used but still behave as erratically as described above). IBM's  xlc and DEC's for the VMS are also likely common.

The important point though is not that we "could" add support for them, but the fact that we are opening ourselves to issues like this by using this builtins() that are specifically described as triggering "undefined behaviour" and so will need to be more careful going forward on our code review (I guess) or have to invent something else to protect us from that.

src/pcre2_compile.c
alexdowad: I see that Carlo has added a `MAX_LIST` variable in GenerateUcd.py, though it doesn't seem to be used yet in that script.

Would adding a check around line 851 that throws an exception if `len(s) > (MAX_LIST - 3)` be a good idea? Then anyone who is regenerating the property tables will notice if this is a problem.

The reason for comparing `len(s)` to `MAX_LIST - 3` is because two slots in `list` are used for other purposes, and it needs to be terminated with the special value `NOTACHAR`, so the number of slots available for codepoints are `MAX_LIST - 3`.

If this ever happens, simply increasing the value of `MAX_LIST` in pcre2_auto_possess.c and GenerateUcd.py is all that is needed... then the stack-allocated arrays used to store those values will get bigger and everything will be OK.

alexdowad:
Opened a PR improving comments in this PR a bit more; see PR 3 in carenas/debugass.

src/pcre2_compile.c
alexdowad: Ah... sorry... these comments apply below, not here. Let me copy them to the part which they were intended to address.

src/pcre2_auto_possess.c
alexdowad: I see that Carlo has added a MAX_LIST variable in GenerateUcd.py, though it doesn't seem to be used yet in that script.

Would adding a check around line 851 that throws an exception if len(s) > (MAX_LIST - 3) be a good idea? Then anyone who is regenerating the property tables will notice if this is a problem.

The reason for comparing len(s) to MAX_LIST - 3 is because two slots in list are used for other purposes, and it needs to be terminated with the special value NOTACHAR, so the number of slots available for codepoints are MAX_LIST - 3.

If this ever happens, simply increasing the value of MAX_LIST in pcre2_auto_possess.c and GenerateUcd.py is all that is needed... then the stack-allocated arrays used to store those values will get bigger and everything will be OK.

I'm just thinking more about this... If we adjust GenerateUcd.py so that it crashes if ever ucd_caseless_sets are too big... do we still need the fail-safe in pcre2_auto_possess.c? Would it be enough to simply assert PCRE2_ASSERT(clist_dest < list + MAX_LIST)?

This is not simply a technical issue, but more of a "project policy" issue.

src/pcre2_study.c
```diff
@@ -755,11 +755,12 @@ for (;;)
     new ones get added they are properly considered. */
 
     default:
+    PCRE2_DEBUG_UNREACHABLE();
```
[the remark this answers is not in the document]
alexdowad: It should be unreachable, because it is part of a `switch` which is intended to explicitly list every possible opcode. If it is reached, that means the regex bytecode is corrupted, or a new opcode has been added and the `switch` was not updated.

I'm not very sure what the idea of using `PCRE2_ASSERT(0)` would be; could you explain more?

src/pcre2_substring.c
alexdowad: Looks like this was already removed, which is fine.

src/pcre2_match.c
alexdowad: As above, I would love to understand what the advantage of using `PCRE2_ASSERT(0)` might be.

alexdowad:
I have read through the entire PR diff again. Aside from a couple small improvements to comments, I don't personally see anything else to improve right now.

I think the rationale for `DEBUG_UNREACHABLE` at sites which return `PCRE2_ERROR_INTERNAL` is clear; this error should never really occur unless the library is buggy, and failing a debug assertion in such cases will help fuzzers, etc. to discover such cases.

Aside from `PCRE2_ERROR_INTERNAL`, some non-public, internal-only functions use other values to indicate "can't happen" conditions. For exactly the same reasons, it makes sense to use `DEBUG_UNREACHABLE` in such cases.

There a number of other UNREACHABLE and DEBUG_UNREACHABLE assertions in this PR which are not specifically commented. For most of them, the function has obviously been designed to exit from inside an "infinite loop", and if ever the statement after that loop is reached, then something which the author did not anticipate must have happened. Is there anything else in particular which we should say in the comments? I think `Control flow should not reach here` is fairly clear.

alexdowad:
Ah, one more thing... should I fuzz this PR for a while to see if any of the assertions are triggered?

src/pcre2_match.c
carenas: it is functionally equivalent but reduces the cognitive load of thinking about marking something UNREACHABLE when it "might" happen under some circumstances. It might be as flawed though and has the disadvantage of the failure (if it happens) not be that clear, so I dunno.

IMHO (like we saw in the autopossess case), once it getst better documentation around it we can come out with a much better overall solution and then instead of having an UNREACHABLE we end up with a clearer ASSERT.

alexdowad: OK, let me look at this one another time...

alexdowad: Essentially, we're searching a linked list of heapframes to try to find one which is of type GF_CAPTURE. If `offset == PCRE2_UNSET`, I think it means we reached the end of the linked list without finding it... just studying the code some more...

alexdowad: If that `DEBUG_UNREACHABLE` is reached, I think it means the regex bytecode is malformed. Before OP_CLOSE, there should always be another op which creates a heapframe for the capturing group (such as OP_CBRA).

That sounds like a pretty reasonable place to put DEBUG_UNREACHABLE to me, but anything is fine.

PhilipHazel:
Is this PR ready to go?

carenas:
> Is this PR ready to go?

It improves over the current state by both removing the risk of Undefined Behavior through the use of UNREACHABLE in non debug builds and adding some more failsafes for debug ones.

PR #491, opened 2024-09-23
Substitute: add support for $& $` $' $_ and \b \v

NWilson:
The `$&` replacement, and `` $` `` and `$'` are common sequences for `$0`, and "left of $0" / "right of $0" respectively. Perl supports these, so PCRE2 surely should.

The `$_` is less common but supported by some engines.

---

Finally, very very minor:

The `\b` escape is supported by Perl, so we should add it too.

The `\v` escape is there for completeness - it's not apparently supported by Perl, but many other dialects such as JavaScript or .NET use this C-inspired escape sequence for "vertical tab". It's obviously not possible to support this in patterns, but in the replacement string I can't see a reason not support it.

### DIFF
--- a/src/pcre2_substitute.c
+++ b/src/pcre2_substitute.c
@@ -143,6 +143,8 @@ for (; ptr < ptrend; ptr++)
     switch(erc)
       {
       case 0:      /* Data character */
+      case ESC_b:  /* Data character */
+      case ESC_v:  /* Data character */
       case ESC_E:  /* Isolated \E is ignored */
       break;
 
@@ -647,6 +649,7 @@ do
       BOOL star;
       PCRE2_SIZE sublength;
       PCRE2_UCHAR next;
+      PCRE2_SPTR subptr, subptrend;
 
       if (++ptr >= repend) goto BAD;
       if ((next = *ptr) == CHAR_DOLLAR_SIGN) goto LOADLITERAL;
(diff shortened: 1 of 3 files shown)


zherczeg (approved):
LGTM

carenas:
would be nice to see updated documentation as well (ex: doc/pcre2syntax.3 and maybe even doc/pcre2pattern.3)

the html is generated with the help of some scripts that are called from `PrepareRelease`, so those files might not need to be included, but sure are nicer to generate locally for a quick inspection.

NWilson:
@PhilipHazel Thank you for merging, and adding documentation!

I saw in your documentation that you have documented the extra substitutions `$&` etc to require `PCRE2_SUBSTITUTE_EXTENDED`. But actually, I implemented them to be available always (like `$0` is). They are intended to be Perl-compatible, unlike the `${name:+yes:no}` syntax.

Would you like me to open a PR to change the code to require PCRE2_SUBSTITUTE_EXTENDED, or should I update the documentation?

carenas:
> They are intended to be Perl-compatible

I think it will be nice to keep the code as-is and update the Documentation, but what is also really missing is a way to put the test that is perl compatible on testinput1 (which means we need to add functionality to both pcre2test and also perltest.sh) and which will also help in the future testing substitutions better IMHO

PhilipHazel:
Oh , sorry I missed that. Obvious, really, but as your test of these had SUBSTITUTE_EXTENDED set, I assumed that was needed, without checking. I will fix the documentation , and also update the test. // When I implemented substitution I wasn't really thinking about Perl-compatibility at all, I have to confess. That's why I never extended perltest.sh.

PR #492, opened 2024-09-23
Reserve syntax for (unimplemented) extended character classes

NWilson:
I'm not expecting this to be merged - I'm guessing you don't want this code until the "extended classes" are actually implemented!

But this is a "for your information" PR. We will be shipping this diff in our product.

It is benign: if there is an unescaped '[' or '--', '&&' sequence inside a character class, then it rejects it. These are the new metacharacters in extended character classes. As you can see from the tests, this has backwards-compatibility implications, so even though we haven't implemented the feature yet, we'd like to "reserve" the syntax for future use in our application.

Do let me know if there's anything bad about this code though!

I'll be following up with the actual implementation in due course.


carenas:
Perl's (`?[...]`) own syntax should also generate the same error with this flag IMHO

PR #495, opened 2024-09-24
ci: add a full clang integer sanitizer DEV job

carenas:
Make sure that no undesired integer wraparounds happen in the code, and that it is clear from other common issues.

It includes a "fix" in `pcre2_match.c` that could be handled better by the sanitizer and could be considered
a "false positive", and that future versions of the compiler would be able to avoid by using (the yet not released) `-fsanitize-undefined-ignore-overflow-pattern=unsigned-post-decr-while` option.

### DIFF
--- a/src/pcre2_match.c
+++ b/src/pcre2_match.c
@@ -5937,8 +5937,11 @@ fprintf(stderr, "++ %2ld op=%3d %s\n", Fecode - mb->start_code, *Fecode,
 #ifdef SUPPORT_UNICODE
     if (utf)
       {
-      while (number-- > 0)
+      /* We used to do a simpler `while (number-- > 0)` but that triggers
+      clang's unsigned integer overflow sanitizer. */
+      while (number > 0)
         {
+        --number;
         if (Feptr <= mb->check_subject) RRETURN(MATCH_NOMATCH);
         Feptr--;
         BACKCHAR(Feptr);
@@ -6870,9 +6873,6 @@ if (use_jit)
 #ifdef SUPPORT_UNICODE
   if (utf && (options & PCRE2_NO_UTF_CHECK) == 0 && !allow_invalid)
     {
-#if PCRE2_CODE_UNIT_WIDTH != 32
-    unsigned int i;
-#endif
 
     /* For 8-bit and 16-bit UTF, check that the first code unit is a valid
     character start. */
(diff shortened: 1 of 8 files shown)


src/pcre2_match.c
```diff
@@ -5937,8 +5937,11 @@ fprintf(stderr, "++ %2ld op=%3d %s\n", Fecode - mb->start_code, *Fecode,
 #ifdef SUPPORT_UNICODE
     if (utf)
       {
-      while (number-- > 0)
+      /* We used to do a simpler `while (number-- > 0)` but that triggers
+      clang's unsigned integer overflow sanitizer. */
+      while (number > 0)
```
zherczeg, line 5942: I think number cannot be 0 (no reverse), so `--number > 0` could work as well.

src/pcre2test.c
```diff
@@ -7182,14 +7182,14 @@ while ((c = *p++) != 0)
...
-      least one MacOS environment. */
+      least one macOS environment. */
 
       for (pt++; isxdigit(*pt); pt++)
         {
         if (++i == 9)
           fprintf(outfile, "** Too many hex digits in \\x{...} item; "
                            "using only the first eight.\n");
-        else c = c * 16 + tolower(*pt) - ((isdigit(*pt))? '0' : 'a' - 10);
+        else c = c * 16 + (tolower(*pt) - ((isdigit(*pt))? '0' : 'a' - 10));
```
carenas, line 7192: nitpick: remove the strange additional parenthesis around `isdigit()`?

src/pcre2_match.c
carenas: that is what I had [originally](https://github.com/PCRE2Project/pcre2/commit/6c07f89084934af619ac9842b9616b653f8818b0) but I think this looks slightly less ackward.

specially when you add the (what "if somehow number becomes 0 in a future implementation")

src/pcre2test.c
zherczeg: the number is UINT16, and you can add +1 if you worry about that

carenas: not sure I understand what you mean by that, but it sure doesn't worry me, it is just distracting and makes me curious.

zherczeg: Github added the comment to the wrong place somehow. It is a comment for above.

src/pcre2_match.c
zherczeg: The number is computed by GET2, so it is uint16_t, and adding +1 could solve the possible 0 value.

carenas: good idea, but I sill think that just the current version shows intent better and decouples better the concerns that would raise.

zherczeg (approved):
LGTM

PR #496, opened 2024-09-25
Simplify range data construction.

zherczeg:
This patch uses the computed ranges to generate byte code rather than using `add_to_class`. It is a considerable simplification of the code.


src/pcre2_compile_class.c
```diff
@@ -372,9 +372,9 @@ while (*ptr != META_CLASS_END)
 #ifdef SUPPORT_UNICODE
     if (options & PARSE_CLASS_CASELESS_UTF)
       {
-      size_t usize = utf_caseless_extend(start_char, *ptr++, options, buffer);
-      if (buffer != NULL) buffer += usize;
-      total_size += usize;
+      size = utf_caseless_extend(start_char, *ptr++, options, buffer);
```
carenas, line 375: FYI this will revert Phillip's last fix for `-Wshadow`, is that intended?

zherczeg: Yes. I wanted to remove the `size_t` but forgot it.

carenas: nevermind, shold had pulled it first before commenting, guess that is why it was using `size` in the original to begin with then ;)

zherczeg: Oh no. I found another issue with negated ascii classes. And this is still just the range merge, the logarithmic search is still very far.

zherczeg:
@PhilipHazel if you agree my suggestion in https://github.com/PCRE2Project/pcre2/issues/497 , this patch is ready

carenas:
I think the following assertion is not correct:

> Obviously a character class and its negated form cannot match to the same character

in PCRE2 it can, and the reasons are historic and described in #186.

In summary our "/u" Perl equivalent requires both utf and ucp modifies to be set

zherczeg:
I am not sure I understand that part, it talks about configuring the modifier. Normally if `[C]` matches to something, `[^C]` must not match to that except for invalid utf characters, which never matches to anything like NaN in numbers.

carenas:
The point is that without `PCRE2_UCP`(as you pointed out) all characters above 255 (in the 8-bit library) are not defined, so any [^C] would match them if `PCRE2_UTF` is enabled.

Agree with you that they "shouldn't" match and that is arguably a bug, but it is the currently expected behaviour when ONLY one of those options are set.

carenas:
indeed I think this might had just introduced a regression:

```
  re> /[^\D\P{Nd}]/B,utf,ascii_bsd
        Bra
        [^\x00-/:-\xff\P{Nd}]
        Ket
        End
  re> /[^\D\P{Nd}]/B,utf,ascii_bsd
        Bra
        [^\x00-/:-\xff\P{Nd}\x{100}-\x{10ffff}]
        Ket
        End
```

zherczeg:
It is fixed that regression, and this is what I am talking about. \D matches anything not [0-9], which includes all > 255 characters. 
```
  re> /[\d]/B,utf
        Bra
        [0-9]
        Ket
        End
  re> /[\D]/B,utf
        Bra
        [\x00-/:-\xff] (neg)
        Ket
        End
```

The `[\x00-/:-\xff] (neg)` is the same as `[\x00-/:-\xff\x{100}-\x{10ffff}]`. This is negated above with `^`.

PR #498, opened 2024-09-27
Fix handling of \g<0> in pcre2_substitute

NWilson:
I made a mess-up, and didn't notice because I foolishly didn't add a test for this case.

The `check_escape()` function returns a negative value for backreferences, which is fine for backrefs in a pattern, because they can't be zero. But the special `\g<...>` handling I added to `pcre2_substitute()` does need to support a zero backref.

I hummed and haa'd about the right way to do it. The `read_number()` helper isn't even exported for `pcre2_substitute.c` to use, and it feels like this is the responsibility of `check_escape()`.

In the end, I've just offset the values by 1, and checked all four callsites of `check_escape()` to make sure they are all updated to cope with that.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -1430,7 +1430,7 @@ return yield;
 
 /* This function is called when a \ has been encountered. It either returns a
 positive value for a simple escape such as \d, or 0 for a data character, which
-is placed in chptr. A backreference to group n is returned as negative n. On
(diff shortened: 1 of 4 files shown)


PhilipHazel:
This is a typical case of "If I'd known about this when I started I would have done this differently" but given the current need for \g<0> I guess this hack is as good as any.

PR #503, opened 2024-09-29
Unicode: Update to UCD 16

Ayesh:
Updates to Unicode files to Unicode 16, adjusts tests, and the scripts used to parse UCD, to adapt to minor formatting differences in UCD 16.

The `GenerateTest26.py` and `GenerateCommon.py` had a regexp to extract properties from the `ScriptExtensions.txt` file. Previously, all property lines had one space after space-separated list of scripts. In UCD-16, this list is adjusted with right-padding, which throws off the parser.

This PR adjusts the regexps to ignore padding spaces.


carenas:
Ideally; there should be also some tests added in `maint/ucptestdata` that exercise the new characters.

no need to keep all changes in independent commits (specially if they are fixups) as they will be squashed together before merging.

do you happen to know when Perl will be upgrading to this Unicode version?

Ayesh:
Thank you @carenas, I updated `ucptestdata` adding two new scripts in Unicode 16. I don't think we actively run and test `ucptest` in CI, because I tried with `find script DoesNotExist` and it still passed.

> do you happen to know when Perl will be upgrading to this Unicode version?

I do not know internals enough to know this. From what I see, 3.12.6 has UCD-15.

carenas:
> I do not know internals enough to know this.

Let me retarget the question then; why do you need UCD-16 in PCRE2?, Perl is still in 15.0 eventhough 15.1 was released a year ago.

Ayesh:
> why do you need UCD-16 in PCRE2?

I misunderstood then, sorry.

For background, we use PCRE2 in PHP, in the `pcre` extension. Although it's possible to provide any PCRE2 source, we also embed PCRE2 source in the `pcre` extension code, along with UCD data. We have an extension called `mbstring`, which also embeds extracted UCD data.

As part of the upcoming PHP 8.4 release, the `mbstring` extension received the UCD-16 update, but this leaves us with a situation that our embedded PCRE2 extensions has a different UCD version compared to `mbstring`. I wanted to see if we can equalize this, to avoid [rare but possible edge cases like this](https://php.watch/versions/8.4/unicode-16#parity).

zherczeg:
I don't mind an upgrade, but would be good to have some tests for the newest features.

PhilipHazel:
I don't mind either, but please also add a note to ChangeLog, which is where all the other Unicode updates are logged.

carenas (changes requested):
need to update and verify ALL ucptestdata:

```
$ ./ucptest < ucptestdata/testinput1 > ucptestdata/testoutput1
$ ./ucptest < ucptestdata/testinput2 > ucptestdata/testoutput2
```

maint/GenerateTest26.py
```diff
@@ -55,7 +55,7 @@ def gen_script_tests():
   script_data = [None] * len(script_names)
   char_data = [None] * 0x110000
 
-  property_re = re.compile("^([0-9A-F]{4,6})(?:\\.\\.([0-9A-F]{4,6}))? +; ([A-Za-z_ ]+) #")
+  property_re = re.compile("^([0-9A-F]{4,6})(?:\\.\\.([0-9A-F]{4,6}))? +; ([A-Za-z_ ]+[A-Za-z]) +#")
```
carenas, line 58: it seems that the original code included spaces by mistake and should had been `([A-Za-z_]+)`, why was the expression changed?

carenas (changes requested):
need to rebase and squash all commits

testdata/testinput26
```diff
@@ -11,17 +11,17 @@
 
 # Script extension check
 /^\p{Latin}/utf
-  \x{363}
+  \x{b7}
```
zherczeg, line 14: Why these characters are changed? Maybe it would be better to keep both.

carenas: UCD 16 just added that character (which is not even a letter) to a lot of scripts, wonder if we should keep test26 AS-IS and use this for a new test27 instead.

maint/GenerateTest26.py
Ayesh: The format went from something like this
```
A9CF          ; Bugi Java # Lm       JAVANESE PANGRANGKEP
```

to:

```
A9CF          ; Bugi Java                      # Lm      JAVANESE PANGRANGKEP
```

The previous expression's `[A-Za-z_ ]+` capture-group matches the space between `Bugi` and `Java` for example, as well as all the trailing spaces except for the last one. So the `split(" ")` call was thrown off. I adjusted it to capture all in-between spaces, but not the trailing spaces used in the UCD-16's for padding.

Ayesh:
Rebased from the latest in the master branch.

testdata/testinput26
Ayesh: Adding a new test like this is a bit more than what I can chew :(

You are right, the test26 diff got unnecessarily big because of these changes.

maint/GenerateTest26.py
carenas: The point I was trying to make is that the same code change is also most likely needed in the other places where this file is processed.

Ayesh: Thank you, you are right, there was indeed one instance of `ScriptExtensions.txt` parsing in the `GenerateCommon.py` script that had a similar regex as well, which I adjusted and force-pushed.

testdata/testinput26
carenas: It is actually not that difficult, you would need to hand modify testinput26 so it passes (something like I posted above would work) move the generated files with a `*27` name instead and add that test to `RunTest` and `RunTest.bat`

carenas (changes requested):
Looks like the last rebase accidentally dropped the changes to ucptestdata/testoutput1, can you add those back, and also make sure none of the Unicode files got munged and maybe regenerate the code as well so that the latest changes on the generators are applied?, sorry about that.

FWIW if the changes would be squashed in a single commit, rebases are usually easier too.

Ayesh:
You are right @carenas I had indeed missed the changes in `ucptestdata/testoutput1`.

I reworked the whole thing from scratch...

1. Downloaded fresh UCD 16.
2. Moved and replaced files in `maint/Unicode.tables` with the files downloaded.
3. Adjusted regexes in `GenerateTest26.py` and `GenerateCommon.py` to be compatible with space padding used in the `ScriptExtensions.txt` file.
4. Generate Test26 with `python3 GenerateTest26.py`
5. Compile and generate ucptestfiles with `./ucptest < ucptestdata/testinput1 > ucptestdata/testoutput1` and `./ucptest < ucptestdata/testinput2 > ucptestdata/testoutput2`
6. Move generated files to replace existing files.
7. Commit, force-push, and cross fingers :)

carenas:
>  cross fingers

A little strange you didn't mention about running the Generators to recreate the C files as well, but did it for you and there were no changes to the ones provided.

I also checked again that the test output in ucpdata for the current tests made sense, and lamented again we have missed the opportunity to add a new test for the newly added U+1FAE9 or some of the other new characters, but I guess we could do those in a following commit, since one would be likely needed anyway.

carenas (approved):
the unnecessary changes to PCRE's test26 will be reverted (as suggested) in a following change, that is also likely to add a little more tests

PR #505, opened 2024-09-30
ci: add perl DEV job

carenas:
Use the latest dev version of Perl to confirm all the perl compatible tests match.

While at it, use `cmake` in Linux with `Ninja` for better coverage.


zherczeg (approved):
LGTM

PR #506, opened 2024-09-30
Add property data to classbits

zherczeg:
Another perf optimization, the properties are added to the class bits. Currently, it keeps the rule that if a class has no explicit < 256 characters, there is no bitset.

```
/[a-c\pN]++/utf
  a345bc345ba22c

Old compile time   2.9976 microseconds
New compile time   4.2943 microseconds
Old match time     0.3723 microseconds
New match time     0.2841 microseconds
```

As expected, the compile time is lower, but the matching time is better for c < 256 characters.


src/pcre2_intmodedep.h
```diff
@@ -749,6 +749,7 @@ typedef struct compile_block {
   PCRE2_SIZE workspace_size;       /* Size of workspace */
   PCRE2_SIZE small_ref_offset[10]; /* Offsets for \1 to \9 */
   PCRE2_SIZE erroroffset;          /* Offset of error in pattern */
+  uint8_t classbits[32];           /* Temporary store for classbits */
```
zherczeg, line 752: This reduces stack usage, since it is allocated only once now.

PR #513, opened 2024-10-05
complete update to Unicode 16

carenas:
A fixup to #503 that keeps the original generated properties test, allowing for both Perl compatibility tests and wider coverage.

Updates the code to complete the changes needed as well as the documentation.


zherczeg:
Nice change! If we do automatically generated files, perhaps we could put more characters into it. Something like min, max, middle, and a test with min-1, min, middle, max, max+1 characters (total of five characters, the test use `\p{script}++`)

PhilipHazel:
You forgot to update Makefile.am with the new test files; I have fixed this minor oversight.

Issue #11, opened 2021-08-23
Caseless ASCII matching


PhilipHazel:
This is #2625 in the old Bugzilla, submitted by Rich Siegel.

RS: Given this text represented as UTF-16:

    this is a test

Search using this pattern, which as written should *not* match any ASCII characters:

    [\x{00FF}-\x{FFEE}]

If the pattern was compiled with PCRE2_CASELESS turned on, pcre2_match() will return a match at the first "s" in the subject text, even though that is outside the explicit range of characters. (And the uppercase version "S" would be, as well.)

Further testing shows that "k" and "K" are matching as well, presumably with the same underlying cause.

ZH: This is expected. Please check:
http://www.unicode.org/Public/12.1.0/ucd/CaseFolding.txt

Examples:

017F; C; 0073; # LATIN SMALL LETTER LONG S
212A; C; 006B; # KELVIN SIGN

In other words, a caseless 0x212A codepoint matches to K.

RS: Thank you! I'm a little ashamed to admit that Unicode case folding didn't even occur to me. This may pose more of a UI and/or documentation challenge, since I'm quite certain that it won't occur to end users either. :-)

Does it make sense to consider a flag in the PCRE2_EXTRA* space which would limit case folding to the ASCII range when PCRE2_CASELESS is specified? (I'm not yet advocating for it; I can see some clear limitations and disadvantages, and trying to express all of the possible variations could rapidly turn into a snake pit.)

PH: I think the best way of doing this would be to add PCRE2_ASCII_CASELESS to the main options, because having two separated flags seems very untidy. There are only two bits left in the main options, so I am slightly reluctant, but on the other hand leaving them unused just in case something more important comes along could last for ever. Zoltan, what do you think? Implementing this would need changes to JIT as well as the interpreters.

On further reflection, I've changed my mind and think that a PCRE2_EXTRA option would be better, as you suggested, partly because there may be a number of variations needed. And indeed, some additions to the documentation.

ZH: 
I feel this is a non trivial change, and it can be easy done on pattern level.

The issue here is caseless, and you can temporarily disable it:

(?-i:[\x{00FF}-\x{FFEE}])

Or use an assertion on which separates ascii from the rest.

PhilipHazel:
Implemented in commit a6089462a (with some later bug fixes).

PR #516, opened 2024-10-07
Fix OP_REFI for caseless_restrict

NWilson:
When you introduced caseless_restrict earlier in the year, it looks like you forgot to add it to the `OP_REFI` (and `OP_DNREFI`) comparison.

It's actually a bit tricky to add, because it seems the "flags" that are tracked during compile are not written out to the bytecode.

So, I added one extra field to `OP_REFI` to record the caseless flag. (Currently, it's just "caseless_restrict", but as I mentioned, I'll be adding "turkish_casing" in a future PR.)

I _think_ there's no other way to do it? I hope the flags aren't stored in the bytecode already and I just missed it.

It seems a bit wasteful, perhaps, to spend a uint8 (or uint32 in the case of the 32-bit library), but I think you basically forced this outcome back when you added the `(?r)` flag allowing caseless_restrict to be varied through the pattern.

### DIFF
--- a/HACKING
+++ b/HACKING
@@ -365,8 +365,10 @@ Changeable options
 The /i, /m, or /s options (PCRE2_CASELESS, PCRE2_MULTILINE, PCRE2_DOTALL) and
 some others may be changed in the middle of patterns by items such as (?i).
 Their processing is handled entirely at compile time by generating different
-opcodes for the different settings. The runtime functions do not need to keep
-track of an option's state.
+opcodes for the different settings. Some options are copied into the opcode's
+data, for opcodes such as OP_REFI which depends on the (?r)
+(PCRE2_EXTRA_CASELESS_RESTRICT) option. The runtime functions do not need to
+keep track of an option's state.
 
 PCRE2_DUPNAMES, PCRE2_EXTENDED, PCRE2_EXTENDED_MORE, and PCRE2_NO_AUTO_CAPTURE
 are tracked and processed during the parsing pre-pass. The others are handled
@@ -638,6 +640,9 @@ generates OP_DNREF or OP_DNREFI. These are followed by two counts: the index
 (not the byte offset) in the group name table of the first entry for the
 required name, followed by the number of groups with the same name. The
 matching code can then search for the first one that is set.
+
+OP_REFI and OP_DNREFI are further followed by an item containing any
+case-insensitivity flags.
 
 
 Repeating character classes and back references
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -7188,6 +7188,9 @@ for (;; pptr++)
       *code++ = ((options & PCRE2_CASELESS) != 0)? OP_DNREFI : OP_DNREF;
       PUT2INC(code, 0, index);
       PUT2INC(code, 0, count);
+      if ((options & PCRE2_CASELESS) != 0)
+        *code++ = ((xoptions & PCRE2_EXTRA_CASELESS_RESTRICT) != 0)?
+            REFI_FLAG_CASELESS_RESTRICT : 0;
       }
     break;
 
@@ -8142,6 +8145,9 @@ for (;; pptr++)
     if (firstcuflags == REQ_UNSET) zerofirstcuflags = firstcuflags = REQ_NONE;
     *code++ = ((options & PCRE2_CASELESS) != 0)? OP_REFI : OP_REF;
     PUT2INC(code, 0, meta_arg);
+    if ((options & PCRE2_CASELESS) != 0)
+      *code++ = ((xoptions & PCRE2_EXTRA_CASELESS_RESTRICT) != 0)?
+          REFI_FLAG_CASELESS_RESTRICT : 0;
 
     /* Update the map of back references, and keep the highest one. We
     could do this in parse_regex() for numerical back references, but not
--- a/src/pcre2_internal.h
+++ b/src/pcre2_internal.h
@@ -1780,9 +1780,9 @@ in UTF-8 mode. The code that uses this table must know about such things. */
   1+(32/sizeof(PCRE2_UCHAR)),    /* NCLASS                                 */ \
   0,                             /* XCLASS - variable length               */ \
   1+IMM2_SIZE,                   /* REF                                    */ \
-  1+IMM2_SIZE,                   /* REFI                                   */ \
+  1+IMM2_SIZE+1,                 /* REFI                                   */ \
   1+2*IMM2_SIZE,                 /* DNREF                                  */ \
-  1+2*IMM2_SIZE,                 /* DNREFI                                 */ \
+  1+2*IMM2_SIZE+1,               /* DNREFI                                 */ \
   1+LINK_SIZE,                   /* RECURSE                                */ \
   1+2*LINK_SIZE+1,               /* CALLOUT                                */ \
   0,                             /* CALLOUT_STR - variable length          */ \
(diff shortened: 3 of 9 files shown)


NWilson:
@zherczeg Zoltan, I've half-heartedly updated the JIT code here. It hardcodes the length of the opcodes in many places, so after extending OP_REFI and OP_DNREFI I had to find all references I could.

The JIT code seems to be working on all the existing tests, but fails on the one new test I added, which exercises the `caseless_restrict` behaviour in a backref.

I'd be grateful if you'd be able to help me out on that.

zherczeg:
Nice catch! I am not sure which one is better, adding a new opcode (only the case insensitive variant may use restrict), or adding an argument. I would prefer a new opcode. @PhilipHazel what do you think?

NWilson:
When I add `PCRE2_EXTRA_TURKISH_CASING` it would double the number of possibilities, so it would end up as 4 opcodes. That's not terrible, so I suppose either is possible.

NWilson:
Unfortunately, all four options are possible, in theory (-r -turk, +r -turk, -r +turk, +r +turk).

And it can't use global options because Philip already added an inline option `/(?r: ... )/` rather than making it a pre-pattern option instead.

zherczeg:
I completely forgot that it is not a pre-pattern. I simply used pattern flags here:
https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_compile_class.c#L448

No test failed. Anyway it could be fixed by passing extra options.

zherczeg:
JIT fix https://github.com/PCRE2Project/pcre2/commit/ae118788448e669fef8997ab9ac3f9dbb20daf4c

Added some test as well. It turned out that DNREFI support was not added to jit :(

Feel free to use the commit, no need to mention me in the patch.

PhilipHazel:
I see the CI tests are failing. 

Historical note: the original PCRE did keep track of the options during matching, and there was (for example) only OP_CHAR, not OP_CHARI. However, all the saving and restoring got complicated and here is a ChangeLog entry for 8.13:

> 9.  Some internal refactoring has changed the processing so that the handling  
>     of the PCRE_CASELESS and PCRE_MULTILINE options is done entirely at compile
>     time (the PCRE_DOTALL option was changed this way some time ago: version   
>     7.7 change 16). This has made it possible to abolish the OP_OPT op code,   
>     which was always a bit of a fudge.

Doing all the options handling at compile time should also in theory make matching a little bit faster. PCRE1 still uses the stack for backtracking, and the above change reduced the number of arguments to the recursive function, which helped with stack usage. This is not relevant for PCRE2 since the 10.30 refactoring.

I added (?r) because it seemed right to make it the same as other options, but also so that pattern creators who have no access to the calling code can set it for the whole pattern in the same way as (?i) etc., though it could have been (*CASELESS_RESTRICT).

As for whether to add a new opcode or an argument, either will require checking every reference to the original, so much the same amount of work. We are not short of opcodes so it may make sense to add four but isn't it eight? Four versions of OP_REFI and four versions of OP_DNREFI, or am I missing something? Eight is rather a lot; if I'm right about that, then perhaps an argument makes more sense. But I don't really mind which you do, though multiple opcodes might execute faster? BUT there are actually quite a lot of opcodes that end in ...I. Do they all need looking at now there are (will be) four different ways of doing a caseless match?

zherczeg:
Oh, a `/* Fall through. */` is missing from the patch.

carenas:
> Oh, a `/* Fall through. */` is missing from the patch.

Fixed in ac76507bf515628446a5ace942ca503d03e5dca5; took me a while to find the "standard" spelling though since it seems only sljit has those until now.

NWilson:
> Historical note: the original PCRE did keep track of the options during matching, and there was (for example) only OP_CHAR, not OP_CHARI. However, all the saving and restoring got complicated and here is a ChangeLog entry for 8.13:

Thank you Philip, that's really useful and interesting!

> I added (?r) because it seemed right to make it the same as other options, but also so that pattern creators who have no access to the calling code can set it for the whole pattern in the same way as (?i) etc., though it could have been (*CASELESS_RESTRICT).

There's no reason to regret it now, it's most-flexible and works well.

> We are not short of opcodes so it may make sense to add four but isn't it eight? Four versions of OP_REFI and four versions of OP_DNREFI, or am I missing something?

Yes, it would be four each, you've counted correctly.

> BUT there are actually quite a lot of opcodes that end in ...I. Do they all need looking at now there are (will be) four different ways of doing a caseless match?

Aha, no! Because the other `OP_...I` opcodes have quite specific behaviours tied to the default case-folding relationship. For example, OP_CHARI is fixed to match exactly two characters (the one that follows it in the bytecode, and that char's "other case"). The notion of "other case" isn't affected by caseless-restrict, so no change is needed.

For caseless-restrict, it really is just the REFI and DNREFI code that needs to care about the specifics of the case equivalence.

NWilson:
> IMHO the new "turk" flag would be incompatible with +r, because by definition what the later does is IGNORE any othercase that crosses the ASCII/UTF boundary (see Perl's [/aa mode](https://perldoc.perl.org/perlre#/a-(and-/aa))).

I think they're compatible, logically. `(?r)` removes the case-folding rules that cross the ASCII/non-ASCII boundary (which is actually just two characters). And Turkish _adds_ two case-folding rules for the Turkish alphabet. They aren't touching the same characters, so they both make sense.

Maybe you don't like the "Kelvin sign"? Not a problem - you can remove that, and you still get to choose whether to have the Turkish mappings or not.

To put it another way: if caseless-restrict is useful, why shouldn't Turkish users be able to select it?

NWilson:
Thank you very much all three of you for your help, on this very fussy little detail fix!

carenas:
> To put it another way: if caseless-restrict is useful, why shouldn't Turkish users be able to select it?

because it is useful only in an specific context (see #11, although the Perl link is probably clearer IMHO), where the user intentionally wants to make sure that a caseless match is only within (or not) ASCII to avoid "surprises".

a "turk" flag user wants to have Unicode characters, and indeed will add 2 more cross between ASCII and Unicode when doing caseless matches, which is what `(?r)` is meant to prevent.

zherczeg:
@carenas I see `/* Fall through */` in the other places as well, if you prefer to use `/* fallthrough */`, change it everywhere (in another patch). I also prefer one patch for one PR.

zherczeg:
I hope Perl does not plan to use 'r' for something different.

NWilson:
Rebased, and fixed to "Fall through" for consistency.

carenas (approved):
Minor nitpick, but since we are not doing multiple OPs instead of the flag, the following also needs updating in HACKING IMHO:

https://github.com/PCRE2Project/pcre2/blob/dcbf9a0a135350344275830083475ac2edcbf88a/HACKING#L362-L369

zherczeg (approved):
LGTM

zherczeg:
@carenas if you are ok with the patch, I will land it. @NWilson please update the test conflicts and merge commits

NWilson:
Rebased, squashed, and fixed the conflicts with the PR you just merged.

carenas (approved):

carenas:
Usually Philip follows up with some documentation in Changelog, which this change (and most others) were missing.

carenas:
I didn't meant from my comment to be a policy discussion, just a reminder for a task that needed to be done and indeed an excuse for myself to do it as part of #519.

I do think though that keeping the ChangeLog in good shape is important, and also that is done in a timely way so that it could be described properly (specially considering that commits and PR descriptions as of now are not consistently used to work as thes source of an automated replacement).

Updating them with the committed changes is not ideal though (as they will likely result in conflicts), so probably 2 might be the only reasonable short term option?

PR #518, opened 2024-10-08
Fix case restricted range processing

zherczeg:
I totally forgot about the 'r' flag. I am sorry.


carenas (changes requested):
A copy of the test for the non 8bit libraries without UTF but only UCP would be nice

zherczeg:
Tests copied

NWilson (approved):

carenas (approved):

PR #519, opened 2024-10-08
pcre2grep: add $& as an alias for $0

carenas:
Make `pcre2grep` consistent with recent changes on the syntax used to refer to the matched text and that reflect recent Perl as well.

While at it, update the documentation and other somehow unrelated minor changes.


PhilipHazel:
The documentation pcre2grep.1 needs the addition of $& to the section describing --output. This is a minor backwards incompatibility because previously $& would have inserted &.

carenas:
> &amp; would have inserted &.

I couldn't find that in my sources, so generated the other files and inspected them by hand, so hopefully this is not an issue

zherczeg (approved):
LGTM

PR #521, opened 2024-10-08
Add support for Turkish 'i' casing

NWilson:


maint/GenerateUcd.py
```diff
@@ -855,6 +862,17 @@ def write_bitsets(list, item_size):
...
 f.write('};\n\n')
 
+# --- Output the indices of the Turkish caseless character sets ---
+
+f.write("""\
+/* These are the indices, within ucd_caseless_sets, of the additional
+Turkish case-equivalences. */
+
+const uint32_t PRIV(ucd_turkish_dotted_i_caseset) = %d;
+const uint32_t PRIV(ucd_turkish_dotless_i_caseset) = %d;
```
NWilson, line 872: I don't feel very happy about adding these.

I wanted a simple and vaguely efficient way to transform `/[i]/i` → `/[iİ]/` at compile-time, so I wouldn't need to change anything in the matcher. How to neatly represent a set of exactly two characters? I thought I could re-use the existing PT_CLIST mechanism.

It's vaguely tidy. But even better would be an "OP_2CHAR" that gives a dedicated efficient encoding for either-of-two characters.

src/pcre2_match.c
```diff
@@ -422,8 +423,13 @@ if (caseless)
...
 
-      ur = GET_UCD(d);
-      if (c != d && c != (uint32_t)((int)d + ur->other_case))
+      if (turkish_casing && UCD_ANY_I(d))
+        {
+        c = (c == 0x0130)? 0x69 : (c == 0x49)? 0x0131 : c;
+        d = (d == 0x0130)? 0x69 : (d == 0x49)? 0x0131 : d;
+        if (c != d) return -1;  /* No match */
+        }
+      else if (c != d && c != (uint32_t)((int)d + (ur = GET_UCD(d))->other_case))
```
NWilson, line 432: This will need to be added to the JIT code too.

doc/html/pcre2api.html
```diff
@@ -2223,6 +2234,12 @@ <h1>pcre2api man page</h1>
 returning PCRE2_ERROR_CALLOUT_CALLER_DISABLED. This is useful if the application
 knows that a callout will not be provided to <b>pcre2_match()</b>, so that
 callouts in the pattern are not silently ignored.
+<pre>
+  PCRE2_EXTRA_TURKISH_CASING
+</pre>
+This option alters case-equivalence of the 'i' letters to follow the
+alphabet used by Turkish and Azeri languages. The option can be changed within
+a pattern by the (?t) option setting.
```
zherczeg, line 2242: Is this that important to reserve a letter for it? I don't remember anybody requested this before.

maint/GenerateUcd.py
zherczeg: Using PT_CLIST is a good idea.

doc/html/pcre2api.html
NWilson: I don't feel it's important. I just added it for consistency with `(?r)`. I'm very happy either way.

maint/GenerateUcd.py
NWilson: In another thread, I discussed getting rid of PT_CLIST, in which case I'd replace it with a simpler "OP_2CHAR".

doc/pcre2api.3
```diff
@@ -1986,9 +1993,11 @@ The second effect of PCRE2_UCP is to force the use of Unicode properties for
 upper/lower casing operations, even when PCRE2_UTF is not set. This makes it
 possible to process strings in the 16-bit UCS-2 code. This option is available
 only if PCRE2 has been compiled with Unicode support (which is the default).
-The PCRE2_EXTRA_CASELESS_RESTRICT option (see below) restricts caseless
+The PCRE2_EXTRA_CASELESS_RESTRICT option (see above) restricts caseless
 matching such that ASCII characters match only ASCII characters and non-ASCII
-characters match only non-ASCII characters.
+characters match only non-ASCII characters. The PCRE2_EXTRA_TURKISH_CASING option
+(see above) alters the matching of the 'i' characters as they used in Turkish
```
carenas, line 1999: they are used

mentioning here expicitly that there is no conflict between those two options encouraged IMHO.

doc/pcre2unicode.3
```diff
@@ -147,6 +147,28 @@ Recognition of these non-ASCII characters as case-equivalent to their ASCII
...
+      'k' = 'K' = U+212A (Kelvin sign)
+      's' = 'S' = U+017F (long S)
+    With PCRE2_EXTRA_CASELESS_RESTRICT:
+      'k' = 'K'
+      U+212A (Kelvin sign)  only case-equivalent to itself
+      's' = 'S'
+      U+017F (long S)       only case-equivalent to itself
+.P
+One language family, Turkish and Azeri, has its own case-insensitivity rules,
+which can selected by setting PCRE2_EXTRA_TURKISH_CASING. This alters the
```
carenas, line 161: ditto

maint/GenerateUcd.py
```diff
@@ -737,6 +737,13 @@ def write_bitsets(list, item_size):
     if x > 127 and x + other_case[x] < 128:
       other_case[x] = 0  
 
+# Append a couple of extra caseless sets (unreferenced by the record objects)
```
carenas, line 740: the records existin the source file and were skipped (they have a T "type" and we only process the ones with "C" or "S".

I am confused about this though, as it would seem you are hardcoding the values everywhere else.

maint/GenerateUcd.py
carenas: I would prefer a new `OP_VCHARI` that would point to the correct entry on this table depending of PCRE2_EXTRA_TURKISH_CASING. They are always 2 characters equivalences anyway while the casesets was meant to be used by 3 or more.

maint/GenerateUcd.py
```diff
@@ -980,6 +998,6 @@ def write_bitsets(list, item_size):
 /* End of pcre2_ucd.c */
 """)
 
-f.close
+f.close()
```
carenas, line 1001: this is unnecessary and might had been intentional (I think i'd seen more of those as well), so maybe better done independently (if needed)

src/pcre2_match.c
carenas: making your tests `\=no_jit` could help avoiding breaking the CI.
if using a new OP it will likely work even without an implementation so it can be done at your own leisure.

carenas (changes requested):
better remive the non standard `(?t)`, it would seem that the extra ucd classes might not be needed if you are also overloading ucd_casesets

zherczeg:
This code adds many specializations, and I am still thinking how could we optimize it.

If I understand correctly, the Turkish casing removes the `i` == `I` caseless rule, that is what it makes it incompatible with other character sets. It also adds two new rules: 'i' == `U+0130` and `U+0131` == `I`. Is this correct?
`I` is 0x49 and `i` is 0x69.

Their choices for constants does makes it hard. Maybe some macros could help.

doc/html/pcre2api.html
carenas: to avoid the probability of conflict with a future Perl option, it was mentioned before that an uppercase letter would be preferred.

I am having a hard time seeing use cases for this though, so it would be wiser to hold this until those cases are presented IMHO.

PhilipHazel: One obvious use case is someone writing a regex who does not have access to the code, but who wants to do the Turkish thing. They can start the pattern with (*UTF)(?t) but I guess this is pretty rare. But who knows?

carenas: > They can start the pattern with (*UTF)(?t)

correct, but as you mentioned elsewhere for that use case it would be better if it is named `(*TURKISH_CASE)` or something similar which would have also the added benefit of not conflicting with Perl's future syntax so easily.

zherczeg: I just want to mention that nobody complained about turkish casing so far. I would wait until somebody needs it.

NWilson:
> If I understand correctly, the Turkish casing removes the i == I caseless rule, that is what it makes it incompatible with other character sets. It also adds two new rules: 'i' == U+0130 and U+0131 == I. Is this correct?
I is 0x49 and i is 0x69.

Yes, exactly correct.

The most-annoying thing is that even infects the ASCII fast-path processing, not just the slower UCD record access.

Thankfully, there aren't _loads_ of places where PCRE2 needs to worry about casing of characters. And some of them don't need to be updated - for example, we don't need change the idea of casing used by the "first code-unit" support.

What I think we want:
* OP_CHARI should just do "standard case-insensitivity" checking (according to UCD)
* During compiling, when we convert pattern characters to OPs, if we see something like an "i" character (and Turkish is set), or we see something affected by ASCII-caseless-restrict, then we can generate special opcodes like PT_CLIST that are totally explicit about what characters are matched.

So the Turkish (and ASCII-restrict) casing should really only affect whether characters are compiled to standard or special-cased opcodes (...and also the OP_REFI impact). That's not too bad, it's relatively restricted and clear.

I've largely only had to add Turkish pollution to chunks of code that already had ASCII-caseless-restrict pollution.

doc/html/pcre2api.html
zherczeg: We are talking about reserving a letter. The feature is ok. I hope we can even find ways to optimize it to reduce its overhead.

maint/GenerateUcd.py
zherczeg: The problem is, that each of these basic opcodes requires repeated variants, and that is already a lot of code cloning in the matcher. Implementing it is easy in jit, but we should not forget about the interpreter.

src/pcre2_compile.c
```diff
@@ -5736,8 +5739,10 @@ for (;; pptr++)
     /* ===================================================================*/
     /* Empty character classes are allowed if PCRE2_ALLOW_EMPTY_CLASS is set.
     Otherwise, an initial ']' is taken as a data character. When empty classes
-    are allowed, [] must always fail, so generate OP_FAIL, whereas [^] must
-    match any character, so generate OP_ALLANY. */
+    are allowed, [] must generate an empty class - we have no dedicated opcode
+    to optimise the representation, but it's a rare (and useless?) case. The
```
zherczeg, line 5743: I would say `(*F)` is a much better option to fail an alternative.

src/pcre2_compile.c
```diff
@@ -5808,18 +5810,34 @@ for (;; pptr++)
...
+        if ((xoptions & PCRE2_EXTRA_TURKISH_CASING) != 0 && UCD_ANY_I(c))
+          {
+          *code++ = OP_NOTPROP;
+          *code++ = PT_CLIST;
+          *code++ = UCD_DOTTED_I(meta)? PRIV(ucd_turkish_dotted_i_caseset) :
+              PRIV(ucd_turkish_dotless_i_caseset);
+          break;   /* We are finished with this class */
+          }
+
+        caseset = UCD_CASESET(c);
```
zherczeg, line 5831: What about setting `caseset` in a different way? Less code cloning.

src/pcre2_compile.c
```diff
@@ -7189,8 +7209,10 @@ for (;; pptr++)
       PUT2INC(code, 0, index);
       PUT2INC(code, 0, count);
       if ((options & PCRE2_CASELESS) != 0)
-        *code++ = ((xoptions & PCRE2_EXTRA_CASELESS_RESTRICT) != 0)?
-            REFI_FLAG_CASELESS_RESTRICT : 0;
+        *code++ = (((xoptions & PCRE2_EXTRA_CASELESS_RESTRICT) != 0)?
+            REFI_FLAG_CASELESS_RESTRICT : 0) |
+                  (((xoptions & PCRE2_EXTRA_TURKISH_CASING) != 0)?
+            REFI_FLAG_TURKISH_CASING : 0);
```
zherczeg, line 7215: I would increase the indentation of `REFI_FLAG_` lines. If the two options follow each other, then a `(xoptions >> log2(opt)) & 0x3` could work as well.

src/pcre2_compile.c
```diff
@@ -8343,12 +8367,26 @@ for (;; pptr++)
...
+        *code++ = OP_PROP;
+        *code++ = PT_CLIST;
+        *code++ = UCD_DOTTED_I(meta)? PRIV(ucd_turkish_dotted_i_caseset) :
+            PRIV(ucd_turkish_dotless_i_caseset);
+        if (firstcuflags == REQ_UNSET)
+          firstcuflags = zerofirstcuflags = REQ_NONE;
+        break;  /* End handling this meta item */
+        }
+
+      caseset = UCD_CASESET(meta);
```
zherczeg, line 8389: Same as above.

src/pcre2_compile_class.c
```diff
@@ -147,8 +148,23 @@ while (c <= end)
   /* Compute caseless set. */
   co = UCD_CASESET(c);
 
-  if (co != 0 && (!(options & PARSE_CLASS_RESTRICTED_UTF)
-                  || PRIV(ucd_caseless_sets)[co] > 127))
+  if ((options & PARSE_CLASS_TURKISH_UTF) && UCD_ANY_I(c))
+    {
+    list = tmp;
```
zherczeg, line 153: You could simply set the `co` variable. Less code duplication.

src/pcre2_match.c
zherczeg: We should be careful about adding this flag, since it is easy to forget about the tests later.

doc/html/pcre2api.html
NWilson: Sorry, I see, thank you!

If there isn't a letter reserved it should have `(*TURKISH)` and `(*NON_TURKISH)` as pre-pattern options.

I'll just await whatever decision you and Philip prefer. There's no "wrong" decision either way.

zherczeg (commented):
I just noticed something. It seems the turkish flag overwrites the restricted flag. Is this correct?

How the two flags should work together?

src/pcre2_compile.c
NWilson: I think you updated this code recently? I just noticed that the comment seemed not to match.

src/pcre2_compile_class.c
NWilson: Ah yes! That goes back to an earlier version of the code where I didn't have a caseset for the Turkish ones.

NWilson:
> I just noticed something. It seems the turkish flag overwrites the restricted flag. Is this correct?

No, it shouldn't do. They ought to work together fine.

CaseFolding.txt describes the case-equivalences between characters. CASELESS_RESTRICT unlinks the equivalences (from that file) which cross the ASCII boundary; it only affects the "k/Kelvin" and "s/Long-S" equivalences. TURKISH_CASING sets new equivalences for the 4 "i" characters.

I should add tests for them both at once. It looks like I forgot to do that.

maint/GenerateUcd.py
NWilson: It would be worth replacing PT_CLIST with "OP_2CHAR" in a future PR. It's not necessary right now though, just to get the Turkish feature.

doc/html/pcre2api.html
zherczeg: I would not mind if some options could be anywhere in a pattern, not just at the beginning. We have full assertion support now. `(*TURKISH)` would work like `(?t)` and `(*TURKISH:pattern)` like `(?t:pattern)`.

PhilipHazel: I have now forgotten the context, but some time ago there was a worry about the non-Perl things like (*UTF), but whoever it was that was worried was happy when they learned that these "specials" are recognized only at the start of a pattern as a sort of "pre-pattern". I would therefore confine (*TURKISH) and (*NON_TURKISH) in the same way. Perhaps that is best, just in case Perl invents (?t) at some point. I probably should have done (*CASELESS_RESTRICT) instead of (?r) but it is too late now.

src/pcre2_internal.h
```diff
@@ -1908,6 +1909,11 @@ typedef struct {
 #define UCD_SCRIPTX(ch)     UCD_SCRIPTX_PROP(GET_UCD(ch))
 #define UCD_BPROPS(ch)      UCD_BPROPS_PROP(GET_UCD(ch))
 #define UCD_BIDICLASS(ch)   UCD_BIDICLASS_PROP(GET_UCD(ch))
+#define UCD_ANY_I(ch) \
```
carenas, line 1912: needs also an EBCDIC version

http://www.kostis.net/charsets/ebc1026.htm

testdata/testinput5
```diff
@@ -2337,6 +2337,157 @@
...
+    \x{0130}
+    \x{0131}
+
+/\x{0131}/i,utf,turkish_casing
+    i
+    I
+    \x{0130}
+    \x{0131}
+
+/[i]/i,utf
```
carenas, line 2390: I think this will be optimized out into an OP_CHARI unless there are more characters in the class

testdata/testoutput5
```diff
@@ -5312,6 +5312,259 @@ No match
...
+    I
+ 0: I
+    \x{0130}
+No match
+    \x{0131}
+ 0: \x{131}
+
+/\x{0130}/i,utf
+    i
+No match
```
carenas, line 5359: Usually is better to sort them and make the ones expected to match at the end after something like "\= Expect no match", so it is harder to miss  bug like this one.

doc/html/pcre2api.html
NWilson: > I probably should have done (*CASELESS_RESTRICT) instead of (?r) but it is too late now.

Is it too late? It was added back in February, so I guess a fair number of users have already upgraded.

We could change it so that `(?r)` is recognised as a "pre-pattern" option only, with a big notice in the release notes, and add `(*CASELESS_RESTRICT)` as a new synonym. Then we can say `(?r)` is deprecated and in a future release it could be removed.

The reality is very few people, if any at all, are likely to be using it in the middle of a pattern, so this should be safe enough.

You've made breaking changes before, to align with Perl, but with flags to allow users to keep the old behaviour (eg `PCRE2_EXTRA_ALLOW_LOOKAROUND_BSK`). This would be similar - a small break but with simple workarounds for affected users.

It would very, very nice if both TURKISH_CASING and CASELESS_RESTRICT could be global options. Then all comparisons in the pattern would use the same case-folding rules. We wouldn't need the "OP_REFI" hack I added to propagate the casefolding flags; we wouldn't need OP_2CHAR (we'd have the option to just use ordinary OP_CHARI).

The more I think about it, the more I like the idea of having all casefolding be done consistently with one `UCD_CASEFOLD()` macro or similar, so there's no confusion.

src/pcre2_internal.h
NWilson: Philip, Zoltan, I don't think it does? Turkish casing only makes sense in UTF/UCP modes where we have characters U+0130 and U+0131.

All of the UCD_RECORD macros and data are using numeric constants, not character constants like `'A'`. The UCD macros must always assume that a `uint32_t ch` is a Unicode codepoint, not an EBCDIC one.

zherczeg: The question is compilation, not meaning. The code should be able to be compiled in EBCDIC systems.
Unfortunately we only have Ze'ev Atlas, who has access to EBCDIC, and he did not reply to my last mail :(

NWilson: This PR just uses numeric constants like `0x0130` and `0x49`, which are valid C code on any machine.

It will certainly _compile_ on an EBCDIC machine.

doc/html/pcre2api.html
carenas: > > I probably should have done (*CASELESS_RESTRICT) instead of (?r) but it is too late now.
> 
> Is it too late?

Yes, and unlike TURKISH_CASING, it wouldn't work, because it is not a full flag, but meant to be used together with (?i). That might had been arguably another mistake, but there is no point arguing about it since it is now part of the API.

Luckily we can avoid making a mistake with TURKISH_CASING.

NWilson:
Zoltan, using the "PT_CLIST" for the Turkish I characters causes an assertion failure in SLJIT:

> SLJIT_ASSERT(other_cases[0] != NOTACHAR && other_cases[1] != NOTACHAR && other_cases[2] != NOTACHAR);

Maybe now would be the time to introduce OP_2CHAR... although that would mean even more JIT work to implement it (but fairly simple at least, I hope).

I've also tried to make the change in `compile_ref_matchingpath()` but struggling with the JIT syntax.

maint/GenerateUcd.py
NWilson: I'll think about it. I'm still not sure we even want the Turkish characters in there, it's a temporary hack really.

doc/html/pcre2api.html
NWilson: I'm not convinced. Changing `(?r)` mid-pattern is so rare (and it's a new option this year, not an old one), we should at least consider removing that capability.

testdata/testinput5
NWilson: The intention is just to observe & test the behaviour of all those edge-cases and optimizations, and check they work correctly. I don't particularly mind what the compiled OPcodes are, as long it matches the correct characters.

testdata/testoutput5
NWilson: Thanks, good point.

doc/html/pcre2api.html
carenas: On the contrary; it is very likely that it is mostly used "in pattern" because of the way it interacts with caseless. If I use it as a global it does nothing by itself, but if I use PCRE2_CASELESS globally then in every section of the pattern where I want to make sure only non Unicode characters match, then I can do `(?r:..)`. It even makes the matching faster because now all those sections are compiled to OP_CHARI.

zherczeg:
Let me do more free work for microsoft:
https://github.com/zherczeg/pcre2/commit/090f194b877e289044790e8775f3ffa8d8159cf6

Btw, it would be good to test ucp (no utf) caseless turkish backreferences. Is there tests for turkish + restricted?

maint/GenerateUcd.py
carenas: it is just a slightly cleaner hack IMHO; and even if somehow `OP_CHAR2` materializes, doing something more data driven and with less hardcoding would be easier to adjust.

PhilipHazel:
Picking up on a couple of points:

As far as (?r) goes, we could introduce (*CASELESS_RESTRICT) without causing any incompatibility. At the same time, the use of (?r) could be deprecated for a release or two before removal. I agree with NIck (and therefore disagree with Carlo) that the need for this to vary in different parts of the pattern is minimal. _**However**_, I note that the nearest thing Perl has to (?r) is (?aa) which **_can_** be varied throughout the pattern. This output is from perltest.sh:

```
Perl v5.40.0
/k(?aa)k/i
\x{212a}k
 0: \x212ak
\x{212a}\x{212a}
No match
```
Perl's /aa is not the same as PCRE2's /r because /aa also makes \d etc into ASCII-only items. That must have been why I implemented (?aD) etc in PCRE2 as separate switches. Perhaps we should change (?r) into (?aR) and (?t) into (?aT), though I'm just throwing out ideas here. Oops, no, (?aT) is already taken. (?aI) is free.... This is all a mess.

As far as EBCDIC is concerned, the code should compile in two different ways in an EBCDIC environment: (a) dealing with 8-bit EBCDIC as input and (b) reading Unicode encoded as UTF-8 as input. This was requested for the processing of UTF-8 files in an EBCDIC environment. In case (a) there is clearly no need for ASCII/Unicode support. As I understand it, IBM's z/OS has a native environment, which is what Ze'ev Atlas works in, but also "Unix system services" which "allows UNIX applications from other platforms to run on [IBM System z](https://en.wikipedia.org/wiki/IBM_System_z) mainframes running z/OS" (Wikipedia). In theory, there is a small amount of testing that can be done by compiling PCRE2 --with-ebcdic on an ASCII system, but it is many years since I tried it.

carenas:
> (b) reading Unicode encoded as UTF-8 as input.

but there is no UTF-8 in EBCDIC systems, are you sure it is not [UTF-EBCDIC](https://www.unicode.org/reports/tr16/tr16-8.html) which is what Perl supports with `[use utf8](https://perldoc.perl.org/utf8)` there?

if what they were actually using is UTF-16, then the code is correct because it is compatible with  UTF-8, though and uses the same scalars for all 4 characters, but if it is the other UTF which seems to match better with [EBCDIC CCSID 1026](https://en.wikibooks.org/wiki/Character_Encodings/Code_Tables/EBCDIC/EBCDIC_1026) then at least U+69 will be encoded as x89 and U+131 (AKA dotless i) would be in x79 (which are still 1 byte encodings).

NWilson:
You certainly can have UTF-8 files on an EBCDIC system. There's nothing magic about an EBCDIC system that stops you storing any binary on it. I think in UTF mode on an EBCDIC system, PCRE2 appears to just do... use normal UTF.

NWilson:
I have:
* Removed the `(?t)` inline option, and updated the documentation
* Added a start-of-pattern option `(*TURKISH_CASING)`
* Pulled in Zoltan's commit (many, many thanks!)
* Added some additional tests

There is one test case that's failing for the JIT, which I'll have a go fixing in the morning tomorrow.

Otherwise, I think this is basically good to go.

testdata/testinput5
```diff
@@ -2337,6 +2337,217 @@
...
+    i I
+
+/(.) \1/i,ucp
+    i I
+
+/(.) \1/i,ucp,turkish_casing
+\= Expect no match
+    i I
+
+/[a-z][^i]I/utf,caseless_restrict,turkish_casing
```
zherczeg, line 2520: Do we have caseless_restrict,turkish_casing with backreferences?

src/pcre2_jit_compile.c
```diff
@@ -9488,6 +9493,21 @@ if (common->utf && (*cc == OP_REFI || *cc == OP_DNREFI))
...
+  if (refi_flag & REFI_FLAG_TURKISH_CASING)
+    {
+    OP2(SLJIT_OR, SLJIT_TMP_DEST_REG, 0, char1_reg, 0, SLJIT_IMM, 0x20);
+    turkish_ascii_i = CMP(SLJIT_EQUAL, SLJIT_TMP_DEST_REG, 0, SLJIT_IMM, 0x69);
+
+    if (refi_flag & REFI_FLAG_CASELESS_RESTRICT)
+      add_jump(compiler, &no_match, turkish_ascii_i);
+
+    OP2(SLJIT_OR, SLJIT_TMP_DEST_REG, 0, char1_reg, 0, SLJIT_IMM, 0x1);
+    turkish_non_ascii_i = CMP(SLJIT_EQUAL, SLJIT_TMP_DEST_REG, 0, SLJIT_IMM, 0x131);
```
zherczeg, line 9505: I just remember that I read somewhere that 0x130/0x131 has no othercase by default. If this is true, then we don't need to do anything when REFI_FLAG_CASELESS_RESTRICT is set, and the second part can be simplified to:

```
if (!(refi_flag & REFI_FLAG_CASELESS_RESTRICT))
  {
  OP2(SLJIT_OR, SLJIT_TMP_DEST_REG, 0, char1_reg, 0, SLJIT_IMM, 0x1);
  turkish_non_ascii_i = CMP(SLJIT_EQUAL, SLJIT_TMP_DEST_REG, 0, SLJIT_IMM, 0x131);
  }
```

testdata/testinput5
carenas: There seem to be some blind spots there, added more tests in fad3d34b118bd91b214dd001e49c110ad17fd0bb that sadly also show that turkish_casing still has the bug that reported earlier.

```
  re> /[^i]/Bi,utf,turkish_casing
        Bra
        not clist 0049 0131
        Ket
        End
data> i
 0: i
```

NWilson: !!! Thank you Carlo, I've added your test - this case was covered by an existing test, but I'd got the test results flipped!

There was a bug. I had `caseset = UCD_DOTTED_I(meta)? ...` instead of `caseset = UCD_DOTTED_I(c)? ...` due to a copy-paste bug.

Fixed, thank you again.

src/pcre2_jit_compile.c
NWilson: I've tried to debug your code Zoltan, but I can't work it out.

It's clearly not doing the same thing as the C in pcre2_match.c's `match_ref()`, but I don't understand why you changed the logic.

I've spent a couple of hours going through the sljit tutorial, and its header file. I'll keep trying...

maint/GenerateUcd.py
zherczeg: I hope a 2CHAR could also replace CHARI. Should be faster than reading UCD records. The CHARI could be limited to 8 bit only.

testdata/testinput5
NWilson: ?? I do not get that...

src/pcre2_jit_compile.c
NWilson: The basic logic I can follow, and the overall structure seems correct & sensible.

But there's something about the "jump" instructions that isn't working the way I expect, or something like that?

testdata/testinput5
NWilson: All the tests are passing in the NoJIT build, and in the JIT build, there's just a specific test with backrefs that's failing that I'm struggling to fix.

src/pcre2_compile.c
```diff
@@ -10813,7 +10856,7 @@ re->blocksize = re_blocksize;
 re->magic_number = MAGIC_NUMBER;
 re->compile_options = options;
 re->overall_options = cb.external_options;
-re->extra_options = ccontext->extra_options;
+re->extra_options = xoptions;
```
zherczeg, line 10859: Cannot we simply modify this flag directly?

src/pcre2_compile.c
```diff
@@ -2835,8 +2838,8 @@ be quantified. */
 
 /* Here's the actual function. */
 
-static int parse_regex(PCRE2_SPTR ptr, uint32_t options, BOOL *has_lookbehind,
-  compile_block *cb)
+static int parse_regex(PCRE2_SPTR ptr, uint32_t options, uint32_t xoptions,
```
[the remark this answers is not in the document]
NWilson: We removed `(?t)` and replaced with `(*TURKISH_CASING)`, so the xoptions can't be read from the compile-context anymore (since they can now be changed with a start-of-pattern option).

src/pcre2_internal.h
```diff
@@ -1908,6 +1911,14 @@ typedef struct {
...
+  /* match any of the four characters 'i', 'I', U+0130, U+0131 */ \
+  (((uint32_t)(ch) | 0x20u) == 0x69u || ((uint32_t)(ch) | 1u) == 0x0131u)
+#define UCD_DOTTED_I(ch) \
+  ((uint32_t)(ch) == 0x69u || (uint32_t)(ch) == 0x0130u)
+#define UCD_FOLD_I_TURKISH(ch) \
+  ((uint32_t)(ch) == 0x0130u ?   0x69u : \
+   (uint32_t)(ch) ==   0x49u ? 0x0131u : (uint32_t)(ch))
```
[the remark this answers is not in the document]
NWilson: There are only two characters that fold, the two uppercase ones. What other case-fold mappings did you expect?

src/pcre2_compile.c
NWilson: No, the pattern can't modify the compile-context, which could be reused with another pattern.

doc/html/pcre2unicode.html
```diff
@@ -157,6 +157,42 @@ <h1>pcre2unicode man page</h1>
...
+<P>
+It is allowed to specify both PCRE2_EXTRA_CASELESS_RESTRICT and
+PCRE2_EXTRA_TURKISH_CASING together. The behaviour is to apply
+PCRE2_EXTRA_CASELESS_RESTRICT first, removing case-folding mappings which
+cross the ASCII boundary; and then to apply PCRE2_EXTRA_TURKISH_CASING to add
+the Turkish case-folding mappings for 'i'. The result is that there are two
+case-folding mappings crossing the ASCII boundary (the Turkish 'i' ones, as
+requested), but the other mappings (Kelvin sign and long S) are removed.
+If you require exact ASCII matching of ASCII characters in patterns, then
+you must use PCRE2_EXTRA_CASELESS_RESTRICT without PCRE2_EXTRA_TURKISH_CASING.
```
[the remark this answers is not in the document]
NWilson: You and Carlo both think that it's confusion. So maybe we could just forbid them from being used together.

That's not easy though... since the `(?r)` flag can be set mid-pattern!

If you use Turkish, you really do need the case mappings that cross the ASCII boundary. It would be broken to have some half-way behaviour, that disassociates the 'i' and 'I' characters, but doesn't add the association between those characters and U+130, U+131.

zherczeg:
It looks like my incorrect suggestion was causing the failure:

```
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -9525,12 +9525,13 @@ if ((common->utf || common->ucp) && (*cc == OP_REFI || *cc == OP_DNREFI))
   OP2(SLJIT_ADD, TMP1, 0, TMP1, 0, TMP3, 0);
   CMPTO(SLJIT_EQUAL, TMP1, 0, char1_reg, 0, loop);

-  if (refi_flag & REFI_FLAG_CASELESS_RESTRICT)
-    add_jump(compiler, &no_match, CMP(SLJIT_LESS, char1_reg, 0, SLJIT_IMM, 128));
   add_jump(compiler, &no_match, CMP(SLJIT_EQUAL, TMP2, 0, SLJIT_IMM, 0));
   OP2(SLJIT_SHL, TMP2, 0, TMP2, 0, SLJIT_IMM, 2);
   OP2(SLJIT_ADD, TMP2, 0, TMP2, 0, SLJIT_IMM, (sljit_sw)PRIV(ucd_caseless_sets));

+  if (refi_flag & REFI_FLAG_CASELESS_RESTRICT)
+    add_jump(compiler, &no_match, CMP(SLJIT_LESS | SLJIT_32, SLJIT_MEM1(TMP2), 0, SLJIT_IMM, 128));
+
   caseless_loop = LABEL();
   OP1(SLJIT_MOV_U32, TMP1, 0, SLJIT_MEM1(TMP2), 0);
   OP2(SLJIT_ADD, TMP2, 0, TMP2, 0, SLJIT_IMM, sizeof(uint32_t));
```

The first character of the caseset needs to be checked for ascii, not the current character. The test was good to reveal it! Thank you!

src/pcre2_internal.h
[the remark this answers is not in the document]
zherczeg: All 4 including the lowercase ones. So this is only used when we know the character is an uppercase letter?

doc/html/pcre2unicode.html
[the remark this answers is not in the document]
zherczeg: I support the mutual exclusivity. I think it was even suggested before. If somebody really needs it in the future, we can think about that. And hear his use case first...

src/pcre2_jit_compile.c
NWilson: This is the test that's failing:

```
pcre2/build-linux$ ./pcre2test -16
PCRE2 version 10.45-DEV 2024-06-09 (16-bit)
  re> /(.) \1/ir,utf,ucp,jit
data> \x{212a} k
 0: \x{212a} k
 1: \x{212a}
data>
```

It's nothing to do with Turkish! There's just something broken in the caseless-restrict logic in the JIT.

zherczeg: Yes. See the last comment.

doc/html/pcre2unicode.html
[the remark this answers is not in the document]
carenas: > If you use Turkish, you really do need the case mappings that cross the ASCII boundary.

Since we are going to produce an error, we should also (at least for now) trigger it when the utf/ucp flags aren't also provided, because `turkish_caseless` ONLY knows how to make that association in that case.

if an 8-bit codepage is in use, is also not clear what `turkish_caseless` should do.

src/pcre2_internal.h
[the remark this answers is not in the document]
NWilson: No, it's used to fold ("lowercase") any of the 4 characters. It's not as optimised as the JIT code, but it follows the canonical pattern of "c1 case-equivalent-to c2 means fold(c1) == fold(c2)".

doc/html/pcre2unicode.html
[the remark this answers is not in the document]
NWilson: What should be the behaviour be if you do `/...(?r: foo ).../i,turkish_casing`? Should it deactivate the turkish casing, when ASCII-restrict is activated, so i and I become equal again? Or just "break" the turkish casing, leaving all 4 of the 'i' characters not equal to each other?

They're both nasty.

NWilson:
I have updated the CASELESS_RESTRICT and TURKISH_CASING behaviour, as requested.
* They generate a compile error used together (if passed as options, or as start-of-pattern options)
* They both generate a compile error if used when neither UTF or UCP is enabled
* TURKISH_CASING generates a compile error in the 8-bit library if used without UTF

Finally, these checks are not triggered if CASELESS_RESTRICT is set mid-pattern. In this case, if both CASELESS_RESTRICT and TURKISH_CASING are set, then Turkish is totally deactivated as long as CASELESS_RESTRICT is in force.

src/pcre2_compile.c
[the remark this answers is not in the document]
zherczeg: I mean we should revert the change then.

src/pcre2_ucd.c
```diff
@@ -142,14 +142,22 @@ const uint32_t PRIV(ucd_caseless_sets)[] = {
...
   0x1c88,  0xa64a,  0xa64b,  NOTACHAR,
+  0x0069,  0x0130,  NOTACHAR,
+  0x0049,  0x0131,  NOTACHAR,
 };
 
+/* These are the indices, within ucd_caseless_sets, of the additional
+Turkish case-equivalences. */
+
+const uint32_t PRIV(ucd_turkish_dotted_i_caseset) = 112;
+const uint32_t PRIV(ucd_turkish_dotless_i_caseset) = 115;
```
zherczeg, line 153: Is it worth to have two variables for this? A +3 could be enough, since they are together.

zherczeg (commented):
I like this code much better.

src/pcre2_ucd.c
NWilson: Carlo would probably complain if he saw a hardcoded +3 :)

zherczeg: You can use a `#define`

NWilson:
Tests updated. Surely we're done now...? I hope!

doc/html/pcre2unicode.html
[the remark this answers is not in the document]
PhilipHazel: I don't mind as long as the behaviour is clearly documented. The /...(?r:foo).../i,turkish_casing example is a plausible scenario for a pattern that is looking for something ascii in the middle of some Turkish, in which case turning off turkish casing does make sense. I think this is, in fact, what Perl does. With a hacked version of perltest.sh (I will tidy this up and commit in due course) I get this:

```
Locale: tr_TR.utf8
Perl v5.40.0
/\x{130}/gi,utf
    \x{130}\x{131}iI
 0: \x{130}
 0: i

/\x{131}/gi,utf
    \x{130}\x{131}iI
 0: \x{131}
 0: I

/(?aa)\x{130}/gi,utf
    \x{130}\x{131}iI
 0: \x{130}

/(?aa)\x{131}/gi,utf
    \x{130}\x{131}iI
 0: \x{131}

```
In other words, U+0130 matches i and U+0131 matches I in Turkish \i mode, but if ascii is forced, they don't.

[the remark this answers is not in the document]
carenas: I prefer to return a compiler error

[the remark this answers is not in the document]
PhilipHazel: A compiler error would not be compatible with Perl, and there is a plausible use case for what Perl does.

src/pcre2_ucd.c
carenas: Ironically, I think it is less ugly if hardcoded all the way like this, but sure hope we can get `OP_2CHAR` done soon enough to remove this ASAP

carenas:
> Surely we're done now...?

I think the extra restrictions put up on (?r) were unnecessarily harsh so have them reverted in a couple of followup commits in [my fork of this](https://github.com/carenas/pcre2/tree/turkish-i).

we are missing documentation on the interaction between `(?:r)` and `PCRE2_EXTRA_TURKISH_CASING`, but the only place were such documentation could be added in `pcre2pattern.3` is already convoluted enough without my help so I gave up on a patch.

rebasing the while thing on top of HEAD and squashing all commits with a nice descriptive commit message also a missing.

NWilson:
OK, thanks Carlo.

I don't really see that we need to support `CASELESS_RESTRICT` when neither UTF nor UCP is set - it does absolutely nothing in this case - but if you want it, fine.

I've accepted your two commits, and rebased & squashed.

zherczeg:
This patch looks good, maybe we should land it, and discuss / fix the remaining issues later.

PhilipHazel:
> I don't reallyI don't really see that we need to support CASELESS_RESTRICT when neither UTF nor UCP is set - it does absolutely nothing in this case - but if you want it, fine.

I think the use case is when the application sets CASELESS_RESTRICT and the creator of the pattern uses (*UTF) or (*UCP).

I will do the merge and have a play.

PhilipHazel:
Does anybody understand why the scorecards action is now failing?

zherczeg:
Online services are not available. It says:

429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit.
https://github.com/PCRE2Project/pcre2/actions/runs/11331094692/job/31510350495

Ignore this message.

PhilipHazel:
I will ignore the message, but we now have a red cross on the Actions tab. I am not sure who set up this action, or exactly what it does.

zherczeg:
Me neither. It says it is about security risks:

https://github.com/PCRE2Project/pcre2/blame/master/.github/workflows/scorecards.yml
https://github.com/PCRE2Project/pcre2/commit/77ce1ff528edd1aac31772cbee1f19c069db32c7
https://openssf.org/projects/scorecard/

carenas:
It will probably work again in the next merge, it was just rate limited this time.

PR #523, opened 2024-10-09
Implement the new set-based character classes

NWilson:
This is a mostly-complete implementation!

It needs some tidy-up, in places where I've left "XXX" comments. But I've hammered out the code and I'll start testing it.

I'm afraid I used recursive descent in the operator-precedence parser, but it only goes down to a maximum depth of 15 levels, so it should be OK. I just wanted to prototype quickly. I don't know "how much stack is too much" in PCRE2 - it's only pushing ~hundreds of bytes on the stack at most.

The two big bits of logic that need to be shared with existing code, are the OP_XCLASS interpreter code, and the compiler code to build the OP_CLASS/NCLASS/XCLASS code. The current code is inlined into the one place where it's used.

Anyway, this is a draft of where it's going.


zherczeg:
Pattern (bracket) parsing is already recursive. Although you can also a use a simple array for simulating recursion. I suspect a pointer in the META code is enough to track the status.

PhilipHazel:
I do know that there are applications running PCRE2 in a large number of threads, each of which has very limited stack, but I don't know how limited. One would hope a couple of hundred bytes would be OK. I suppose we do need a new option to turn this on because changes the meaning of things, but it seems a shame to use one of the only two remaining main options bits. If/when the Perl syntax is implemented, it shouldn't need to be enabled. 

I got a compile error when trying to build your branch:

```
src/pcre2_compile.c:6110:28: error: passing argument 1 of 'compile_class_nested' from incompatible pointer type [-Wincompatible-pointer-types]
 6110 |       compile_class_nested(&pptr, &code);
      |                            ^~~~~
      |                            |
      |                            uint32_t ** {aka unsigned int **}
src/pcre2_compile.c:2669:39: note: expected 'const uint32_t **' {aka 'const unsigned int **'} but argument is of type 'uint32_t **' {aka 'unsigned int **'}
 2669 | compile_class_nested(const uint32_t **pptr, PCRE2_UCHAR **pcode)
```
I fixed this by inserting the obvious cast, but is that right? Haven't done any testing yet. I see the CI tests are failing; probably the same thing.

NWilson:
I'm expecting that the Perl syntax will be able to reuse the same interpreter support and OPcodes.

I won't tackle implementing the parser & compiler for it in this PR though, there's enough in it already. I agree it should be OK to add in a follow-on PR, and it wouldn't need any flags.

I imagined that it might be worth spending a "normal" parse bit on this flag? This flag implements a behaviour that's in JavaScript/Python/Rust and other engines, so it's more mainstream than all the other minor things I've recently added to the extra options. But you're quite possibly right, those bits are too precious now to spend on non-Perl-related syntax extensions.

PhilipHazel:
How ready is this for testing? I added the obvious to pcre2test to set the flag, but my first test failed:

```
PCRPCRE2 version 10.45-DEV 2024-06-09 (8-bit)
/[[a-c] && [b-d]]/set_class
Failed: error 189 at offset 16: internal error: unknown code in parsed pattern
```
This happens at line 6441 in pcre2_compile.c.

PhilipHazel:
Oops sorry, I told a lie. It's around line 8682 (didn't look for multiple occurrences of this error).

PhilipHazel:
Thinking about the name PCRE2_SET_CLASS ... I assume that some time in the future, the Perl syntax will always be recognized, so what this option is enabling is an alternative syntax for what Perl calls "extended bracketed character classes". So perhaps a name like PCRE2_ALT_EXTENDED_CLASS is more descriptive. This is the same style as other PCRE2_ALT_xxx options.

NWilson:
> How ready is this for testing? I added the obvious to pcre2test to set the flag, but my first test failed:

Not quite ready. It's just for getting feedback, if anyone's interested in looking (thanks for trying!).

I'm happy to rename from "set[-based] class" to "extended class". That will be easy enough to do.

src/pcre2_match.c
```diff
@@ -1929,6 +1929,104 @@ fprintf(stderr, "++ %2ld op=%3d %s\n", Fecode - mb->start_code, *Fecode,
...
+        switch (*Fecode)
+          {
+          case OP_ECLASS_OR:
+          ++Fecode;
+          stack = (stack >> 1) | (stack & 1u);
+          break;
+
+          case OP_ECLASS_AND:
+          ++Fecode;
+          stack = (stack >> 1) & (stack & 1u);
```
zherczeg, line 1980: Note: this clears all upper bits of the stack. `(stack >> 1) & (stack | ~(uint32_t)1u)` is the correct form. Sub is also a similar case.

src/pcre2_compile.c
```diff
@@ -2493,6 +2500,224 @@ return -1;
...
+              META_CLASS_END itself.
+            FALSE otherwise. In this case, *pendptr is set to the position
+              of the META code that caused the function to return FALSE.
+
+In both cases, the span ptr...*pendptr is the longest sequence of simple
+codes starting at ptr.
+*/
+
+static BOOL
+check_simple_class(uint32_t *ptr, uint32_t **pendptr)
```
zherczeg, line 2528: Many of these class operations should go to `pcre2_compile_class.c`, it was created for this purpose.

src/pcre2_match.c
NWilson: Ouch, that was embarrassing, thank you.

I've started doing more testing, and fixed a few silly bugs.

src/pcre2_compile.c
NWilson: Great - are you pausing your work on character classes for a few days? I'll move the code into there, but it would be nice to wait until you're done with any major changes, or else do it quickly while things are quiet.

I can cope with some conflicts, just ideally not massive ones.

NWilson:
I've added docs.

I'm down to these tasks:
- [ ] JIT support - is that needed for this PR, or can it come in a future one?
- [ ] Need a lot more tests, I'm currently writing them. Will be done in a day or two
- [ ] One chunk of code (marked by a big "XXX") needs to be pulled out into a separate function.
- [ ] I'll move all the code into pcre2_compile_class.c as requested

src/pcre2_compile.c
zherczeg: Currently I have no plans to work on the compiler part. Maybe some jit optimizations.

zherczeg:
Some random comments.

I don't see a negate operation (can be implemented as `^ 0x1`). For example, it is needed for `[^\PL-\x{100}-\x{200}]`.

There is no point to duplicate bitsets (for character 0..255). If one xclass has a bitset, the full bitset should be computed for the entire class. It is faster to evaluate and requires less space.

The parser can be implemented with using a simple stack in the parser context (e.g. 32 uint32_t values, which contains the offsets of the currently active operators). This way the recursion does not require any recursion.

There should be tests for reaching the maximum allowed depth. They should throw parse errors.

Is `[&&]` allowed? Is it simply matches to nothing? Does `[[]&&[]]` depends on allow empty classes flag? Btw how "allow empty classes" affects the parsing?

There should be optimizations for `||`. `[[[\pS]||]||[[[\pL]]]]` is just `[\pS\pL]`. This is easy to do during parsing.

Personally I would just implement `[]` and `||` at the first stage, do all the optimizations, parse errors, discuss all the minor details. This only requires parser modifications, so you don't need to worry about matching.

NWilson:
Thank you @zherczeg for reviewing!

> I don't see a negate operation (can be implemented as `^ 0x1`). For example, it is needed for `[^\PL-\x{100}-\x{200}]`.

It was there in the PR, it's OP_ECLASS_NOT. There isn't a META for it, because it's encoded as META_CLASS / META_CLASS_NOT distinction. But I had implemented the interpreter support.

> There is no point to duplicate bitsets (for character 0..255). If one xclass has a bitset, the full bitset should be computed for the entire class. It is faster to evaluate and requires less space.

True. I haven't done this optimisation.

> The parser can be implemented with using a simple stack in the parser context (e.g. 32 uint32_t values, which contains the offsets of the currently active operators). This way the recursion does not require any recursion.

Hmm. Can it really? We have a grammar with two operators, with different precedence levels. I can code up a simple shunting-yard thing... but it would be more code than recursive descent.

I can check how much stack space it _actually_ uses. Currently it looks like it chews up around 100 bytes per recursion, in a Release build, so that comes out to a couple of KiB of stack space consumed. The shunting-yarn approach would cut that down a bit, but I'm really not sure it's worth it.

> There should be tests for reaching the maximum allowed depth. They should throw parse errors.

Added. Lots more tests added.

> Is `[&&]` allowed? Is it simply matches to nothing? Does `[[]&&[]]` depends on allow empty classes flag? Btw how "allow empty classes" affects the parsing?

Good question. `[&&]` is not allowed: an operator must have something non-empty on either side.

The `[[]&&[]]` case does indeed depend on allow_empty_class. That flag affects not just the outermost `[...]`, but all of the nested classes opened by a '[' character. The allow_empty_class flag affects the inner nested classes in the same way as an outermost class: if the first literal is a ']', then the flag determines whether it's a literal or a class close ']'.

> There should be optimizations for `||`. `[[[\pS]||]||[[[\pL]]]]` is just `[\pS\pL]`. This is easy to do during parsing.

That would be nice... but I'm so far totally focussed on completeness and correctness. I'd rather get the PR merged, than hold it up for edge-case optimisations.

> Personally I would just implement `[]` and `||` at the first stage, do all the optimizations, parse errors, discuss all the minor details. This only requires parser modifications, so you don't need to worry about matching.

Even with just `||`, you need changes to the matcher: `[a[^b]]` is pretty simple, but requires runtime op support.

NWilson:
Updated task list:

- Could add some more tests, there's a handful of missing cases. But it's pretty thoroughly exercised already.
- Still no JIT support. Unsure whether I need to do this to get it merged.
- And certainly no `(?[...])` support. That can come in a follow-up PR, soon after

NWilson:
Also - regarding Zoltan's example of `[[[\pS]||]||[[[\pL]]]]`. The `[thing||]` syntax is not legal. And the extra brackets `[[[thing]]]` syntax is already optimised away (compiles to the same OPs as `[thing]` automatically).

PhilipHazel:
> Still no JIT support. Unsure whether I need to do this to get it merged.

Merging without JIT support should be possible. Because you have added new opcodes, a call to pcre2_jit_compile() will fail, and matching should fall back to the interpreter. There are still, I think, one or two other constructions that JIT does not support.

NWilson:
Well, as I said, "I can code up a simple shunting-yard thing... but it would be more code than recursive descent."

You are describing the classic shunting-yard thing... and it really is more code, and more complexity.

At the moment, I have implemented everything (`^`, `||`, `&&`, and `--`), and it's both robust, and correct. So I don't see it as a bad foundation at all. I _could_ rewrite the second parsing phase, but in my opinion the only reason to do so would be to save a tiny bit of stack space.

I have implemented most of the optimisations you describe. I haven't implemented merging `[[a-f][i-k]]` into a single class, which would be a bit fiddly, but I guess could be possible.

Note also that the `||` is very unlikely to be used in practice. It's simply there for completeness. JavaScript chose not to add it, but Rust and Python `regex` module did. I don't think it's worth doing any special optimisations for it.

> Btw [a&&bc] is [a&&[b]c] or [a&&[bc]]?

It's `[a&&[bc]]`. The `^` character is not an operator, it's a modifier on nested classes to make them into "yes-classes" or "not-classes". The implicit-union operator (juxtaposition) has tightest precedence (implicit union is any juxtaposition of atoms like single characters or ranges, and also nested classes; for example, `[[a-b][^c-e]]`). The explicit operators `&& || --` have looser precedence (and undefined precedence relative to each other - you aren't allowed to mix them).

---

I'm not at all opposed to changing the parser. I just want to be clear on the goal of the improvement.

PhilipHazel:
If it's all implemented and working, I think we (that is, you :-) should proceed and get on with what is left to be done. Internal algorithms can always be changed later if that proves to be beneficial.

testdata/testoutput2
```diff
@@ -19612,6 +19612,693 @@ Failed: error -49 at offset 3 in replacement: unknown substring
...
+No match
+
+/[^[A][B]]/B,alt_extended_class
+------------------------------------------------------------------
+        Bra
+        eclass[
+          cls:[A]
+          cls:[B]
+          op: ||
+          op: ^
```
zherczeg, line 19715: Even if ^ is a modifier, it is an operator in practice. Btw is `[^[^a-z]]` optimized to `[a-z]`?

testdata/testoutput2
```diff
@@ -19612,6 +19612,693 @@ Failed: error -49 at offset 3 in replacement: unknown substring
...
+    A
+ 0: A
+    C
+ 0: C
+    [
+ 0: [
+    ]
+ 0: ]
+
+/[A-C--B]/B,alt_extended_class
```
zherczeg, line 19826: Is there tests for `[A--B--C--D]` `[A--^B]` `[A---B]` `[A----B]` `[^--A]`.
Is there tests to reach the maximum depth?

testdata/testoutput2
```diff
@@ -19612,6 +19612,693 @@ Failed: error -49 at offset 3 in replacement: unknown substring
...
+    b
+No match
+
+/[[a]-&&-[a]]/B,alt_extended_class
+------------------------------------------------------------------
+        Bra
+        eclass[
+          cls:[a]
+          cls:[\-]
+          op: ||
```
zherczeg, line 20074: We should optimize these before the release.

testdata/testoutput2
```diff
@@ -19612,6 +19612,693 @@ Failed: error -49 at offset 3 in replacement: unknown substring
...
+Failed: error 210 at offset 5: expected operand after operator in character class
+
+/[--a]/alt_extended_class
+Failed: error 209 at offset 3: unexpected operator in character class (no preceding operand)
+
+/[--]/alt_extended_class
+Failed: error 209 at offset 3: unexpected operator in character class (no preceding operand)
+
+/[a---b]/alt_extended_class
+Failed: error 208 at offset 5: invalid operator in character class
```
zherczeg, line 20257: Should be `[a--[-b]]`

testdata/testoutput2
```diff
@@ -19612,6 +19612,693 @@ Failed: error -49 at offset 3 in replacement: unknown substring
...
+
+/[a|||b]/alt_extended_class
+Failed: error 208 at offset 5: invalid operator in character class
+
+/[a||b&&c]/alt_extended_class
+Failed: error 211 at offset 7: brackets needed to clarify operator precedence in character class
+
+/[a--b&&c]/alt_extended_class
+Failed: error 211 at offset 7: brackets needed to clarify operator precedence in character class
```
zherczeg, line 20269: Why?

zherczeg (commented):
Unfortunately a lots of choices is confusing for me here. For me these should work like a usual expression evaluator, with unary/binary operators and precedence. The primary expressions should be boolean values, [] can be used for grouping.

PhilipHazel:
I see there is still one failing CI test.

NWilson:
> For me these should work like a usual expression evaluator, with unary/binary operators and precedence. The primary expressions should be boolean values, [] can be used for grouping.

That is exactly how it works, I think. I didn't do anything "unusual" on purpose. There are primary expressions (atoms) which are characters/ranges/POSIX-props/Unicode-props, and there are binary operators (no unary), and [] can be used for grouping. Just as you say.

Very classic recursive descent with precedence, producing a RPN sequence of operators.

> I see there is still one failing CI test.

Yes, it's weird. Moving some code around triggers the `Werror=maybe-uninitialized` warning in alpine's version of GCC, although I didn't really touch that code. It's full of gotos; I guess the detector is a bit sensitive.

testdata/testoutput2
NWilson: Ah I see. In the grammar, it's not an operator, because it's attached to the `[^`, so it's parsed as a special form of "open paren". But in the evaluator, it is an operator. The close-paren "]" emits a final OP_ECLASS_NOT if it was opened with a "[^".

> Btw is [^[^a-z]] optimized to [a-z]?

Not yet, no. I _will_ do optimisations (although these are pathological cases you suggest...). If possible I'll do it in a PR next week, although I can include it in the current PR if you require.

testdata/testoutput2
NWilson: I have most of those tests.

The expression `[A--^B]` has a _literal_ ASCII `^` character, and so it's equivalent to simply `[A]`.

The expression `[A---B]` is a parse error. So are `[A----B]` `[^--A]`.

I have tests to reach maximum depth - a successful test at max depth, and a parse error test at max+1.

testdata/testoutput2
NWilson: Why? We're following UTS#18 syntax here, not inventing our own. Other implementations forbid tripled operator characters.

A single `-` hyphen is either a literal or a range separator, with complex disambiguation. A double hyphen `--` is always an operator (never one end of a range: `[--a]` used to mean "characters from a to hyphen" but in UTS#18 syntax it's invalid and you must instead use `[\--a]`). A triple hyphen is just invalid.

testdata/testoutput2
NWilson: Good question. I address this in the manpage updates included in the PR.

There's also the (overly?) detailed preliminary survey I did of other implementations of UTS#18 syntax, which I linked in an email thread: https://github.com/NWilson/pcre2/blob/user/niwilson/utr18/NEW-CLASSES.md

* UTS#18 doesn't define the relative precedence of the operators
* Some implementations ban mixing operators (ECMAScript spec requires this)
* Some implementations give them all the same precedence... but some implementations give them a different precedence

So it's a compatibility nightmare, what a mess. I don't think we can justify picking a "side" in the battle. It would be really unfortunate if we picked some precedence for the operators, and then they updated UTS#18 with a different choice.

testdata/testoutput2
zherczeg: There is no tripled operator here. `[#--]` and `[--a]` are currently valid.

It looks like unicode recommendation lists several chars, which cannot be in a pattern without `\`. Do you have tests for them?

testdata/testoutput2
zherczeg: There are no requirements. Since I prefer to have these optimizations, I should do them.

testdata/testoutput2
zherczeg: This choice definitely simplify the current code. However, perl supports multiple operators, so the code needs to be able to handle it. This might be another thing I need to implement.

NWilson: I'll be implementing the Perl syntax, don't worry. It will drop into the current parser very easily.

testdata/testoutput2
NWilson: I really will do them, I'd just rather split into a PR next week, as an extra piece of work.

Just give me a few days to get onto it. I'm away this weekend.

src/pcre2_compile.h
```diff
@@ -160,6 +161,10 @@ versions. */
 #define META_FIRST_QUANTIFIER META_ASTERISK
 #define META_LAST_QUANTIFIER  META_MINMAX_QUERY
 
+#define META_ECLASS_OR        0x80440000u  /* || in a class */
+#define META_ECLASS_AND       0x80450000u  /* && in a class */
+#define META_ECLASS_SUB       0x80460000u  /* -- in a class */
```
carenas, line 166: this should go after the last META (ex: META_MINMAX_QUERY), so move up before the convenience aliases (META_FIRST_QUANTIFIER) and special cases.

src/pcre2_extuni.c
```diff
@@ -40,7 +40,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 /* This module contains an internal function that is used to match a Unicode
 extended grapheme sequence. It is used by both pcre2_match() and
-pcre2_def_match(). However, it is called only when Unicode support is being
+pcre2_dfa_match(). However, it is called only when Unicode support is being
```
carenas, line 43: is this (the need for the dummy when unicode support is not compiled that will be never called) still true?, it is confusing but it would seem that ECLASS calls this unconditionally.

PhilipHazel:
Re my previous comment: under DFA matching it *does* match "acac".

src/pcre2_jit_compile.c
```diff
@@ -1271,6 +1270,9 @@ while (cc < ccend)
     cc++;
     break;
```
carenas, line 1272: I think this also needs on top (for safety):

```c
#if !defined(SUPPORT_UNICODE) || PCRE2_CODE_UNIT_WIDTH == 8
case OP_XCLASS:
#endif
```

PhilipHazel:
Another example where DFA does what I expect, but non-DFA doesn't. May be same bug as before.

```
/[\pL || [:digit:]]+/alt_extended_class
    +a03+
No match
    +a03+\=dfa
 0: a03
```

NWilson:
Yes, it is a bug. I found a similar case earlier this afternoon while adding more tests of my own.

Very odd, something do with the fact that I changed OP_ECLASS to take part in auto-possessification, in the same way that OP_XCLASS does. I introduced the regression at that point.

It turns out to be a small one-line copy-and-paste error, when I copied a bit of code from XCLASS to the ECLASS case, but changed a break into a return (oops!).

src/pcre2_auto_possess.c
```diff
@@ -1115,6 +1117,12 @@ for(;;)
       break;
 #endif
 
+      case OP_ECLASS:
+      if (PRIV(eclass)(chr,
+          (list_ptr == list ? code : base_end) - list_ptr[2] + LINK_SIZE,
+          (list_ptr == list ? code : base_end) - list_ptr[3], utf)) return FALSE;
```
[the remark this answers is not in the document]
NWilson: Interesting. There are five instances of `[^ ]\?` and ten instances of ` \?` in this particular file.

It seems to be formatted consistently as `list_ptr == list ? code : base_end` throughout the file, so I'll keep it for now, rather than do extra reformatting.

src/pcre2_extuni.c
NWilson: I don't think that the ECLASS code calls into pcre2_extuni at all.

I haven't changed anything in here, except noticed a typo in a comment that was copy-and-pasted into multiple files.

src/pcre2_jit_compile.c
NWilson: I removed the ifdefs for cases where they were used to remove code that wasn't needed.

@PhilipHazel would you prefer:

```c
  case OP_ECLASS:
  case OP_XCLASS:
  return cc + GET(cc, 1);
```

or:

```c
  case OP_ECLASS:
#ifdef SUPPORT_WIDE_CHARACTERS
  case OP_XCLASS:
#endif
  return cc + GET(cc, 1);
```

It seems a bit excessive use ifdefs just to remove the "case", when the code underneath is required regardless.

carenas: the point I was making is that if the JIT compiler ever gets an OP_XCLASS in the 8 bit library without utf (which before couldn't happen because it was not even defined) it might break, so it is safer to detect that case here as unsupported IMHO.

NWilson: Oh I see. That's fair; I'll add the check you describe to check_opcode_types().

testdata/testoutput2
```diff
@@ -19612,6 +19612,939 @@ Failed: error -49 at offset 3 in replacement: unknown substring
...
+Failed: error 212 at offset 5: missing terminating ] for character class (note '[' must be escaped under PCRE2_ALT_EXTENDED_CLASS)
+
+/[a[B]]C/B,alt_extended_class
+------------------------------------------------------------------
+        Bra
+        eclass[
+          cls:[a]
+          cls:[B]
+          op: ||
+        ]
```
zherczeg, line 19636: I just realized an optimization is missing here.

PCRE2 traditionally uses a bitset for the first 256 characters of sets. Even for the new ECLASS, there should only be one bitset.

Should work like this:
- The ECLASS contains only XCLASS byte codes. 
- If, and only if an XCLASS has a bitset, then the ECLASS has a bitset.
- The ECLASS bitset is pre-computed, no extra operation is needed
- The ECLASS should be turned to CLASS / NCLASS if nothing else only the bitset is present

I know this is a complex, although important optimization, since it reduces storage and improves performance. I can do it in a follow-up patch.

NWilson: If we do full constant-folding for the CLASS/NCLASS cases, we'd not have many redundant bitsets, in practice.

I agree we could lift out the bitset, if that's important. I guess we'd need a way to fold things like `CLASS && XCLASS` and so on - remove the bitset from the XCLASS, and since the CLASS includes all chars >255 we can reduce `CLASS && XCLASS → XCLASS`.

I think I can do this fine in next week's PR, when I do the `CLASS op CLASS → CLASS` constant-folding.

NWilson: Thanks though @zherczeg, it's a great observation, and worth doing.

NWilson:
I'm done with this!!

Whew. I'm not aware of anything broken, and it's feature-complete, as far as the ALT_EXTENDED_CLASS flag goes.

I've filed tickets for the follow-up work, so it's not forgotten (optimizations; and `(?[...])` support), and I'll commit to doing it beginning next week.

I've either addressed all of Zoltan and Carlo's feedback so far, or asked for it the points raised to be deferred. 

Many, many thanks for all the help reviewing and testing this!

testdata/testoutput2
zherczeg: If you have other priorities, I can also do these things. I don't want to put more work on you.

src/pcre2_auto_possess.c
```diff
@@ -479,13 +480,13 @@ switch(c)
 
   case OP_NCLASS:
   case OP_CLASS:
-#ifdef SUPPORT_WIDE_CHARS
```
zherczeg, line 482: I would add some `// revert` comments because we should revert these after ECLASS will have a single bitset. It is easy to forget them otherwise.

src/pcre2_auto_possess.c
```diff
@@ -1216,20 +1224,18 @@ for (;;)
       }
     c = *code;
     }
-  else if (c == OP_CLASS || c == OP_NCLASS || c == OP_XCLASS)
+  else if (c == OP_CLASS || c == OP_NCLASS || c == OP_XCLASS || c == OP_ECLASS)
     {
-#ifdef SUPPORT_WIDE_CHARS
```
zherczeg, line 1221: Ditto.

src/pcre2_auto_possess.c
```diff
@@ -1295,11 +1301,10 @@ for (;;)
     code += GET(code, 1 + 2*LINK_SIZE);
     break;
 
-#ifdef SUPPORT_WIDE_CHARS
```
zherczeg, line 1298: Ditto.

src/pcre2_compile.c
```diff
@@ -3737,9 +3716,12 @@ while (ptr < ptrend)
...
         error on the user's part. */
 
-        if (ptr < ptrend - 1 && *ptr == CHAR_MINUS &&
-            ptr[1] != CHAR_RIGHT_SQUARE_BRACKET)
+        if (ptr < ptrend && *ptr == CHAR_MINUS &&
+            !((options & PCRE2_ALT_EXTENDED_CLASS) != 0 && ptr + 1 < ptrend &&
+              ptr[1] == CHAR_MINUS) &&
+            (ptr + 1 == ptrend || ptr[1] != CHAR_RIGHT_SQUARE_BRACKET))
           {
+          // TODO: We need a better lookahead here to match Perl; should skip \Q\E, and whitespace in /xx mode.
```
zherczeg, line 3724: Maybe a `/* TODO: EC revert */` will be a better prefix. Then we can just find `TODO: EC` to update all related things.

Note: PCRE2 does not use // comments at all.

zherczeg (commented):
We have discussed a lot of things in this PR, and would be good to add a specific comment (with a fixed string which is easy to search), to the code where we will likely do more things. This way we could not forget about them.

src/pcre2_compile.c
PhilipHazel: That's because C90 did not have // comments, but nowadays I do occasionally use them for "temporary" comments that can easily be found for removal.

carenas:
> do we need to add (*ALT_EXTENDED_CLASS) ?

I actually had a patch implementing `(?X)`, but didn't send it because the "incompatibilities" between the old syntax and the Unicode one are so minimal that I think it would be better not to have one, and wait to see what happens after it gets released.

src/pcre2_compile_class.c
```diff
@@ -319,14 +319,14 @@ append_non_ascii_range(uint32_t options, uint32_t *buffer)
 }
 
 static size_t
-parse_class(uint32_t *ptr, uint32_t options, uint32_t *buffer)
+parse_class(uint32_t *ptr, uint32_t *end_ptr, uint32_t options, uint32_t *buffer)
```
carenas, line 322: end_ptr could be const

carenas (approved):

src/pcre2_auto_possess.c
NWilson: I'm not quite sure what you intend here. Even when ECLASS has a single bitset, it will behave roughly the same as XCLASS, for the purposes of the auto-possessification code at least. There will always need to be the LINK_SIZE slot, holding the size of the stack which follows.

I'll add the TODO comment as you request.

zherczeg: Wide chars mean unicode or > 8 bit. In 8 bit mode, without utf, the regular OP_CLASS with 256 bits must handle the extended class, so it should never encounter with this opcode. In general, eclasses should be turned to normal classes/nclasses whenever it is possible, just like xclasses.

zherczeg: As for the implementation, eclasses should have a prediction system, which predicts their size if they cannot be turned to class/nclass. The prediction is discarded, if they can. The prediction bit (can or cannot) can be saved in the META code (you have 16 free bits for that) during the byte code length computation phase, and reused when the actual byte code is generated.

NWilson: Of course, I see! If the constant-folding is sufficiently aggressive, then OP_ECLASS will never be emitted without wide-char support, the same as XCLASS. OK, that makes good sense.

zherczeg: Btw, both the "need a bitset for eclass" and "eclass can be turned to class/nclass" bits can be computed when the byte code length is computed (two bits in the META code), and this simplifies the actual byte code generation later, since "the oracle" tell us what to do.

src/pcre2_xclass.c
```diff
@@ -441,4 +441,96 @@ while (TRUE)
...
+    case OP_ALLANY:
+    ++ptr;
+    stack = (stack << 1) | 1u;
+    break;
+
+    /* This should never occur, but compilers may mutter if there is no
+    default. */
+
+    default:
+    PCRE2_DEBUG_UNREACHABLE();
```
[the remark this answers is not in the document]
NWilson: I just copied it from the existing code above.

I'm happy with return statements after DEBUG_UNREACHABLE() assertions, because in release builds, the assertions aren't emitted at all, so _something_ has to happen if a customer actually triggers this condition.

Regardless of whether compilers complain, I complain if I see switch statements without a default condition! An assertion is highly beneficial for maintainability.

NWilson:
> Incidentally, do we need to add (*ALT_EXTENDED_CLASS) ?

I think not Philip. We are currently missing start-of-pattern, and in-pattern, controls for all of the following: ALLOW_EMPTY_CLASS, ALT_BSUX, DOLLAR_ENDONLY, MATCH_UNSET_BACKREF, ALT_CIRCUMFLEX, and others.

These are all flags which affect the regex dialect in some way, rather than fundamentally change what is or isn't expressible.

I don't think we need a start-of-pattern flag for any of these, nor for ALT_EXTENDED_CLASS: if a pattern author knows to add the flag, they can just as easily edit the pattern itself to match PCRE's dialect. (Admittedly, that would be fiddly for a flag like ALT_CIRCUMFLEX, but certainly possible to rewrite it in terms of `^` and `\z`. Similarly, anyone who wants ALT_EXTENDED_CLASS can just rewrite to use the Perl extended syntax.)

NWilson:
I have addressed Carlo and Zoltan's latest review comments: more TODOs; Carlo's minor improvements.

src/pcre2_compile.c
NWilson: I have added TODOs as indicated by Zoltan, plus some more all.

All of them have the string `TODO: [EC]` and a link to the GitHub issue.

src/pcre2_xclass.c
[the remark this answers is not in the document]
carenas: fair enough, but then the argument still stands for using a `PCRE2_UNREACHABLE()` instead as that return should never happen in  release build anyway.

testdata/testoutput2
zherczeg: Thank you.

src/pcre2_xclass.c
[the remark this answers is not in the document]
NWilson: Currently it looks like the macros emit:

|---|---|---|
| | Debug | Non-debug |
|---|---|---|
| PCRE2_DEBUG_UNREACHABLE | __assume(0) | do {} while (0) |
| PCRE2_UNREACHABLE | __assume(0) | __assume(0) |

This looks correct to me.

The DEBUG_UNREACHABLE version should have error-recovery code, such as a return statement, but the UNREACHABLE() version shouldn't. I like that.

I think __assume(0) is correct for MSVC.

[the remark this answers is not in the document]
NWilson: Actually, I think it should emit the following instead:

| | Debug | Non-debug |
|---|---|---|
| PCRE2_DEBUG_UNREACHABLE | PCRE2_ASSERT(FALSE) | do {} while (0) |
| PCRE2_UNREACHABLE | PCRE2_ASSERT(FALSE) | __assume(0) |

We want some actual runtime code to call abort() when we're in a debug build.

I can make a PR to do that separately.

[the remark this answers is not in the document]
zherczeg: I have noticed recently that `assert()` built-in function has  pretty nice output, looks better than `abort()`

src/pcre2_internal.h
```diff
@@ -574,6 +574,13 @@ modes. */
 #define REQ_CU_MAX       2000
 #endif
 
+/* The maximum nesting depth for Unicode character class sets.
+Currently fixed. Warning: the interpreter relies on this so it can encode
+the operand stack in a uint32_t. A nesting limit of 15 implies (15*2+1)=31
+stack operands required. */
+
+#define ECLASS_NEST_LIMIT  15
```
zherczeg, line 582: I thought 1 bit is enough for an xclass result value. Why do you need two?

src/pcre2_pattern_info.c
```diff
@@ -384,11 +384,10 @@ while (TRUE)
 #endif
     break;
 
-#if defined SUPPORT_UNICODE || PCRE2_CODE_UNIT_WIDTH != 8
```
zherczeg, line 387: This should have the revert comment as well.

src/pcre2_printint.c
```diff
@@ -442,6 +443,148 @@ return (PCRE2_SPTR)next_char;
...
+      if (PRINTABLE(i)) fprintf(f, "%c", i);
+        else fprintf(f, "\\x%02x", i);
+      if (--j > i)
+        {
+        if (j != i + 1) fprintf(f, "-");
+        if (j == '-' || j == ']') fprintf(f, "\\");
+        if (PRINTABLE(j)) fprintf(f, "%c", j);
+          else fprintf(f, "\\x%02x", j);
+        }
+      i = j;
```
zherczeg, line 520: It seems all code is moved here. Is this correct?

zherczeg (commented):
The commits should be squashed. I have only a few questions. Overall this patch is a good first step.

Let's wait a bit longer for @PhilipHazel , but if he is away, I can also land this patch.

PhilipHazel:
I am not away, but I have been letting you guys get on with this patch. I have not looked at it in detail.Please go ahead and merge when you are all happy. (I am also distracted getting a new desktop PC up and configured because my old one is over 10 years old and I'm not sure how long it will go on.)

src/pcre2_internal.h
NWilson: Aha, because you can have two operators within a '[...]'. So if you have two brackets: `[A--B[C--DE]]` then the worst-case and peak stack usage is that the leafmost `[C--DE]` takes three slots on the stack (result of C, D and E, before applying || to DE and --), and the non-leafmost `[A--B ...foo...]` takes two slots on the stack (need to store result of matching A and B).

Hence 5 stack slots for two `[ ... [...] ]` brackets. It's 2n+1.

It would be worse if we allowed mixing precedence within the brackets; then each `[...]` could consume even more stack slots.

src/pcre2_pattern_info.c
NWilson: I thought that it was OK to just leave the (test-only) code here clean of ifdefs. But I can do if you prefer the ifdefs to come back.

src/pcre2_printint.c
NWilson: Yes. I lifted out the existing code for printing character classes into a function.

Similarly, much of the "new" code in pcre2_compile_class.c is lifted out of pcre2_compile.c.

src/pcre2_internal.h
zherczeg: I thought the pushes/pops are tracked individually. Something like `/[A]&&[B]||[C]--[D]/` is `push [A], push [B], &&, push [C], ||, push[D], --` and the max stack size is 2.

Recursive descent should be able to compute it. This looks like another thing to do then.

zherczeg:
Ok, patch is landed. I except a lot of follow up works, but it is the nature of these large features.

src/pcre2_internal.h
NWilson: For your example (`/[A]&&[B]||[C]--[D]/`), yes, the maximum stack size is 2.

The worst case is a stack size of three, if you have `A op1 B op2 C` where op2 is higher (tighter) precedence.

We have a hard limit of 15 nested brackets; which gives us a maximum stack consumption of 31 slots.

Of course we can calculate (if we want) what the actual stack slots consumed will be, but we guarantee it will fit into a uint32_t if we limit the brackets to at most 15 deep.

NWilson:
Great, thank you! My next PR should follow later this week.

PR #524, opened 2024-10-11
Improve character range matching with binary search

zherczeg:
Quite large patch, and only the interpreter is supported at this point. At least the RunTest works with -nojit.

With -O3 and -tm 100000.
Old code:
```
  re> /[\h\v]++/utf
data> \x{3000}\x{3000}\x{3000}\x{3000}\x{3000}\x{3000}\x{3000}\x{3000}
Match time  0.9488 microseconds
 0: \x{3000}\x{3000}\x{3000}\x{3000}\x{3000}\x{3000}\x{3000}\x{3000}
```
New code:
```
  re> /[\h\v]++/utf
data> \x{3000}\x{3000}\x{3000}\x{3000}\x{3000}\x{3000}\x{3000}\x{3000}
Match time  0.6354 microseconds
 0: \x{3000}\x{3000}\x{3000}\x{3000}\x{3000}\x{3000}\x{3000}\x{3000}
```

At least it has a measurable effect even with interpreter.


zherczeg:
@carenas you are very good at these things, could you help me? Clang complains:

src/pcre2_xclass.c:410:19: runtime error: left shift of 2147483648 by 1 places cannot be represented in type 'uint32_t' (aka 'unsigned int')

This is exactly the definition of shifting, and what I want. Why clang complains? Why only here? It happens many times in other places. Casting did not help. I can add an `and mask` operation, but it just slows down the code.

zherczeg:
The code is mostly complete, it adds 1000 line new code, and another 500 rework. It is really a complex code, I have to admit.

A test:
```
  re> /[\x{100}-\x{200}\x{300}-\x{400}]+/Bi,utf
        Bra
        [Ss\xb5\xff\x{100}-\x{201}\x{220}\x{23d}\x{243}\x{253}-\x{254}\x{256}-\x{257}\x{259}\x{25b}\x{260}\x{263}\x{268}-\x{269}\x{26f}\x{272}\x{275}\x{280}\x{283}\x{288}\x{28a}-\x{28b}\x{292}\x{300}-\x{400}\x{450}\x{1fbe}\x{1fd3}\x{1fe3}\x{2126}\x{a7dc}]++
        Ket
        End

Old, no-jit: 2.5799 microseconds
New, no-jit: 1.4730 microseconds
Old, jit: 1.0704 microseconds
New, jit: 1.1159 microseconds
```

It is clearly better for interpreter, not much for jit.

When I fully turn the list to machine code (simple checks / lots of branches), I get:
```
New, jit: 0.7576 microseconds
```
This is faster, although the code size is also bigger.

It is a good question what shall we do. Increased code size is ok or not. Based on my past experiences, the increased code size is usually not a problem, but it is worth to discuss.

carenas:
> Why clang complains? 

Because of the "Undefined Behaviour Sanitizer", specifically the "integer" one that only clang has and that does a lot of controversial checks that are not even "Undefined Behaviour" like this one.

> Why only here? It happens many times in other places.

All other places are likely also problematic, but with the provided data this is the only place that triggered. The sanitizers add additional code and do the checks at runtime.

> Casting did not help. I can add an `and mask` operation, but it just slows down the code.

I think this is another place were the sanitizer seems to be at fault, I think the best option would be to add `-fno-sanitize=unsigned-shift-base` to those jobs, and sorry for the disruption

src/pcre2_intmodedep.h
```diff
@@ -729,8 +729,11 @@ typedef struct named_group {
 of translating META code to byte code. */
 
 typedef struct class_ranges {
-  struct class_ranges *next;  /* Next class ranges */
-  size_t range_list_size;     /* Size of ranges array */
+  struct class_ranges *next;       /* Next class ranges */
+  uint16_t range_list_size;        /* Size of ranges array */
+  uint16_t char_lists_types;       /* Bit set for char lists. */
```
carenas, line 734: move the uint16_t at the end of the struct to avoid unnecessary padding

zherczeg: I will, although in this particular case, I just move the padding at the end.

zherczeg:
I decided to simplify the jit part, since it seems faster. I can make it more sophisticated later. This patch is big enough.

zherczeg:
@carenas your patch is working, could you make a PR from it?

zherczeg:
I have updated the patch and I think it is ready for review.

carenas:
> could you make a PR from it?

Sorry got sidetracked and didn't; but was assuming that it might be better if included in this change itself, after all, the code that requires the restriction comes with it and I tried a shot at documenting its use case in `maint/README` that might be worth including and might need improving.

Eitherway, let me know what you would prefer and I'll execute.

carenas:
Let's avoid the extra PR then (I always worry my PRs, even the trivial ones, are just too much work since they need code review and an independent merge), sorry you will have to endorse my documentation update as well, but hopefully makes more sense when included together with the code, even if I got the description slightly off.

zherczeg:
I have added the readme update as well. Unfortunately git update didn't go well with an indentation change of a large chunk of existing code, and I have to manually restore it from an older version of the code and applying the turkish change manually. I hope I didn't make any mistakes :S

The code is still ready for review. I know it is not a simple task, sorry for taking a lot of your time.

zherczeg:
The table is compiled to machine code basically. This is the current status as well, just without any ordering. A sequence of immediate compares and jumps. Works well for a few dozen items, and I suspect this is the majority of use cases. If people would use huge sets, we probably would get many complaints before.

There are still options to explore (only in future patches):
- Merge clists into the table: increase the size a bit, but will be handled by logarithmic search
- Use conditional move to reduce the number of jumps (less stress on branch predictor, bit more code size)
- JIT collects categories into a 32 bit bitset during compilation, and do all category check in 1 step. This can be done by the interpreter, if we want it. Can be useful for #523

zherczeg:
Is this patch ok?

PhilipHazel:
I'm assuming @carenas will answer your question.... The principle of this seems good; I have not looked in detail.

carenas (approved):
might be nice cherry-picking 3e795ba6f9124cea6dcdf54920ff3bf13433499a and squahing it first though

carenas:
since @NWilson is touching the same areas, it might be good for him to give a more formal "code review" though

src/pcre2_xclass.c
```diff
@@ -297,12 +263,182 @@ while ((t = *data++) != XCL_END)
...
+#if PCRE2_CODE_UNIT_WIDTH == 8
+type = (uint32_t)(data[0] << 8) | data[1];
+data += 2;
+#else
+type = data[0];
+data++;
+#endif  /* CODE_UNIT_WIDTH */
+
+/* Align characters. */
+next_char = (uint8_t*)data;
```
carenas, line 323: ditto

src/pcre2_printint.c
```diff
@@ -318,6 +319,129 @@ else
...
+#if PCRE2_CODE_UNIT_WIDTH == 8
+type = (uint32_t)(code[0] << 8) | code[1];
+code += 2;
+#else
+type = code[0];
+code++;
+#endif  /* CODE_UNIT_WIDTH */
+
+/* Align characters. */
+next_char = (uint8_t*)code;
```
carenas, line 352: ditto, and here it even makes the code more flexible IMHO

src/pcre2_xclass.c
```diff
@@ -70,7 +70,9 @@ PRIV(xclass)(uint32_t c, PCRE2_SPTR data, BOOL utf)
 {
 /* Update PRIV(update_classbits) when this function is changed. */
 PCRE2_UCHAR t;
-BOOL negated = (*data & XCL_NOT) != 0;
+BOOL not_negated = (*data & XCL_NOT) == 0;
+uint32_t type, max_index, min_index, average, value;
```
carenas, line 74: nitpick: how about using "mid_index" instead of average?

src/pcre2_study.c
```diff
@@ -921,6 +921,138 @@ if (table_limit != 32) for (c = 24; c < 32; c++) re->start_bitmap[c] = 0xff;
...
+uint8_t *next_char;
+PCRE2_UCHAR start_buffer[6], end_buffer[6];
+PCRE2_UCHAR start, end;
+
+/* Only needed in 8-bit mode at the moment. */
+type = (uint32_t)(code[0] << 8) | code[1];
+code += 2;
+
+/* Align characters. */
+next_char = (uint8_t*)code;
```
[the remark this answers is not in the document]
zherczeg: The PCRE2_SPTR is uses mode dependent characters, I need to know the exact type here, and byte is useful for aligning as well (it is not nice to store an unaligned pointer).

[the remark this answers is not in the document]
carenas: FWIW, in THIS function PCRE2_SPTR is always a pointer to bytes, which you could even make "const", so here it would work eitherway, but I see why it is better to be explicit after I went wild thinking I might have found a suggestion to improve the code ;)

src/pcre2_jit_compile.c
```diff
@@ -7685,8 +7685,233 @@ return cc;
...
+  /* Stack size must be log2(ranges / 2). */
+  xclass_stack_item *stack;
+  xclass_stack_item local_stack[XCLASS_LOCAL_RANGES_LOG2_SIZE];
+} xclass_ranges;
+
+static void xclass_compute_ranges(compiler_common *common, PCRE2_SPTR cc, xclass_ranges *ranges)
+{
+DEFINE_COMPILER;
+size_t range_count = 0, est_range_count;
+size_t total_size, est_stack_size, tmp;
```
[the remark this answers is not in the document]
zherczeg: It is used once, so I removed it. Why shadowing is a problem btw? The whole point of local scoping is shadowing.

src/pcre2_study.c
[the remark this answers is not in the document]
zherczeg: I forgot this is the study code. I don't want to change it for consistency with other codes. Adding const could work though.

src/pcre2_jit_compile.c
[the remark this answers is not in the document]
carenas: Philip doesn't like it, and I think it is because with functions being so large, it is more difficult to keep track of their values when they shadow.

Of course nothing prevents you using the same variable in smaller scopes, like I was suggesting for "average", so I think by being strict about it, also pushes for smaller scopes overall that can only help long term with maintenance.

src/pcre2_compile.c
```diff
@@ -6405,76 +6407,86 @@ for (;; pptr++)
...
         }
-
-      while (range < end)
+      else
         {
-        uint32_t range_start = range[0];
-        uint32_t range_end = range[1];
+        if ((xclass_props & XCLASS_HIGH_ANY) != 0)
+          {
+          PCRE2_ASSERT(range + 2 == end && range[0] <= 256 &&
```
carenas, line 6420: isn't this supposed to be `< 256` here?

zherczeg: No. The first 256 (0..255) characters are handled by OP_CLASS/OP_NCLASS, and characters >= 256 are all match or don't.

carenas: makes sense now, so should we ship ip?

zherczeg: What does "ip" means? it?

carenas: yes, sorry for the typo

carenas: had prepared 35f47c52aa62c7a4eb1ac31d82583e6e408dbc8e with a few more related const that might be nice to include, but nothing to object.

zherczeg: I have added them. I land the patch then.

src/pcre2_jit_compile.c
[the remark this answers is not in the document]
PhilipHazel: Yes, indeed, I don't like shadow variables because they can be confusing if you are reading unfamiliar code. Suppose you see something like this:

```
       X = some value;
       more code...
       }
    code involving X 
```
You would probably expect the second X to be the same variable as the first X and then when it doesn't make sense you have to look back up the code to figure it out. I do know that I tend to write very large functions, which makes this worse. I blame it on having started out when compilers were not clever and calling a function was more costly than it is now. (And computers were slower, of course. :-)

PR #525, opened 2024-10-11
Address most implicit 64-to-32-bit conversion warnings.

ucko:
Visual Studio and Xcode both issue such warnings by default, and Clang also offers a `-Wshorten-64-to-32` option to expose them elsewhere.  To address them, harmonize types where practical and add casts as appropriate (first formally capping executables' read request sizes as necessary).  NB: `detect_repeat` in `pcre2_jit_compile.c` still yields two such warnings that look plausibly legitimate but not so easy to fix. :-/

### DIFF
--- a/src/pcre2.h.generic
+++ b/src/pcre2.h.generic
@@ -667,7 +667,7 @@ PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
 PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
   pcre2_set_depth_limit(pcre2_match_context *, uint32_t); \
 PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
-  pcre2_set_heap_limit(pcre2_match_context *, uint32_t); \
+  pcre2_set_heap_limit(pcre2_match_context *, PCRE2_SIZE); \
 PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
   pcre2_set_match_limit(pcre2_match_context *, uint32_t); \
 PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
--- a/src/pcre2_chkdint.c
+++ b/src/pcre2_chkdint.c
@@ -63,7 +63,7 @@ The INT64_OR_DOUBLE type is a 64-bit integer type when available,
 otherwise double. */
 
 BOOL
-PRIV(ckd_smul)(PCRE2_SIZE *r, int a, int b)
+PRIV(ckd_smul)(PCRE2_SIZE *r, ptrdiff_t a, ptrdiff_t b)
 {
 #ifdef HAVE_BUILTIN_MUL_OVERFLOW
 PCRE2_SIZE m;
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -2362,7 +2362,7 @@ SLJIT_ASSERT(stackpos == STACK(stacktop));
 typedef struct delayed_mem_copy_status {
   struct sljit_compiler *compiler;
   int store_bases[RECURSE_TMP_REG_COUNT];
-  int store_offsets[RECURSE_TMP_REG_COUNT];
+  sljit_sw store_offsets[RECURSE_TMP_REG_COUNT];
   int tmp_regs[RECURSE_TMP_REG_COUNT];
   int saved_tmp_regs[RECURSE_TMP_REG_COUNT];
   int next_tmp_reg;
@@ -3302,7 +3302,8 @@ OP2(SLJIT_SUB | SLJIT_SET_Z, COUNT_MATCH, 0, COUNT_MATCH, 0, SLJIT_IMM, 1);
 add_jump(compiler, &common->calllimit, JUMP(SLJIT_ZERO));
 }
 
-static SLJIT_INLINE void allocate_stack(compiler_common *common, int size)
+static SLJIT_INLINE void allocate_stack(compiler_common *common,
+                                        PCRE2_SIZE size)
 {
 /* May destroy all locals and registers except TMP2. */
 DEFINE_COMPILER;
@@ -3319,7 +3320,7 @@ OP1(SLJIT_MOV, SLJIT_MEM1(SLJIT_SP), LOCALS1, TMP1, 0);
 add_stub(common, CMP(SLJIT_LESS, STACK_TOP, 0, STACK_LIMIT, 0));
 }
 
-static SLJIT_INLINE void free_stack(compiler_common *common, int size)
+static SLJIT_INLINE void free_stack(compiler_common *common, PCRE2_SIZE size)
 {
 DEFINE_COMPILER;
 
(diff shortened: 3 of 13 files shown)


src/pcre2_chkdint.c
```diff
@@ -63,7 +63,7 @@ The INT64_OR_DOUBLE type is a 64-bit integer type when available,
 otherwise double. */
 
 BOOL
-PRIV(ckd_smul)(PCRE2_SIZE *r, int a, int b)
+PRIV(ckd_smul)(PCRE2_SIZE *r, ptrdiff_t a, ptrdiff_t b)
```
carenas, line 66: the whole point of this function is to make sure those 2 parameters are `int`so we can check for overflow

src/pcre2_jit_compile.c
```diff
@@ -3302,7 +3302,7 @@ OP2(SLJIT_SUB | SLJIT_SET_Z, COUNT_MATCH, 0, COUNT_MATCH, 0, SLJIT_IMM, 1);
 add_jump(compiler, &common->calllimit, JUMP(SLJIT_ZERO));
 }
 
-static SLJIT_INLINE void allocate_stack(compiler_common *common, int size)
+static SLJIT_INLINE void allocate_stack(compiler_common *common, sljit_uw size)
```
carenas, line 3305: probably should had been PCRE2_SIZE but were missed when PCRE2 was created from PCRE1 (which used `int` for all offsets)

src/pcre2_jit_neon_inc.h
```diff
@@ -87,7 +87,7 @@ POSSIBILITY OF SUCH DAMAGE.
 	&& ((__clang_major__ == 3 && __clang_minor__ >= 3) || (__clang_major__ > 3)))
 __attribute__((no_sanitize_address))
 #endif
-static sljit_u8* SLJIT_FUNC FF_FUN(sljit_u8 *str_end, sljit_u8 **str_ptr, sljit_uw offs1, sljit_uw offs2, sljit_uw chars)
+static sljit_u8* SLJIT_FUNC FF_FUN(sljit_u8 *str_end, sljit_u8 **str_ptr, sljit_uw offs1, sljit_uw offs2, sljit_u32 chars)
```
carenas, line 90: this should break the 16bit library

carenas (commented):
Presume from your commit message you were trying to remove warnings from the code when built in a 64bit system, could you elaborate on what is the driver behind that?

src/pcre2_jit_compile.c
```diff
@@ -2362,7 +2362,7 @@ SLJIT_ASSERT(stackpos == STACK(stacktop));
 typedef struct delayed_mem_copy_status {
   struct sljit_compiler *compiler;
   int store_bases[RECURSE_TMP_REG_COUNT];
-  int store_offsets[RECURSE_TMP_REG_COUNT];
+  sljit_sw store_offsets[RECURSE_TMP_REG_COUNT];
```
zherczeg, line 2365: These are intentionally small values.

src/pcre2_jit_compile.c
```diff
@@ -7706,7 +7706,10 @@ static void compile_xclass_matchingpath(compiler_common *common, PCRE2_SPTR cc,
 DEFINE_COMPILER;
 jump_list *found = NULL;
 jump_list **list = (cc[0] & XCL_NOT) == 0 ? &found : backtracks;
-sljit_uw c, charoffset, max = 256, min = READ_CHAR_MAX;
+/* These values should never be negative, suggesting the use of sljit_u32.
+However, they should also remain (far) below 1 << 31, and (would-be)
+negative *differences* won't otherwise undergo correct sign extension. */
+sljit_s32 c, charoffset, max = 256, min = READ_CHAR_MAX;
```
zherczeg, line 7712: c can be UINT32_MAX, so this is also a bad change.

zherczeg (changes requested):
Please be very careful when changing jit types.

ucko:
Thanks for the quick feedback!  I've force-pushed a largely corrected (and rebased) version and will reply on individual points shortly.  AFAICT, some of these changes really should help in corner cases; although the remainder are formalities, the (configuration-dependent) warnings they address could bury other warnings (none here at present, I'm happy to report).

I acknowledge that this PR still needs more work.

src/pcre2_chkdint.c
ucko: Ah, yeah, the hand-coded branch would need reworking.  (`__builtin_mul_overflow` should DTRT regardless.)  However, without this change, `pcre2test`'s `process_data` could in principle pass in a truncated `replen`, yielding incorrect results.  Should it instead bail if `replen > INT_MAX`?

src/pcre2_jit_compile.c
ucko: `PCRE2_SIZE` substituted, thanks!

src/pcre2_jit_neon_inc.h
ucko: Not AFAICT, though I must admit I'd initially focused the 8-bit one.

src/pcre2_jit_compile.c
ucko: Without this change, `delayed_mem_copy_move`'s assignment into `store_offsets` can wind up truncated.  Attempting to narrow the type of its `store_offset` argument instead reveals that three call sites pass in 64-bit values.

src/pcre2_jit_compile.c
ucko: Good point.  I've scrapped all changes to `compile_xclass_matchingpath` in favor of formally making `read_char`'s `min` and `max` `sljit_uw`.

src/pcre2.h.generic
```diff
@@ -667,7 +667,7 @@ PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
 PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
   pcre2_set_depth_limit(pcre2_match_context *, uint32_t); \
 PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
-  pcre2_set_heap_limit(pcre2_match_context *, uint32_t); \
+  pcre2_set_heap_limit(pcre2_match_context *, PCRE2_SIZE); \
```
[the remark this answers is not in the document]
ucko: Oops, yeah, I didn't mean to introduce an ABI break. :-/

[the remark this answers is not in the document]
ucko: Ah, right, I'll need to revert the POSIX API changes too.

src/pcre2_chkdint.c
carenas: pcre2test is a test application, AFAIK `replen` was made a size_t just to avoid having to do a cast, and it is expected never to have a value larger than int either, so a truncation there shouldn't be possible, but feel free to poke at it and prove me wrong.

the call for ckd_smul was indeed how we are making sure that the value of the multiplication is always correct.

NWilson:
Thank you Aaron (@ucko) for pushing us to fix these warnings.

I have addressed them in PR #655, hopefully. I took a slightly different direction to you on some of the fixes. We now pass with `/W3` warnings on Windows (MSVC).

I have not addressed all the warnings generated by `-Wshorten-64-to-32` (GCC, Clang) since that is far spammier than MSVC's equivalent.

Let me know if you are satisfied with the level of warnings suppression on `master`. I would be happy to fix any additional warnings that are causing you problems with your company's build policy, if you can point out where they are.

ucko:
@NWilson Great, thank you very much!  Your fixes are a big improvement, but I'd appreciate it if you could please take another look at `pcre2_jit_neon_inc.h`.  I also still see a handful of warnings from `pcre2grep.c` and `pcre2test.c`, but they're less of a concern.  I'll elaborate on your #655.

NWilson:
Thank you for this PR. I think we have now incorporated all the changes here. Many thanks!

Issue #526, opened 2024-10-12
Bug in char class handling, in caseless mode, with UCP (caused by PR 474)


alexdowad:
I've bisected this one and found that behavior was changed by #474:

```
/[a\x{c1}]/iI,ucp
------------------------------------------------------------------
  0  36 Bra
  3     [Aa\xc1]
 36  36 Ket
 39     End
------------------------------------------------------------------
Capture group count = 0
Options: caseless ucp
Starting code units: A a \xc1
Subject length lower bound = 1
    \x{e1}
No match

/[a\x{c1}]/iI,ucp,no_start_optimize
------------------------------------------------------------------
  0  36 Bra
  3     [Aa\xc1]
 36  36 Ket
 39     End
------------------------------------------------------------------
Capture group count = 0
Options: caseless no_start_optimize ucp
Optimizations: auto_possess,dotstar_anchor
    \x{e1}
No match
```

Before #474, `\xe1` would be included in the starting code units when `\xc1` was in a leading char class, and caseless mode was enabled. Now it is not.

Further, such a char class no longer matches `\xe1`, whereas previously it did. This is not just the starting code point optimization causing early failure of the match... see the above test case with `no_start_optimize`.

For reference, these are the results on the commit just before #474:

```
/[a\x{c1}]/iI,ucp
------------------------------------------------------------------
  0  36 Bra
  3     [Aa\xc1\xe1]
 36  36 Ket
 39     End
------------------------------------------------------------------
Capture group count = 0
Options: caseless ucp
Starting code units: A a \xc1 \xe1
Subject length lower bound = 1
    \x{e1}
 0: \xe1

/[a\x{c1}]/iI,ucp,no_start_optimize
------------------------------------------------------------------
  0  36 Bra
  3     [Aa\xc1\xe1]
 36  36 Ket
 39     End
------------------------------------------------------------------
Capture group count = 0
Options: caseless no_start_optimize ucp
Optimizations: auto_possess,dotstar_anchor
    \x{e1}
 0: \xe1
```

Attention please, @zherczeg.

zherczeg:
Nice! Can you fix it or need my help? In 8 bit mode we don't run range computation, it looks like it is needed if ucp is set.

carenas:
> it looks like it is needed if ucp is set.

`ucp` shouldn't be relevant for this IMHO and is probably  red herring.

[...]

PR #527, opened 2024-10-13
Even in 8-bit mode, perform range computation for char classes if UCP flag is set

alexdowad:
When testing another patch, I discovered that #474 caused a small change in the behavior of character classes when caseless mode and UCP were enabled.

Thank you to Zoltan Herczeg for suggesting a fix.

Closes GH-526.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -5891,7 +5891,7 @@ for (;; pptr++)
 #if PCRE2_CODE_UNIT_WIDTH == 8
     cranges = NULL;
 
-    if (utf)
+    if (utf || ucp)
 #endif
       {
       if (lengthptr != NULL)
@@ -6388,6 +6388,12 @@ for (;; pptr++)
         range = end;
         }
 
+#if PCRE2_CODE_UNIT_WIDTH == 8
+      /* If code unit width is 8 bits, and UCP flag is set, but UTF flag is not, we still
+       * generate cranges, but in that case we should not process any crange > 0xFF,
+       * because it's impossible to encounter code points > 0xFF in the subject string */
+      if (utf)
+#endif
       while (range < end)
         {
         uint32_t range_start = range[0];
@@ -6479,6 +6485,9 @@ for (;; pptr++)
 #ifdef SUPPORT_WIDE_CHARS  /* Defined for 16/32 bits, or 8-bit with Unicode */
     if ((xclass_props & XCLASS_REQUIRED) != 0)
       {
+      /* We should never generate a (useless) xclass in 8-bit library when UTF flag is false */
+      PCRE2_ASSERT(PCRE2_CODE_UNIT_WIDTH != 8 || utf);
+
       *class_uchardata++ = XCL_END;    /* Marks the end of extra data */
       *code++ = OP_XCLASS;
       code += LINK_SIZE;
--- a/src/pcre2_compile_class.c
+++ b/src/pcre2_compile_class.c
@@ -125,7 +125,7 @@ const uint32_t *skip_range = get_nocase_range(c);
 uint32_t skip_start = skip_range[0];
 
 #if PCRE2_CODE_UNIT_WIDTH == 8
-PCRE2_ASSERT(options & PARSE_CLASS_UTF);
+PCRE2_ASSERT(options & (PARSE_CLASS_UTF | PARSE_CLASS_CASELESS_UTF));
 #endif
 
 #if PCRE2_CODE_UNIT_WIDTH == 32
--- a/testdata/testinput10
+++ b/testdata/testinput10
@@ -623,6 +623,9 @@
 /X(\x{e1})Y/i,ucp,replace=>\L$1<,substitute_extended
     X\x{c1}Y
 
+/[a\x{c1}]/iI,ucp
+    \x{e1}
+
 # Without UTF or UCP characters > 127 have only one case in the default locale.
 
 /X(\x{e1})Y/replace=>\U$1<,substitute_extended
--- a/testdata/testoutput10
+++ b/testdata/testoutput10
@@ -1883,6 +1883,14 @@ Subject length lower bound = 1
     X\x{c1}Y
  1: >\xe1<
 
+/[a\x{c1}]/iI,ucp
+Capture group count = 0
+Options: caseless ucp
+Starting code units: A a \xc1 \xe1
+Subject length lower bound = 1
+    \x{e1}
+ 0: \xe1
+
 # Without UTF or UCP characters > 127 have only one case in the default locale.
 
 /X(\x{e1})Y/replace=>\U$1<,substitute_extended


alexdowad:
@zherczeg, I've been going through your char class code and have figured out how some parts of it work, though there is still a lot which I don't understand well. If I understood your comment on #526 well, you suggested that this is what is needed to fix the unexpected behavior change from #474 which I discovered. Did I understand you correctly?

src/pcre2_compile_class.c
```diff
@@ -125,7 +125,7 @@ const uint32_t *skip_range = get_nocase_range(c);
 uint32_t skip_start = skip_range[0];
 
 #if PCRE2_CODE_UNIT_WIDTH == 8
-PCRE2_ASSERT(options & PARSE_CLASS_UTF);
+PCRE2_ASSERT(options & PARSE_CLASS_UTF || options & PARSE_CLASS_CASELESS_UTF);
```
zherczeg, line 128: `options & (PARSE_CLASS_UTF | PARSE_CLASS_CASELESS_UTF)`

zherczeg (changes requested):
I worry that this is not enough. The caseless 's' and 'k' (and turkish 'i' soon) might force an xclass, we need to exclude them in some way.

A check around here could help probably, to terminate processing ranges >256 in 8 bit mode.

https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_compile.c#L6383

zherczeg:
Btw quite good patch for first attempt.

carenas:
> The caseless 's' and 'k' (and turkish 'i' soon) might force an xclass,

  re> /[sk]/iBI,ucp
        Bra
        [KSks]
        Ket
        End
Capture group count = 0
Options: caseless ucp
Starting code units: K S k s
data> 
  re> /[sk]/iBI,ucp,utf
        Bra
        [KSks\x{17f}\x{212a}]
        Ket
        End
Capture group count = 0
Options: caseless ucp utf
Starting code units: K S k s \xc5 \xe2

Without `utf` those characters can't be represented in the class.

zherczeg:
In 8 bit mode, they cannot, but in 16 bit mode, they can.

For me 8 bit + ucp is the same as 16 bit + ucp, except characters > 255 are simply not present, as characters > 65535 are simply not present for 16 bit mode.

It looks like perl has a different approach for this.

carenas:
The main problem I have with our approach is that we are just grossly misinterpreting characters with UCP and not UTF. It is just not the same character.

UCP without UTF makes sense in the 16/32bit libraries were it could represent UCS, but in the 8 bit library we are reading 1 byte and assigning it properties that belong to a different character, just because the ord was the same.

alexdowad:
Carlo, if UCP mode should be 'banned' for the 8-bit library because it doesn't make sense (which I tend to agree with), I suggest a separate issue could be opened to discuss that issue. I don't know the policies followed by PCRE2, but for some other OSS projects which I have worked on before, this would require a deprecation first.

If UCP was banned for the 8-bit library (i.e. using the PCRE2_UCP compile flag or (*UCP) would cause compilation to fail with an error), then all code to support that use case could be removed.

In the meantime, as much as it doesn't make sense, I think that UCP+8bit should be supported, because there is nothing in the documentation saying that it is illegal. The obvious thing for UCP+8bit to do is to treat each byte as a Unicode code point from U+0000 up to U+00FF.

There are a lot of things I don't know well here, so I may be completely off base.

zherczeg:
> The obvious thing for UCP+8bit to do is to treat each byte as a Unicode code point from U+0000 up to U+00FF.

As far as I know this is not happening. We use the pre-compiled tables for many things, and use ucp for other things. The pre-compiled tables depend on locals, and they can be very different from unicode codepoints.

We might need to clean up these things at some point.

PhilipHazel:
> The obvious thing for UCP+8bit to do is to treat each byte as a Unicode code point from U+0000 up to U+00FF.

Yes, and I think that was behind the thinking when PCRE2_UCP was allowed in 8-bit mode. Note this change from 10.35:

> 18. Changes in many areas of the code so that when Unicode is supported and    
> PCRE2_UCP is set without PCRE2_UTF, Unicode character properties are used for 
> upper/lower case computations on characters whose code points are greater than
> 127. 

There must have been a reason for this but I can't remember it. It might have related to 16/32-bit. I think this is a relatively minor issue, because there are not likely to be many (any?) 8-bit use cases where PCRE2_UCP is set without PCRE2_UCP. So we shouldn't spend a lot of time on it. Simplest not to introduce any incompatibilities.

zherczeg:
I agree. Fortunately the new code is perfectly capable of generating the full bitmask for any properties, and handling anything caseless ranges. We just make to ensure that XCLASS is never generated in this case.

https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_compile_class.c#L531

The 8 bit case simply does not worth any optimizations though. The extra checks should be guarded to 8 bit as well.

NWilson:
> The obvious thing for UCP+8bit to do is to treat each byte as a Unicode code point from U+0000 up to U+00FF.

That seems reasonable to me, and it's what I've been assuming.

Note that this aligns well with Latin-1, which is a popular 8-bit encoding that was grandfathered into Unicode to fill the codepoints U+0080 to U+00FF. There is quite a lot of text out there where using 8-bit+UCP for Latin-1 interpretation would be accurate.

alexdowad:
@zherczeg, I've tried to apply your advice, please have a look and tell me if this looks right or not.

src/pcre2_compile.c
```diff
@@ -6479,6 +6485,9 @@ for (;; pptr++)
 #ifdef SUPPORT_WIDE_CHARS  /* Defined for 16/32 bits, or 8-bit with Unicode */
     if ((xclass_props & XCLASS_REQUIRED) != 0)
       {
+      /* We should never generate a (useless) xclass in 8-bit library when UTF flag is false */
+      PCRE2_ASSERT(PCRE2_CODE_UNIT_WIDTH != 8 || utf);
```
zherczeg, line 6489: This is usually an `#if` since it looks better, even if this variant is working.

src/pcre2_compile.c
```diff
@@ -6388,6 +6388,12 @@ for (;; pptr++)
         range = end;
         }
 
+#if PCRE2_CODE_UNIT_WIDTH == 8
+      /* If code unit width is 8 bits, and UCP flag is set, but UTF flag is not, we still
+       * generate cranges, but in that case we should not process any crange > 0xFF,
+       * because it's impossible to encounter code points > 0xFF in the subject string */
+      if (utf)
```
zherczeg, line 6395: Do we need code indentation below? An `if (!utf) range = end;` also works. I have no preference.

zherczeg (changes requested):

src/pcre2_compile.c
carenas: Avoiding conditional indentation can only help making the logic easier to follow IMHO

NWilson:
@alexdowad I picked up your work on this PR - many thanks!

I have tried an alternative approach in PR #644, which is to keep the same rules for generating the "cranges", but instead to do case-expansion in the UCP condition for the U+0000 to U+00FF bitmap.

alexdowad:
@NWilson Thank you very much!
I'll close this PR now.

PR #528, opened 2024-10-15
pcre2test: add support for \N{U+hh...} escapes in subject

carenas:
As a solution that could be used to avoid the ambiguous syntax of the other escape sequences supported by pcre2test, add `\N{U+hh...}`

Includes bugfixes to the handling of `\o{dd...}` and hopefully improved documentation and tests.

### DIFF
--- a/src/pcre2test.c
+++ b/src/pcre2test.c
@@ -963,6 +963,13 @@ static coptstruct coptlist[] = {
 #undef SUPPORT_32
 #undef SUPPORT_EBCDIC
 
+/* Types for the parser, to be used in process_data() */
+
+enum force_encoding {
+  FORCE_NONE,         /* No preference, follow utf modifier */
+  FORCE_RAW,          /* Encode as a code point or error if too wide */
+  FORCE_UTF           /* Encode as a character or error if too wide */
+};
 
 /* ----------------------- Static variables ------------------------ */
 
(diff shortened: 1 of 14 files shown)


src/pcre2test.c
```diff
@@ -7368,27 +7389,37 @@ while ((c = *p++) != 0)
         }
       *q8++ = (uint8_t)c;
       }
+    else
+      {
+      if (c > 0x7fffffff)
+        {
+        fprintf(outfile, "** Character \\N{U+%x} is greater than 0x7fffffff "
+                         "and therefore cannot be encoded as UTF-8\n", c);
```
zherczeg, line 7397: `0x10ffff` is the highest unicode character regardless of encoding.

zherczeg:
Nice patch!

src/pcre2test.c
carenas: Correct, but pcre2test allows encoding up to this original 6 byte maximum with the other escapes, and indeed Perl goes even further.

It becomes handy as  way to input very high values for the 32bit library too, which is why it is also allowed even in non utf mode.

PhilipHazel: True, but pcre2test uses the original UTF-8 encoding scheme (RFC 2279) so as to handle characters up to `0x7fffffff.`

PhilipHazel:
Oops. Forgot to check that the "last updated" line had been updated in pcre2test.1. Perhaps it will get updated again before the next release, so I won't update it specially. I do think it is useful to have dates on documentation.

src/pcre2test.c
carenas: noticed that Perl now [recommends](https://perldoc.perl.org/perlunicode#Comparison-of-%5CN%7B...%7D-and-%5Cp%7Bname=...%7D) using the `\p{name=` property for that which might come out handy, ex: `\p{name=/(SMILING|GRINNING) FACE/}`

carenas: hopefully addressed in the followup

PR #530, opened 2024-10-18
Use less branches when matching xclass in jit

zherczeg:


src/pcre2_jit_compile.c
```diff
@@ -9157,6 +9147,91 @@ while (cc < end_subject)
 return cc;
 }
 
+static void compile_clist(compiler_common *common, PCRE2_SPTR cc, jump_list **backtracks)
+{
+DEFINE_COMPILER;
+const sljit_u32 *other_cases;
+struct sljit_jump *jump;
+sljit_u32 min = 0, max = READ_CHAR_MAX;
+BOOL has_cmov = sljit_has_cpu_feature(SLJIT_HAS_CMOV) != 0;
```
carenas, line 9156: not that it makes any difference now, but wouldn't it be better to do `sljit_has_cpu_feature(SLJIT_HAS_CMOV) == 1`

zherczeg: The non-zero value represents it is supported, higher integer numbers represent partial support.

carenas: And that is why we don't want to use it if it is somehow emulated or slow, right?

zherczeg: It is more like slow in some cases. Reg-reg transfer is always efficient if it is supported in some way. Imm/mem support can be partial.

PR #531, opened 2024-10-18
pcre2_jit_compile: avoid potential wraparound if framesize == 0

carenas:
Additional code would be generated when framesize == 0 that would be dangerous so exempt it and add an assert to catch any situations that could trigger it in the future.

### DIFF
--- a/doc/pcre2syntax.3
+++ b/doc/pcre2syntax.3
@@ -1,4 +1,4 @@
-.TH PCRE2SYNTAX 3 "24 September 2024" "PCRE2 10.45"
+.TH PCRE2SYNTAX 3 "20 October 2024" "PCRE2 10.45"
 .SH NAME
 PCRE2 - Perl-compatible regular expressions (revised API)
 .SH "PCRE2 REGULAR EXPRESSION SYNTAX SUMMARY"
@@ -254,7 +254,7 @@ The recognized classes are:
   RLI         right-to-left isolate
   RLO         right-to-left override
   S           segment separator
-  WS          which space
+  WS          white space
 .
 .
 .SH "CHARACTER CLASSES"
@@ -398,7 +398,7 @@ of the group.
   (?^)            unset imnrsx options
 .sp
 (?aP) implies (?aT) as well, though this has no additional effect. However, it
-means that (?-aP) is really (?-PT) which disables all ASCII restrictions for
+means that (?-aP) also implies (?-aT) and disables all ASCII restrictions for
 POSIX classes.
 .P
 Unsetting x or xx unsets both. Several options may be set at once, and a
@@ -411,10 +411,10 @@ The following are recognized only at the very start of a pattern or after one
 of the newline or \eR sequences or options with similar syntax. More than one
 of them may appear. For the first three, d is a decimal number.
 .sp
-  (*CASELESS_RESTRICT) set PCRE2_EXTRA_CASELESS_RESTRICT when matching
   (*LIMIT_DEPTH=d)     set the backtracking limit to d
   (*LIMIT_HEAP=d)      set the heap size limit to d * 1024 bytes
   (*LIMIT_MATCH=d)     set the match limit to d
+  (*CASELESS_RESTRICT) set PCRE2_EXTRA_CASELESS_RESTRICT when matching
   (*NOTEMPTY)          set PCRE2_NOTEMPTY when matching
   (*NOTEMPTY_ATSTART)  set PCRE2_NOTEMPTY_ATSTART when matching
   (*NO_AUTO_POSSESS)   no auto-possessification (PCRE2_NO_AUTO_POSSESS)
@@ -692,6 +692,6 @@ Cambridge, England.
 .rs
 .sp
 .nf
-Last updated: 24 September 2024
+Last updated: 20 October 2024
 Copyright (c) 1997-2024 University of Cambridge.
 .fi
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -10898,7 +10898,8 @@ if (opcode == OP_ASSERT || opcode == OP_ASSERTBACK)
     {
     JUMPTO(SLJIT_JUMP, backtrack->matchingpath);
     JUMPHERE(brajump);
-    if (framesize >= 0)
+    SLJIT_ASSERT(framesize != 0);
+    if (framesize > 0)
       {
       OP1(SLJIT_MOV, STACK_TOP, 0, SLJIT_MEM1(SLJIT_SP), private_data_ptr);
       add_jump(compiler, &common->revertframes, JUMP(SLJIT_FAST_CALL));
@@ -11799,13 +11800,19 @@ if (bra == OP_BRAMINZERO)
     /* We need to release the end pointer to perform the
     backtrack for the zero-length iteration. When
     framesize is < 0, OP_ONCE will do the release itself. */
-    if (opcode == OP_ONCE && BACKTRACK_AS(bracket_backtrack)->u.framesize >= 0)
+    if (opcode == OP_ONCE)
       {
-      OP1(SLJIT_MOV, STACK_TOP, 0, SLJIT_MEM1(SLJIT_SP), private_data_ptr);
-      add_jump(compiler, &common->revertframes, JUMP(SLJIT_FAST_CALL));
-      OP2(SLJIT_ADD, STACK_TOP, 0, STACK_TOP, 0, SLJIT_IMM, (BACKTRACK_AS(bracket_backtrack)->u.framesize - 1) * sizeof(sljit_sw));
+      int framesize = BACKTRACK_AS(bracket_backtrack)->u.framesize;
+
+      SLJIT_ASSERT(framesize != 0);
+      if (framesize > 0)
+        {
+        OP1(SLJIT_MOV, STACK_TOP, 0, SLJIT_MEM1(SLJIT_SP), private_data_ptr);
+        add_jump(compiler, &common->revertframes, JUMP(SLJIT_FAST_CALL));
+        OP2(SLJIT_ADD, STACK_TOP, 0, STACK_TOP, 0, SLJIT_IMM, (framesize - 1) * sizeof(sljit_sw));
+        }
       }
-    else if (ket == OP_KETRMIN && opcode != OP_ONCE)
+    else if (ket == OP_KETRMIN)
       free_stack(common, 1);
     }
   /* Continue to the normal backtrack. */
@@ -12053,12 +12060,7 @@ while (*cc != OP_KETRPOS)
       add_jump(compiler, &emptymatch, CMP(SLJIT_EQUAL, TMP1, 0, STR_PTR, 0));
 
     if (!zero)
-      {
-      if (framesize < 0)
-        OP1(SLJIT_MOV, SLJIT_MEM1(STACK_TOP), STACK(stacksize - 1), SLJIT_IMM, 0);
-      else
-        OP1(SLJIT_MOV, SLJIT_MEM1(STACK_TOP), STACK(0), SLJIT_IMM, 0);
-      }
+      OP1(SLJIT_MOV, SLJIT_MEM1(STACK_TOP), STACK(0), SLJIT_IMM, 0);
     }
 
   JUMPTO(SLJIT_JUMP, loop);
@@ -13504,7 +13506,7 @@ PCRE2_SPTR ccbegin;
 PCRE2_SPTR ccprev;
 PCRE2_UCHAR bra = OP_BRA;
 PCRE2_UCHAR ket;
-assert_backtrack *assert;
+const assert_backtrack *assert;
 BOOL has_alternatives;
 BOOL needs_control_head = FALSE;
 BOOL has_vreverse;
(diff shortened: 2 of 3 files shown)


carenas:
well, that was embarrasing, and indeed it is being triggered in the newly added assert:

```
frame #5: 0x00000001001916dc pcre2test`compile_bracket_backtrackingpath(common=0x00007ff7bfefcf50, current=0x00006210026d8c18) at pcre2_jit_compile.c:13893:5
   13890	    {
   13891	    SLJIT_ASSERT(opcode == OP_COND || opcode == OP_SCOND);
   13892	    assert = CURRENT_AS(bracket_backtrack)->u.assert;
-> 13893	    SLJIT_ASSERT(assert->framesize != 0);
   13894	    if ((ccbegin[1 + LINK_SIZE] == OP_ASSERT_NOT || ccbegin[1 + LINK_SIZE] == OP_ASSERTBACK_NOT) && assert->framesize >= 0)
   13895	      {
   13896	      OP1(SLJIT_MOV, STACK_TOP, 0, SLJIT_MEM1(SLJIT_SP), assert->private_data_ptr);
p *assert
(assert_backtrack) {
  common = {
    prev = 0x00006210026d8cc0
    simple_backtracks = NULL
    top = 0xbebebebebebebebe
    own_backtracks = 0xbebebebebebebebe
    cc = 0x0000000000000000
  }
  condfailed = NULL
  framesize = 0
  private_data_ptr = 0
  matchingpath = NULL
}
```

zherczeg:
I don't remember but the code has special meaning for framesize < 0. What is the pattern?

zherczeg:
Framesizes usually come form here: https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_jit_compile.c#L2047

The end is also interesting: https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_jit_compile.c#L2231
Looks like it does not really return with 0.

carenas:
> What is the pattern?

```
^(?: \n(?: A| (1? (?=(?<cond>2)?) (1) 2 (?('cond')|3)) )\n(Z)\n)+$
```

zherczeg:
Btw, `13893 SLJIT_ASSERT(assert->framesize != 0);` this is not in the patch.

The code is not wrong, but not nice. The `assert` contains a random value, if the condition is not an assert, but its never resolved if the left side of the logical and fails (short circuit evaluation). The if could be split into two `ifs`, and read the `assert` before the second.

carenas:
I don't think I am fully understanding your advice, but if you were suggesting the original assert that was removed should had been doing instead:

```
SLJIT_ASSERT((ccbegin[1 + LINK_SIZE] == OP_ASSERT_NOT || ccbegin[1 + LINK_SIZE] == OP_ASSERTBACK_NOT) && assert->framesize != 0);
```

I agree with you, but sadly that assert will also cause a crash, not sure why the normal code does not, neither how we get some random `assert` that happens to have a value of 0 for its `framesize`.

zherczeg:
```
if (ccbegin[1 + LINK_SIZE] == OP_ASSERT_NOT || ccbegin[1 + LINK_SIZE] == OP_ASSERTBACK_NOT)
  {
  assert = ;
  ASSERT(...);
  if (...) {}
  }
```
The assert is loaded from a union, and may contain a completely invalid pointer. Reading it can be dangerous.

carenas:
Updated patch, not including the following which still reliably crashes:

```
diff --git a/src/pcre2_jit_compile.c b/src/pcre2_jit_compile.c
index f7aee7c3..68296949 100644
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -13902,6 +13902,7 @@ if (has_alternatives)
         OP2(SLJIT_ADD, STACK_TOP, 0, STACK_TOP, 0, SLJIT_IMM, (assert->framesize - 1) * sizeof(sljit_sw));
         OP1(SLJIT_MOV, SLJIT_MEM1(SLJIT_SP), assert->private_data_ptr, TMP1, 0);
         }
+      else SLJIT_UNREACHABLE();
       }
     JUMPHERE(cond);
     }
```

it should be slightly better than the current state, but after a full nighter on it, I am not getting any closer to a resolution anytime soon.

src/pcre2_jit_compile.c
```diff
@@ -13887,14 +13886,18 @@ if (has_alternatives)
...
+      assert = CURRENT_AS(bracket_backtrack)->u.assert;
+      if (assert->framesize > 1)
+        {
+        OP1(SLJIT_MOV, STACK_TOP, 0, SLJIT_MEM1(SLJIT_SP), assert->private_data_ptr);
+        add_jump(compiler, &common->revertframes, JUMP(SLJIT_FAST_CALL));
+        OP1(SLJIT_MOV, TMP1, 0, SLJIT_MEM1(STACK_TOP), STACK(-2));
+        OP2(SLJIT_ADD, STACK_TOP, 0, STACK_TOP, 0, SLJIT_IMM, (assert->framesize - 1) * sizeof(sljit_sw));
+        OP1(SLJIT_MOV, SLJIT_MEM1(SLJIT_SP), assert->private_data_ptr, TMP1, 0);
+        }
+      else if (assert->framesize == 0) SLJIT_UNREACHABLE();
```
zherczeg, line 13900: put the `SLJIT_ASSERT(assert->framesize > 0);` after the `assert = CURRENT_AS(bracket_backtrack)->u.assert;`.

src/pcre2_jit_compile.c
```diff
@@ -13703,14 +13699,17 @@ if (SLJIT_UNLIKELY(opcode == OP_COND) || SLJIT_UNLIKELY(opcode == OP_SCOND))
   if (ccbegin[1 + LINK_SIZE] >= OP_ASSERT && ccbegin[1 + LINK_SIZE] <= OP_ASSERTBACK_NOT)
     {
     SLJIT_ASSERT(has_alternatives);
-    assert = CURRENT_AS(bracket_backtrack)->u.assert;
```
zherczeg, line 13706: Is there a reason to change this? Optimization? Assert is valid here, since the type is an `OP_ASSERT*`?

src/pcre2_jit_compile.c
```diff
@@ -12053,12 +12054,7 @@ while (*cc != OP_KETRPOS)
       add_jump(compiler, &emptymatch, CMP(SLJIT_EQUAL, TMP1, 0, STR_PTR, 0));
 
     if (!zero)
-      {
-      if (framesize < 0)
-        OP1(SLJIT_MOV, SLJIT_MEM1(STACK_TOP), STACK(stacksize - 1), SLJIT_IMM, 0);
-      else
-        OP1(SLJIT_MOV, SLJIT_MEM1(STACK_TOP), STACK(0), SLJIT_IMM, 0);
-      }
+      OP1(SLJIT_MOV, SLJIT_MEM1(STACK_TOP), STACK(0), SLJIT_IMM, 0);
```
zherczeg, line 12057: Would be good to know why the original code is there.

zherczeg (changes requested):

src/pcre2_jit_compile.c
carenas: It looks like a copy & paste mistake that is also 10 years old, dunno though, I was not the author, it was part of f0855fab5

src/pcre2_jit_compile.c
```diff
@@ -13703,14 +13699,17 @@ if (SLJIT_UNLIKELY(opcode == OP_COND) || SLJIT_UNLIKELY(opcode == OP_SCOND))
...
-    if (assert->framesize >= 0 && (ccbegin[1 + LINK_SIZE] == OP_ASSERT || ccbegin[1 + LINK_SIZE] == OP_ASSERTBACK))
+    if (ccbegin[1 + LINK_SIZE] == OP_ASSERT || ccbegin[1 + LINK_SIZE] == OP_ASSERTBACK)
       {
-      OP1(SLJIT_MOV, STACK_TOP, 0, SLJIT_MEM1(SLJIT_SP), assert->private_data_ptr);
-      add_jump(compiler, &common->revertframes, JUMP(SLJIT_FAST_CALL));
-      OP1(SLJIT_MOV, TMP1, 0, SLJIT_MEM1(STACK_TOP), STACK(-2));
-      OP2(SLJIT_ADD, STACK_TOP, 0, STACK_TOP, 0, SLJIT_IMM, (assert->framesize - 1) * sizeof(sljit_sw));
-      OP1(SLJIT_MOV, SLJIT_MEM1(SLJIT_SP), assert->private_data_ptr, TMP1, 0);
+      assert = CURRENT_AS(bracket_backtrack)->u.assert;
+      if (assert->framesize > 1)
```
zherczeg, line 13705: An `SLJIT_ASSERT(assert->framesize != 0)` could be added here as well.

src/pcre2_jit_compile.c
carenas: sorry; I was sleep deprived and really meant to use `PCRE2_UNREACHABLE()` here and only as a last resort to catch any more of those heisenbugs before removing the asserts completely, since they IMHO no longer make sense because the code explicitly filters out any framesize below 1.

will put the asserts back in their "normal" place as you requested though, but still think we are better without them.

src/pcre2_jit_compile.c
carenas: consistency with the other "similar" condition, it also made the intention clearer and obvious that no assert was needed.

will move it back.

src/pcre2_jit_compile.c
```diff
@@ -10898,7 +10898,8 @@ if (opcode == OP_ASSERT || opcode == OP_ASSERTBACK)
     {
     JUMPTO(SLJIT_JUMP, backtrack->matchingpath);
     JUMPHERE(brajump);
-    if (framesize >= 0)
+    SLJIT_ASSERT(framesize != 0);
+    if (framesize > 1)
```
zherczeg, line 10902: framesize == 1 should be valid

src/pcre2_jit_compile.c
zherczeg: I recommend to sleep, this is definitely not urgent.

src/pcre2_jit_compile.c
carenas: but the code it generates is useless, no?

src/pcre2_jit_compile.c
carenas: I know, it was just the frustration of seeing such a simple change cause crashes, and the shame of not caching those (twice) in local testing; plus my unfamiliarity with the code that required a lot of reading.

luckily I found another 10 year old "bug", so definitely worth it.

src/pcre2_jit_compile.c
zherczeg: Why? Not just `>= 2` cases are valid, the `== 1` is valid as well. `>= 0` can be turned to `> 0` or `>= 1`

carenas: Is doing something like this worth it then?

```
diff --git a/src/pcre2_jit_compile.c b/src/pcre2_jit_compile.c
index ca13cd76..dc78f52a 100644
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -10904,7 +10904,8 @@ if (opcode == OP_ASSERT || opcode == OP_ASSERTBACK)
       OP1(SLJIT_MOV, STACK_TOP, 0, SLJIT_MEM1(SLJIT_SP), private_data_ptr);
       add_jump(compiler, &common->revertframes, JUMP(SLJIT_FAST_CALL));
       OP1(SLJIT_MOV, TMP1, 0, SLJIT_MEM1(STACK_TOP), STACK(-2));
-      OP2(SLJIT_ADD, STACK_TOP, 0, STACK_TOP, 0, SLJIT_IMM, (framesize - 1) * sizeof(sljit_sw));
+      if (framesize > 1)
+        OP2(SLJIT_ADD, STACK_TOP, 0, STACK_TOP, 0, SLJIT_IMM, (framesize - 1) * sizeof(sljit_sw));
       OP1(SLJIT_MOV, SLJIT_MEM1(SLJIT_SP), private_data_ptr, TMP1, 0);
       }
     set_jumps(backtrack->common.own_backtracks, LABEL());
```

zherczeg:
So <= 1 is invalid?

carenas:
Not sure I understand the context for your question, but no, there are at least two negative values of framesize that are valid, eventhough they were AFAIK never supported by the code I am changing.

zherczeg:
The code looks ok

doc/pcre2syntax.3
```diff
@@ -411,10 +411,10 @@ The following are recognized only at the very start of a pattern or after one
 of the newline or \eR sequences or options with similar syntax. More than one
 of them may appear. For the first three, d is a decimal number.
 .sp
-  (*CASELESS_RESTRICT) set PCRE2_EXTRA_CASELESS_RESTRICT when matching
   (*LIMIT_DEPTH=d)     set the backtracking limit to d
   (*LIMIT_HEAP=d)      set the heap size limit to d * 1024 bytes
   (*LIMIT_MATCH=d)     set the match limit to d
+  (*CASELESS_RESTRICT) set PCRE2_EXTRA_CASELESS_RESTRICT when matching
```
zherczeg, line 417: Why this is moved down?

zherczeg (approved):
Just one question

doc/pcre2syntax.3
carenas: because it is preceded by "For the first three, d is a decimal number"

NWilson: Oops! I alphabetized them, without noticing that problem. Thank you for fixing.

PR #539, opened 2024-10-26
Clear padding bytes

zherczeg:
Caught by OSS-Fuzz


zherczeg:
It is probably not a god idea to land this patch, because I might need to redesign the code.

PR #540, opened 2024-10-27
Move character lists data before the byte code in a pattern

zherczeg:
This patch moves the character lists to the end of the pattern, and only a reference is stored. This way repeating, repeat detection, and serializing works without modifications. Also, for repeating character sets inside brackets can save a lot of space. The disadvantage is that an extra variable is needed to be maintained, which has a very low overhead, but still an overhead. My first idea was storing pointers, but that is not serialization friendly, although not that hard to implement.


zherczeg:
Btw the original problem was that repeating character classes can break alignment rules unfortunately. It would not be bad if `(PATTERN){n,m}` could be implemented without repeating the `(PATTERN)`.

zherczeg:
I tried various variations for this code, but the engine always have a feature which blocked them. In the end the character list data is stored before the byte code. The cost is an extra argument for xclass, and an extra member in the real code. Computing the byte code start is simplified.

zherczeg:
Context: a fuzzer got a comparing uninitialized memory error, when a subpattern is repeated. The reason is that space for alignemnt was not initialized. Then it turned out that repeating character lists may break alignment, so they need to be moved out from the byte code and stored elsewhere, and only a reference to them is stored in the byte code, which can be repeated. So this patch fixes a complicated and serious error in the new code.

NWilson:
Interesting. I would have thought the simplest thing would simply be to avoid adding alignment padding entirely, and simply have READ8, READ16, READ32 macros to do the unaligned reads using bit-ops. Just let the uint32 values lie wherever they fall naturally in the bytecode, and read them in a safe way.

Compilers are also super-good at optimising this pattern: `uint32_t value; memcpy(&value, unaligned_address, 4)`. It's a good alternative to `#define READ32(addr)  (addr[0] << 24) | (addr[1] << 16) | (addr[2] << 8) | addr[3]`.

NWilson:
It's extra complexity to move data out of the main sequence of bytecode. And it's extra complexity to introduce any weird requirements on the bytecode, that would prevent it being memmove'd around. The bytecode is a PCRE2_SPTR, so its contents should be read and used, simply assuming that it has the PCRE2_SPTR base alignment. Adding padding so that certain slices inside it have stricter alignment then means that the code can't be copied and moved, which is unfortunate.

zherczeg:
The bytecode header structure always needed at least 32 bit alignment, so you cannot move the byte code to any location. Unaligned accesses might be costly.

src/pcre2_compile.c
```diff
@@ -10954,13 +10924,36 @@ if (length > MAX_PATTERN_SIZE)
...
 no longer be possible because nowadays we limit the maximum value of
 cb.names_found and cb.name_entry_size. */
 
-re_blocksize = CU2BYTES(length +
-  (PCRE2_SIZE)cb.names_found * (PCRE2_SIZE)cb.name_entry_size);
+re_blocksize =
+  CU2BYTES((PCRE2_SIZE)cb.names_found * (PCRE2_SIZE)cb.name_entry_size);
+
+#if defined SUPPORT_WIDE_CHARS
+if (cb.char_lists_size > 0)
```
carenas, line 10945: not really making a difference but I think `cb.char_lists_size != 0` is clearer

carenas (approved):

zherczeg:
I have updated the patch

NWilson:
This will probably make a merge conflict with my PR #523, so it's a race to see which is merged first!

zherczeg:
@PhilipHazel will decide. I have no problem with doing the merge.

src/pcre2_compile.c
```diff
@@ -6541,94 +6541,63 @@ for (;; pptr++)
...
-            }
+          *code++ = (PCRE2_UCHAR)(XCL_LIST | cranges->char_lists_types);
 #endif
-          memcpy(char_buffer,
+
+          cb->char_lists_size += char_lists_size;
+
+          memcpy((uint8_t*)cb->start_code - cb->char_lists_size,
             (uint8_t*)(cranges + 1) + cranges->char_lists_start,
             char_lists_size);
```
NWilson, line 6590: You could memset the padding bytes to 0xff or something. It appears you're just leaving whatever junk is in there from malloc?

If you do initialize... then you'll get predictable runtime behaviour (if something reads from it by mistake). If you don't initialize... then you'll get nice valgrind warnings.

src/pcre2_compile.c
```diff
@@ -6541,94 +6541,63 @@ for (;; pptr++)
...
-            code[0] |= 2u << XCL_ALIGNMENT_SHIFT;
-            char_buffer += 2;
-            }
+          *code++ = (PCRE2_UCHAR)(XCL_LIST | cranges->char_lists_types);
 #endif
-          memcpy(char_buffer,
+
+          cb->char_lists_size += char_lists_size;
+
+          memcpy((uint8_t*)cb->start_code - cb->char_lists_size,
```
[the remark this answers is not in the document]
zherczeg: Yes. This way we can reuse the code start value (part of the match block) as the base (list end) value. I could reverse the order (start from max to 0), but I don't see any gain from it. This way the generator is a bit less complex.

src/pcre2_compile.c
zherczeg: Make tools happy. Ok, I will check.

NWilson: The best-of-both would be a memset to 0xff and also an `ifdef SUPPORT_VALGRIND ... VALGRIND_MAKE_MEM_NOACCESS`.

src/pcre2_compile.c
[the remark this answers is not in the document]
NWilson: OK. Maybe a comment on the layout would help.

src/pcre2_compile.c
carenas: I prefer doing only the VALGRIND_MAKE_MEM_NOACCESS, so any mistaken reads can be detected, after all those valgrind warnings are a feature!, the memset is just a workaround IMHO.

zherczeg: I have added a `((uint16_t*)data)[-1] = 0xffff;` to set all bits. Btw valgrind only complains if you check the memory. Just copying it is not a problem.

src/pcre2_compile.c
```diff
@@ -6541,94 +6541,76 @@ for (;; pptr++)
...
+          code += LINK_SIZE;
 
-          /* Padding. */
           if ((char_lists_size & 0x2) != 0)
             {
-            code[0] |= 2u << XCL_ALIGNMENT_SHIFT;
-            char_buffer += 2;
-            }
+            cb->char_lists_size += 2;
+            /* Make tools happy by setting memory data. */
```
carenas, line 6608: s/memory data/memory padding data/

src/pcre2_compile.c
NWilson: Just to check I understand - the data is 16-bit and/or 32-bit packed values, but we want it to be 32-bit aligned. So the `data` variable is always 16-bit aligned, and if there is padding, it will be exactly 2 bytes. So the cast `(uint16_t*)data` is doing a correctly-aligned write of the two padding bytes.

That all looks great, thanks!

zherczeg: Yes. The character list data is a stream of 16 bit values followed by a stream of 32 bit values. Both stream can be empty. Regardless, the final data is considered to be a stream of 32 bit values, and there might be an unused 16 bit value in the beginning sometimes.

zherczeg:
I changed the extra 16 bit setting to debug only, and added some asserts.

src/pcre2_compile.c
```diff
@@ -10206,13 +10206,36 @@ if (length > MAX_PATTERN_SIZE)
   goto HAD_CB_ERROR;
   }
```
NWilson, line 10207: This block here could be in an "#else" block for the code you added.

Sorry, really minor nit.

NWilson (approved):
Great, this look like it should work well

src/pcre2_compile.c
zherczeg: I am not sure I follow. I wanted to be in the safe side here, since `MAX_PATTERN_SIZE - length >= 0` after the check and never underflows.

PR #542, opened 2024-10-30
Fix PCRE2_DEBUG for multi-configuration builds

NWilson:
On Windows (MSVC) and Mac (Xcode), one config.h file is shared by both Debug and Release builds.

So, I think we should instead pass `-DPCRE2_DEBUG` on the commandline instead of including it in the config file, and set it to true only for Debug builds.

This ensures that CMake users on Windows & Mac will get their assertions checked automatically.


carenas (changes requested):
Technically the same is true of the "Ninja Multi-Config" generator that is available in more platforms and will be a good option if added to the CI (ex: changing the Perl job in dev.yml)

This will also require further changes to `CMakeFiles.txt` so it can generate a working `pcre2_test.sh` to be able to complete `ctest`.

NWilson:
Oh yes, the CTest integration has been done a little unusually. 

On Linux I usually run the pcre2 tests with "ninja test" or "make test" (from the Cmake build directory - I never use autoconf if I can avoid it). On windows I use "Cmake --build . --target RUN_TESTS". So those variations work at least. 

I can test with the multi config ninja and make sure it works.

CMakeLists.txt
```diff
@@ -689,6 +690,12 @@ ENDIF(MSVC AND PCRE2_STATIC_RUNTIME)
 
 ADD_DEFINITIONS(-DHAVE_CONFIG_H)
 
+if(PCRE2_DEBUG STREQUAL "IfDebugBuild")
+  ADD_COMPILE_DEFINITIONS("$<$<CONFIG:Debug>:PCRE2_DEBUG>")
```
carenas, line 694: requires cmake 3.12, so need to update required version

NWilson: Thanks, done

NWilson:
I've added support for "Ninja Multi-Config" and tested on Linux & Windows. The Windows support was already working (due to the support in pcre2_test.bat) but I had to tweak pcre2_test.sh a little bit. Not too hard, all works nicely.

CMakeLists.txt
```diff
@@ -103,9 +103,8 @@
 # 2023-01-15 Carlo added C99 as the minimum required
 # 2023-08-06 PH added support for setting variable length lookbehind maximum
 
-# Increased minimum to 3.5 to workaround deprecated backward compatibility
-# since 3.27.
-cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
+# Increased minimum to 3.15 to allow use of generator expressions.
+cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
```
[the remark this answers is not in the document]
NWilson: I used `string(REPEAT ...` which apparently requires 3.15.

Versions of Ubuntu older than 20.04 are in "extended security maintenance". I can't see users of these platforms (>5 years old) needing to use the latest-and-greatest PCRE2; they can always upgrade their CMake from the Kitware website if needed.

I'm happy to remove the `string(REPEAT ...)` line and go back down to 3.12 if you prefer, which is a 6-year-old version.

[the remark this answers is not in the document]
carenas: > I'm happy to remove the `string(REPEAT ...)` line and go back down to 3.12 if you prefer

It would be also nice to update the "perl" `dev.yml` CI job to use this generator too so we have  way to prevent any future regressions.

carenas:
> Mac (Xcode)

Tested with Xcode 15.2

NWilson:
Interesting. I don't have a Mac or Xcode to test with.

On Windows, I can run tests with either `cmake --build . --target test` (when using ninja Multi-config), or `cmake --build . --target RUN_TESTS` (when using native Visual Studio cmake backend); or alternatively using `ctest -C Debug`. However just `ctest` on its own doesn't work.

CMakeLists.txt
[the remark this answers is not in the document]
NWilson: Hmm. That's using `cmake` from the Ubuntu repositories. We can set the build back to using the `ubuntu-20.04` GitHub runner, which would be testing with CMake 3.16.

Or we can go lower-level and really force it to use 3.12 by installing the tarball manually from KitWare.

Do we _really_ need to work with 3.12? We should be covered even on pretty old platforms by 3.15/16.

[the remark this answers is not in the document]
carenas: > That's using `cmake` from the Ubuntu repositories

Not really, it is running in a Docker that is based on Debian.  Debian 10 came with cmake 3.13, but I like your idea of installing our own to make sure.

I have to admit I was more concerned about regressing the multi-config fixes for PCRE2_DEBUG though.

> Do we _really_ need to work with 3.12?

no, but we need to make sure that whatever we said we support actually works.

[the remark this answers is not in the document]
NWilson: Oh I see, it's using the `perl:devel` Docker container. That appears to be currently based on Debian bookworm (12) which includes CMake 3.25.

The `perl:devel-bullseye` container uses Debian bookwork (11) which bundles CMake 3.18.

I'll install an older CMake version from KitWare, to ensure we're supporting all the older targets.

NWilson:
Updated commit description:

---

Fix PCRE2_DEBUG for multi-configuration builds

* Add support for Ninja Multi-Config generator on all platforms
* Ensure one build job is using CMake 3.15
* Provide a build type (Debug or RelWithDebInfo) on all CMake builds
* Ensure PCRE2_DEBUG=ON is set for: dev.yml builds which use the
  RelWithDebInfo configuration

.github/workflows/build.yml
```diff
@@ -77,7 +77,7 @@ jobs:
           submodules: true
 
       - name: Configure
-        run: cmake -DPCRE2_SUPPORT_JIT=ON -DPCRE2_BUILD_PCRE2_16=ON -DPCRE2_BUILD_PCRE2_32=ON -DCMAKE_OSX_ARCHITECTURES='arm64;x86_64' -DCMAKE_C_FLAGS='-Wall -Wextra' -DCMAKE_COMPILE_WARNING_AS_ERROR=ON -B build
+        run: cmake -DPCRE2_SUPPORT_JIT=ON -DPCRE2_BUILD_PCRE2_16=ON -DPCRE2_BUILD_PCRE2_32=ON -DCMAKE_OSX_ARCHITECTURES='arm64;x86_64' -DCMAKE_C_FLAGS='-Wall -Wextra' -DCMAKE_COMPILE_WARNING_AS_ERROR=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -B build
```
NWilson, line 80: I think all the jobs in `build.yml` should be using RelWithDebInfo (for the jobs using CMake of course). And, we want PCRE2_DEBUG to be left to its default, which will disable the assertions.

This mirrors the configuration which we expect customers and downstream packages to be using.

.github/workflows/dev.yml
```diff
@@ -70,6 +70,8 @@ jobs:
   wasp:
     name: No JIT
```
NWilson, line 71: I changed this Ubuntu job to use an explicit CMake version, and ninja. It's still using no-JIT and clang.

.github/workflows/dev.yml
```diff
@@ -107,16 +116,16 @@ jobs:
           submodules: true
 
       - name: Configure
-        run: cmake -DPCRE2_SUPPORT_JIT=ON -DPCRE2_BUILD_PCRE2_16=ON -DPCRE2_BUILD_PCRE2_32=ON -DPCRE2GREP_SUPPORT_CALLOUT_FORK=OFF -DPCRE2_DEBUG=ON -DPCRE2_NEWLINE=ANYCRLF -DPCRE2_STATIC_PIC=ON -DPCRE2_STATIC_RUNTIME=ON -DPCRE2_SUPPORT_BSR_ANYCRLF=ON -DCMAKE_COMPILE_WARNING_AS_ERROR=ON -DCMAKE_VERBOSE_MAKEFILE=ON -B build -A Win32
+        run: cmake -DPCRE2_SUPPORT_JIT=ON -DPCRE2_BUILD_PCRE2_16=ON -DPCRE2_BUILD_PCRE2_32=ON -DPCRE2GREP_SUPPORT_CALLOUT_FORK=OFF -DPCRE2_DEBUG=ON -DPCRE2_NEWLINE=ANYCRLF -DPCRE2_STATIC_PIC=ON -DPCRE2_STATIC_RUNTIME=ON -DPCRE2_SUPPORT_BSR_ANYCRLF=ON -DPCRE2_DEBUG=ON -DCMAKE_COMPILE_WARNING_AS_ERROR=ON -DCMAKE_VERBOSE_MAKEFILE=ON -B build -A Win32
 
       - name: Build
-        run: cmake --build build
+        run: cmake --build build --config RelWithDebInfo
```
NWilson, line 122: All the CMake jobs in `dev.yml` are now using an explicit build type. Most of them use RelWithDebInfo because it makes the tests run so much faster; but at least one is testing and using the Debug configuration.

NWilson (commented):
OK, I'm happy with this and ready for it to be merged, as long as Carlo's happy.

.github/workflows/build.yml
carenas: The thinking was that we use the default of the generator used since they are meant to match what a developer runs localy, but you probably have a lot more experience with cmake as a developer than any of us.

NWilson: I never run CMake without specifying a CMAKE_BUILD_TYPE. Some generators (like the Makefiles backend) just do weird things, and generate configuration that nobody really wants.

I guess we could have one example of that in the dev.yml... but it would only apply on Linux anyway, and we're expecting most Linux users to be using autoconf instead. So I'd just simplify things, and only test on actual proper CMake configurations (Debug, Release, RelWithDebInfo).

CMakeLists.txt
```diff
@@ -1147,81 +1159,100 @@ IF(PCRE2_HEAP_MATCH_RECURSE)
...
   MESSAGE(STATUS "")
   MESSAGE(STATUS "PCRE2-${PCRE2_MAJOR}.${PCRE2_MINOR} configuration summary:")
   MESSAGE(STATUS "")
-  MESSAGE(STATUS "  Install prefix .................. : ${CMAKE_INSTALL_PREFIX}")
-  MESSAGE(STATUS "  C compiler ...................... : ${CMAKE_C_COMPILER}")
-  MESSAGE(STATUS "  C compiler flags ................ : ${CMAKE_C_FLAGS}${cfsp}${CMAKE_C_FLAGS_${buildtype}}")
+  MESSAGE(STATUS "  Install prefix .................... : ${CMAKE_INSTALL_PREFIX}")
+  MESSAGE(STATUS "  C compiler ........................ : ${CMAKE_C_COMPILER}")
+  if (CMAKE_CONFIGURATION_TYPES)
+    FOREACH (config IN LISTS CMAKE_CONFIGURATION_TYPES)
```
carenas, line 1169: nitpick: can we use lowercase foreach?

CMakeLists.txt
```diff
@@ -1147,81 +1159,100 @@ IF(PCRE2_HEAP_MATCH_RECURSE)
 ENDIF(PCRE2_HEAP_MATCH_RECURSE)
 
 IF(PCRE2_SHOW_REPORT)
-  STRING(TOUPPER "${CMAKE_BUILD_TYPE}" buildtype)
-  IF (CMAKE_C_FLAGS)
-    SET(cfsp " ")
-  ENDIF(CMAKE_C_FLAGS)
```
carenas, line 1153: I think it is easier to have this set only once, no?

if it was moved to avoid polluting the rest of the script the maybe just unset it after we are done?, not sure how cmake defines scope.

RunTest
```diff
@@ -923,13 +917,30 @@ for bmode in "$test8" "$test16" "$test32"; do
...
     checkresult $? heap-$bits ""
   fi
 
 # End of loop for 8/16/32-bit tests
 done
 
+
+# ------ Special EBCDIC Test -------
+
+if [ $doebcdic = yes ]; then
```
carenas, line 930: minor nitpick: Philip seems to like a space between "]" and ";"

PhilipHazel: I may be wrong, but I think at one point in the historical past the space was necessary - or I got confused about it - so I got into the habit of separating [ and ] from their surroundings. (Might have been different in different shells?) Anyway, I'm not dogmatic about it.

CMakeLists.txt
NWilson: OK, done.

But someone really needs to just run an autoformatter over the file, it's totally inconsistent at the moment. I can do that in a follow-up.

CMakeLists.txt
NWilson: Done.

RunTest
NWilson: Done.

CMakeLists.txt
carenas: the inconsistency is by design, see #115

.github/workflows/build.yml
carenas: I realize I misread the context.

My comment was bout jobs in dev.yml, for build.yml we want to mirror what end users would do for production, so while I think you are correct `RelWithDebInfo` is valid `Release` might be better (which will also test -O3 vs -O2 with gcc/clang)

CMakeLists.txt
NWilson: Well 23 instances of "if" lowercase and 124 instances of "IF" can't be by design. The formatting should at least be consistent. It's not important though, given it has absolutely no functional impact.

Making it more idiomatic CMake would be nice in future, there are some good ideas in that thread.

.github/workflows/build.yml
NWilson: Interesting. I've personally always considered "RelWithDebInfo" to be the best for actual production releases.

To start with, no-one should be shipping binaries without having the debug info stored somewhere. Linux distributions always build with debug info, and ship the ELF+DWARF binaries in special "-debug" variant of the package, then run `strip` to publish the ordinary packages. That way, when users report a crash, you can symbolise it. Commercial vendors do the same thing (on all the companies and projects I've worked on) - build with symbols, store them, and ship the stripped binaries.

I also (personally) dislike O3 because it does too much inlining with GCC. Binary size for a C++ app can double, with no noticeable performance increase. For apps that are bigger than 10-20MB binary size, `-Os` is often the fastest in fact, because `-O3` does so much loop unrolling it forces the code out of the L1 cache.

But yes, I totally acknowledge that people ship with `-O2`, `-O3`, or `-Os`.

Here's what I'll do: change `build.yml` to use `Release` (`-O3`), and in `build.yml` let's make sure we run the unit tests for each of `-O2`, `-O3`, and `-Os`.

CMakeLists.txt
carenas: Sorry, I meant "by design" ironically, since that file has been hacked without any for way too long IMHO.

Indeed, I have to admit I was a little surprised no one mentioned taking over maintainer-ship of at least this file as part of our previous discussions.

NWilson:
Carlo has asked for a change to the flags we test with (`-O3` rather than `-O2` in build.yml). I'll do that quickly, then it'll be ready.

carenas (approved):

NWilson:
OK, ready to merge.

I have encountered two issues while fiddling for a while with different compile options:
* The GCC build doesn't pass with `-O0 -Werror`.
* The Clang ASAN build doesn't pass with `-O0` either. There's a stack overflow in the recursive JIT function scan_prefix, which uses a little more stack space in `-O0` mode, and the recursion overflows the stack. The failing pattern is `/X?(R||){3335}/` in `testinput1`.

These are pre-existing issues, so I've left them. We can address in a future PR.

zherczeg (approved):
LGTM

carenas:
> LGTM

meant to hit the merge button, but failed somehow?

zherczeg:
LGTM is looks good to me.

PR #543, opened 2024-10-30
Use assert() for unreachable assertions

NWilson:
Using assert rather than abort will give a (very slightly) better error message, which includes the containing function name.

Only affects Debug builds.

### DIFF
--- a/src/pcre2_util.h
+++ b/src/pcre2_util.h
@@ -45,17 +45,20 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #ifdef PCRE2_DEBUG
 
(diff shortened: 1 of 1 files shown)


carenas (approved):
It would be probably nicer to refactor it slightly so there is only one `#include <assert.h>` though

NWilson:
Good thinking. I pulled out the `#include <assert.h>` and moved it up a few lines.

zherczeg (approved):
LGTM

zherczeg:
@carenas is this patch ok now?

carenas:
Yes, merge it please; @zherczeg

PR #544, opened 2024-10-30
Fix the lookahead after [\d or [[:posix] to skip whitespace

NWilson:


src/pcre2_compile.c
```diff
@@ -2498,6 +2498,53 @@ return -1;
...
+static PCRE2_SPTR
+skip_class_whitespace(PCRE2_SPTR ptr, PCRE2_SPTR ptrend, uint32_t options)
+{
+uint32_t c;
+
+while (ptr < ptrend)
+  {
+  c = *ptr;
+
+  if (c == CHAR_BACKSLASH)
```
zherczeg, line 2526: For me a `switch(*ptr)` looks nicer.

testdata/testinput1
```diff
@@ -6721,4 +6715,16 @@ $/x
 /65 00 64/hex
     e\0d
 
+/[[:digit:]-   ]/xx
+    1
+    -
+\= Expect no match
+    z
```
carenas, line 6722: better check for "space" (`\ \`) as not matching

carenas (changes requested):

testdata/testinput1
NWilson: Great, done

src/pcre2_compile.c
NWilson: OK, done.

carenas (approved):

NWilson:
Actually... there's a bug here (pre-existing), which I should fix.

If you do `[\d-&&[foo]]` in ALT_EXTENDED mode, then the hyphen after the `\d` is not a range.

Either:
* Fix the lookahead, so that it looks for a `]` after the hyphen, OR an operator or `[` in ALT_EXTENDED mode. But that's getting quite a convoluted lookahead
* Get rid of the lookahed; just store a boolean flag that says "we saw a \d...\p or [:posix:] escape" and check that flag when we actually reach the hyphen.

I prefer the second. The lookahead was already buggy (and not matching Perl) so it just seems brittle and unmaintainable, especially as I now realise that I failed to update the lookahead properly for the ALT_EXTENDED case.

NWilson:
OK, it's much cleaner now. I've reverted the "improved" lookahead, and removed the lookahead entirely.

Adding a little extra state (on the existing `class_range_state` variable) allows us to track this in a much cleaner way, in my opinion, and we end up matching Perl's behaviour more closely for edge cases like `[\d-\E]`, as was the goal.

Ready for review again.

carenas (approved):
nice work

PR #546, opened 2024-10-31
Various code improvements for eclasses

zherczeg:
I think the new code is quite nice and works nicely. I did some code reworks to improve the operations. There is no byte code optimizations. @NWilson please check.

### DIFF
--- a/src/pcre2_compile_class.c
+++ b/src/pcre2_compile_class.c
@@ -44,6 +44,31 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include "pcre2_compile.h"
 
+/* Checks the allowed tokens at the end of a class structure in debug mode.
+When a new token is not processed by all loops, and the token is equals to
+a) one of the cases here:
+   the compiler will complain about a duplicated case value.
+b) none of the cases here:
+   the loop without the handler will stop with an assertion failure. */
+
+#ifdef PCRE2_DEBUG
+#define CLASS_END_CASES(meta) \
+  default: \
+  PCRE2_ASSERT((meta) <= META_END); \
+  /* Fall through */ \
+  case META_CLASS: \
+  case META_CLASS_NOT: \
+  case META_CLASS_EMPTY: \
+  case META_CLASS_EMPTY_NOT: \
+  case META_CLASS_END: \
+  case META_ECLASS_OR: \
+  case META_ECLASS_AND: \
+  case META_ECLASS_SUB:
+#else
+#define CLASS_END_CASES(meta) \
+  default:
+#endif
+
 #ifdef SUPPORT_WIDE_CHARS
 
 /* Heapsort algorithm. */
@@ -319,15 +344,14 @@ append_non_ascii_range(uint32_t options, uint32_t *buffer)
 }
 
 static size_t
-parse_class(uint32_t *ptr, const uint32_t *end_ptr, uint32_t options,
-  uint32_t *buffer)
+parse_class(uint32_t *ptr, uint32_t options, uint32_t *buffer)
 {
 size_t total_size = 0;
 size_t size;
 uint32_t meta_arg;
 uint32_t start_char;
 
-while (ptr < end_ptr)
+while (TRUE)
   {
   switch (META_CODE(*ptr))
     {
@@ -395,8 +419,8 @@ while (ptr < end_ptr)
       /* Character literal */
       ptr++;
       break;
-    default:
-      PCRE2_ASSERT(*ptr < META_END);
+    CLASS_END_CASES(*ptr)
+      if (*ptr >= META_END) return total_size;
       break;
     }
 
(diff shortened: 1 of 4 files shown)


src/pcre2_compile_class.c
```diff
@@ -1266,32 +1215,14 @@ while (pptr < end_ptr)
...
+      meta = *(pptr++);
+      break;
 
-    if (META_CODE(meta) != META_ESCAPE)
-      {
-#ifdef DEBUG_SHOW_PARSED
-      fprintf(stderr, "** Unrecognized parsed pattern item 0x%.8x "
-                      "in character class\n", meta);
-#endif
-      *errorcodeptr = ERR89;  /* Internal error - unrecognized. */
```
zherczeg, line 1292: I think assertions are better for this, so I removed this functionality. If it is desired, it can be added to the call sites of this function.

NWilson (commented):
Great, I like the changes overall.

It's going to conflict with my WIP branch where I'm doing the optimizations you requested. Don't duplicate the work I've started on that - unless you want to let me know what you're tackling.

I'm on holiday with friends for a few days now, so I won't open another PR before next Tues or Weds.

src/pcre2_compile_class.c
```diff
@@ -1828,7 +1761,7 @@ code += 32 / sizeof(PCRE2_UCHAR);
 
 DONE:
 *pcode = code;
-return TRUE;
+return pptr - 1;
```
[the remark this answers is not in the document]
zherczeg: Usually setting a tmp value used only in an assertion is hard to do nicely. The alternative is using a `while (ptr < end)` to enforce they run to the same point. If it is ok for you, I will do that.

Background: the meta codes for character sets and operators/brackets are different, so it is safe to process them and return with the last non character set meta code. Later we could pass a list of starting locations, and a single range could be constructed from them.

Example:
`[A[B&&C]D]` could be reorganized to `[AD[B&&C]]` internally.

We could just use an operator list and starting meta code pointers.

From the ECLASS above, we construct the following small list:
```
ptr-to-A
ptr-to-B
ptr-to-C
OP_AND
OP_OR
ptr-to-D
OP_OR
```

Then an optimizer converts it to:
```
ptr-to-A and ptr-to-D -> the range is constructed from the combination of these two items.
ptr-to-B
ptr-to-C
OP_AND
OP_OR
```

zherczeg:
I don't want to work on any class optimizations at this point. The task is yours as we discussed. I will share my ideas on the topic, but it is your decision how to use them. This patch tries to improve existing code, nothing more.

NWilson:
Thank you Zoltan! I'm very grateful for your contributions, you are welcome to work on the code of course. I only wanted to make sure we don't duplicate work by mistake.

zherczeg:
I did the requested changes except one. I could not find a nice way to pass around the class end without many defines. Maybe we can do that in a follow up patch if we find some nice solution.

zherczeg:
I found a very different way to do the requested debug check. I have added a macro, `CLASS_END_CASES`, which has a list of the allowed class token endings. With this macro, it is not possible that one loop handles a case, and the other does not:
```
/* Checks the allowed tokens at the end of a class structure in debug mode.
When a new token is processed by one loop, and not the other, and the token is
a) equals to one of the cases here: the compiler will complain
   about a duplicated case value.
b) equals to none of the cases here: the loop without the handler will stop
   with an assertion failure. */
```

src/pcre2_compile_class.c
```diff
@@ -44,6 +44,30 @@ POSSIBILITY OF SUCH DAMAGE.
...
+When a new token is not processed by all loops, and the token is equals to
+a) one of the cases here:
+   the compiler will complain about a duplicated case value.
+b) none of the cases here:
+   the loop without the handler will stop with an assertion failure. */
+
+#ifdef PCRE2_DEBUG
+#define CLASS_END_CASES(meta) \
+  default: \
+  PCRE2_ASSERT((meta) <= META_END); \
```
NWilson, line 57: I'm surprised the compiler lets you write a statement that falls through to another case, with the "Falls through" comment that's enforced elsewhere.

src/pcre2_compile_class.c
```diff
@@ -1828,7 +1773,7 @@ code += 32 / sizeof(PCRE2_UCHAR);
 
 DONE:
 *pcode = code;
-return TRUE;
+return pptr - 1;
```
NWilson, line 1776: Is there a tiny bit of code smell to returning `ptr - 1` here? The last META code we saw might be a two-item one (such as POSIX + code, or BIGVALUE + codepoint). So, the -1 doesn't point to anything meaningful.

It appears you've done it, because the existing callers take the return pointer and always add a fixed +1 to it, and it's simpler in some ways to leave the callers as-is.

Just a suggestion. If the code works out neater this way, that's fine. Maybe the function should have a note in its doc at the top that you must not take the return value and treat it the start of a META code.

NWilson (approved):
Great! I like the changes. There's one thing I don't understand (pptr - 1) but it might be fine, I just can't follow what's happening from looking at the diff.

Those macros look really effective for aligning the for loops, thank you!

carenas (commented):
less code is always better

src/pcre2_compile_class.c
zherczeg: It reverts the ++ here.

https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_compile_class.c#L1141

We read the token which terminates the loop. We don't process the token, just read its type.

src/pcre2_compile_class.c
zherczeg: I forgot it. Thank you for noticing it. Compiler can complain about it with the suitable compiler options, probably something is missing.

NWilson: I guess the "multi-config" CI build doesn't include a debug build, with those options.

Issue #497, opened 2024-09-26
Pattern /[^\D\P{Nd}]/utf matches to \x{1d7cf}


zherczeg:
This test is in testinput5:

```
  re> /[^\D\P{Nd}]/B,utf
------------------------------------------------------------------
        Bra
        [^\x00-/:-\xff\P{Nd}]
        Ket
        End
------------------------------------------------------------------
data> \x{1d7cf}
 0: \x{1d7cf}
```
Currently this pattern matches to \x{1d7cf}

Since \D is in ascii mode (ucp is not enabled), `/[\D]/utf` matches to anything not 0-9. That should include \x{1d7cf}. This looks true:

```
  re> /[\D\P{Nd}]/B,utf
------------------------------------------------------------------
        Bra
        [\x00-/:-\xff\P{Nd}\x{100}-\x{10ffff}]
        Ket
        End
------------------------------------------------------------------
data> \x{1d7cf}
 0: \x{1d7cf}
```

Note: /[\P{Nd}]/utf does not match to \x{1d7cf}

Summary: both `/[^\D\P{Nd}]/utf` and `/[\D\P{Nd}]/utf` matches to \x{1d7cf}. Obviously a character class and its negated form cannot match to the same character, and I think the first one is incorrect. My newest code changes this pattern to no-match, and I wanted to discuss it.

NWilson:
This is a very interesting case!

There's something related here - which is the behaviour of `\P{...}` with `/i` modifier.

They specifically changed the behaviour of `[^\P{...}]` in ECMAScript. If you run the regex using `/[^\P{...}]/iu` vs using `/[^\P{...}]/iv` in JavaScript you get different results. (Note the change from "/u Unicode" to "/v Next-gen Unicode".)

They discussed this change here: https://github.com/tc39/proposal-regexp-v-flag/issues/30

(Note - there's a lot of confusion and irrelevant chatter on that thread. Most of it just irrelevant.)

## How character classes work

* Take the characters and ranges, and union them to form a "set" of characters
* Things like `\D` mean "the complement/inverted-set" for whatever characters are in the `\d` set. This is applied immediately, at compile-time.
* Similarly for `\p{...}` and `\P{...}`.
* Then, `^` is applied last to complement/invert the set of matched characters.

With the `/i` modifier:
* Each character that appears in the set is "case-folded".
  * So `[A]` means → add `case_fold('A')` to the set of matched characters
[...]

PR #550, opened 2024-11-06
Add additional tests for negated /[^\P{Lu}]/i behaviour

NWilson:
PCRE2's behaviour matches Perl currently, and Python's regex module (the standard re module doesn't have \P{...}).

However, JavaScript behaviour changed recently (when they added the new 'v' flag; the old 'u' flag's behaviour is unchanged), so this case is worth additional testing.

Raised in issue #497.

### DIFF
--- a/testdata/testinput4
+++ b/testdata/testinput4
@@ -2913,4 +2913,40 @@
 /c3 b1/hex,utf
     \N{U+00F1}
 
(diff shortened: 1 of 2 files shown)


carenas (approved):

zherczeg (approved):
LGTM

PR #554, opened 2024-11-08
Improve repeated matching of newlines

zherczeg:
The tricky part is that newlines can be two character long, and can start at the middle of \r\n.

I also accidentally found an old bug and fixed it. It requires a somewhat complicated pattern.

### DIFF
--- a/src/pcre2_jit_test.c
+++ b/src/pcre2_jit_test.c
@@ -360,6 +360,7 @@ static struct regression_test_case regression_test_cases[] = {
 	{ MU, A, 0, 0, "(?P<size>\\d+)m|M", "4M" },
 	{ M, PCRE2_NEWLINE_CRLF, 0, 0, "\\n?.+#", "\n,\n,#" },
 	{ 0, A, 0, 0, "<(\\w+)[\\s\\w]+id>", "<br><div id>" },
+	{ MU, A, 0, 0, "([a-z]{0,3}c;)+", "ccccc;c;cc;ccc;cccccccccccccccc;" },
 
 	/* Bracket repeats with limit. */
 	{ MU, A, 0, 0, "(?:(ab){2}){5}M", "abababababababababababM" },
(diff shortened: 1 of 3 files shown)


src/pcre2_jit_test.c
```diff
@@ -481,6 +482,7 @@ static struct regression_test_case regression_test_cases[] = {
 	{ MU, A, 0, 0, "\\R{3,}\n", "\r\n\r\n\nab\n\n\n\r\r\n\n" },
 	{ MU, A, 0, 0, "\\R{0,3}\n", "\r\n\r\n\r\n\n" },
 	{ MU, A, 0, 0, "\\R{0,3}\n", "\r\n\r\n\r\n\r" },
+	{ MU, A, 0, 0, "(\\R{0,3}\n;)+", "\r\n\r\n\r\n\r\n\n;\n;\n\n;\n\n\n;\n\n\n\n\n;" },
```
NWilson, line 485: Should the failing test case be added to one of the testdata tests as well?

zherczeg: These tests are usually not duplicated. Both run at testing time. Btw. the other is the falling test.

carenas (approved):

src/pcre2_jit_test.c
carenas: might be better to add a description (for users) to the ChangeLog though, eventhough there are likely several other entries missing there already.

zherczeg: Ok. Fortunately it is incredibly difficult to trigger it in release code. This is probably the reason why nobody found it before. The effect is incorrect match, no crashes.

NWilson: I assumed the test with newlines was the one causing the failure, given the PR title! I see now it's a broader fix.

Anyway, all fine by me.

zherczeg: The story is that I moved newlines to be handled by a different code path, and because the new code was not fully finished (some optimizations were not enabled by that time), it triggered an error. Then I realized, that I can trigger it with another way, which was present in the old code. The test triggers the issue only in debug mode, it would be far more difficult to trigger it in release mode (but possible).

zherczeg: The new code uses a fixed stack (does not scale with the number of newlines), and I can optimize locals even more in the future. I could do it before, but I was lazy, and thought repeated \R is quite rare, and not worth the effort. Now thanks to eclasses, it is. Only extended unicode sequences (\X) are unoptimized now, but they are so complicated that I don't want to touch them.

PR #556, opened 2024-11-10
Remove fixed locals for iterators

zherczeg:
Not a trivial patch, fuzzing will be useful. Fixes a bug in caseful matching as well.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -107,6 +107,10 @@ well as interpreting \b and \v as characters.
 24. Fixed a bug in JIT affecting greedy bounded repeats. The upper limit of
 repeats inside a repeated bracket might be incorrectly checked.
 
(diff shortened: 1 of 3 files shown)


ChangeLog
```diff
@@ -107,6 +107,10 @@ well as interpreting \b and \v as characters.
 24. Fixed a bug in JIT affecting greedy bounded repeats. The upper limit of
 repeats inside a repeated bracket might be incorrectly checked.
 
+25. Fixed a bug in JIT affecting caseful matching of backreferences. When
+utf is disabled, and dupnames is enabled, casesless matching is used instead
```
NWilson, line 111: When you say "caseless matching **is** used" do you mean, "caseless matching was used, and the behaviour has now been fixed to be caseful"?

NWilson:
This is another of those JIT patches I can't really review.

PR #557, opened 2024-11-11
Fix the -O0 configuration and its warnings

NWilson:
The CI configuration was using `CC='gcc -O0`. However, this doesn't work! Because CFLAGS was unset, `./configure` would set CFLAGS to `-O2`, resulting in a commandline of `gcc -O0 ... -O2 ...`.

Fixing this reveals that GCC generates additional warnings when in `-O0` mode. Presumably it doesn't do dead-code pruning, and so generates additional areas in provable-unreachable sections of code (which would be pruned in `-O2`).

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -21,7 +21,7 @@ jobs:
         run: ./autogen.sh
 
       - name: Configure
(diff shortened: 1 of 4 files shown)


zherczeg (approved):
LGTM

PR #560, opened 2024-11-12
Non-recursive scan prefix in JIT

### LINKED ISSUE #558: scan_prefix has no effective recursion limit
I run this:

```
./pcre2test -q -8 -jit
  re> /X?(R||){3335}/
```

I get a crash due to stack overflow; excessive consumption of system stack, due to recursive calls to `scan_prefix`. The problem is more easily reproducible if you use a configuration which either reduces the stack size, or uses more than normal.

For example, I found this problem while running the unit tests with Clang ASAN, which bloats the binary with extra stuff, so recursion chews up more stack space.

However, a completely default build of PCRE2 can cause the segfault if you do e.g. `ulimit -s 512` (512KiB) to reduce the stack down to something smaller than the Linux default.

zherczeg:
The algorithm is rewritten. It should cover more cases and non-recursive.

Testing is difficult though. When it covers cases that it should not, the matching should fail, so that can be tested at some level. However, the opposite, when it does not cover cases, that it should, the effect is just slower matching.

Fixes #558 since it is non-recursive.

### DIFF
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -5670,11 +5670,38 @@ if (last)
   chars->last_count++;
 }
 
-static int scan_prefix(compiler_common *common, PCRE2_SPTR cc, fast_forward_char_data *chars, int max_chars, sljit_u32 *rec_count)
+/* Value can be increased if needed. Patterns
+such as /(a|){33}b/ can exhaust the stack.
+
+Note: /(a|){29}b/ already stops scan_prefix()
+because it reaches the maximum step_count. */
+#define SCAN_PREFIX_STACK_END 32
+
+/*
+Scan prefix stores the prefix string in the chars array.
+The elements of the chars array is either small character
+sets or "any" (count is set to 255).
+
+Examples (the chars array is represented by a simple regex):
+
+/(abc|xbyd)/ prefix: /[ax]b[cy]/ (length: 3)
+/a[a-z]b+c/ prefix: a.b (length: 3)
+/ab?cd/ prefix: a[bc][cd] (length: 3)
+/(ab|cd)|(ef|gh)/ prefix: [aceg][bdfh] (length: 2)
+
+The length is returned by scan_prefix(). The length is
+less than or equal than the minimum length of the pattern.
+*/
+
+static int scan_prefix(compiler_common *common, PCRE2_SPTR cc, fast_forward_char_data *chars)
 {
-/* Recursive function, which scans prefix literals. */
+fast_forward_char_data *chars_start = chars;
+fast_forward_char_data *chars_end = chars + MAX_N_CHARS;
+PCRE2_SPTR cc_stack[SCAN_PREFIX_STACK_END];
+fast_forward_char_data *chars_stack[SCAN_PREFIX_STACK_END];
+sljit_u8 next_alternative_stack[SCAN_PREFIX_STACK_END];
 BOOL last, any, class, caseless;
-int len, repeat, len_save, consumed = 0;
+int stack_ptr, step_count, repeat, len, len_save;
 sljit_u32 chr; /* Any unicode character. */
 sljit_u8 *bytes, *bytes_end, byte;
 PCRE2_SPTR alternative, cc_save, oc;
@@ -5687,11 +5714,44 @@ PCRE2_UCHAR othercase[1];
 #endif
 
 repeat = 1;
+stack_ptr = 0;
+step_count = 10000;
 while (TRUE)
   {
-  if (*rec_count == 0)
+  if (--step_count == 0)
     return 0;
-  (*rec_count)--;
+
+  SLJIT_ASSERT(chars <= chars_start + MAX_N_CHARS);
+
+  if (chars >= chars_end)
+    {
+    if (stack_ptr == 0)
+      return (int)(chars_end - chars_start);
+
+    --stack_ptr;
+    cc = cc_stack[stack_ptr];
+    chars = chars_stack[stack_ptr];
+
+    if (chars >= chars_end)
+      continue;
+
+    if (next_alternative_stack[stack_ptr] != 0)
+      {
+      /* When an alternative is processed, the
+      next alternative is pushed onto the stack. */
+      SLJIT_ASSERT(*cc == OP_ALT);
+      alternative = cc + GET(cc, 1);
+      if (*alternative == OP_ALT)
+        {
+        SLJIT_ASSERT(stack_ptr < SCAN_PREFIX_STACK_END);
+        SLJIT_ASSERT(chars_stack[stack_ptr] == chars);
+        SLJIT_ASSERT(next_alternative_stack[stack_ptr] == 1);
+        cc_stack[stack_ptr] = alternative;
+        stack_ptr++;
+        }
+      cc += 1 + LINK_SIZE;
+      }
+    }
 
   last = TRUE;
   any = FALSE;
@@ -5768,9 +5828,17 @@ while (TRUE)
 #ifdef SUPPORT_UNICODE
     if (common->utf && HAS_EXTRALEN(*cc)) len += GET_EXTRALEN(*cc);
 #endif
-    max_chars = scan_prefix(common, cc + len, chars, max_chars, rec_count);
-    if (max_chars == 0)
-      return consumed;
+    if (stack_ptr >= SCAN_PREFIX_STACK_END)
+      {
+      chars_end = chars;
+      continue;
+      }
+
+    cc_stack[stack_ptr] = cc + len;
+    chars_stack[stack_ptr] = chars;
+    next_alternative_stack[stack_ptr] = 0;
+    stack_ptr++;
+
     last = FALSE;
     break;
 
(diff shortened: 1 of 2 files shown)


NWilson:
I didn't have time to look at this yesterday; but I will today.

src/pcre2_jit_compile.c
```diff
@@ -5670,11 +5670,17 @@ if (last)
   chars->last_count++;
 }
 
-static int scan_prefix(compiler_common *common, PCRE2_SPTR cc, fast_forward_char_data *chars, int max_chars, sljit_u32 *rec_count)
+#define SCAN_PREFIX_STACK_END 32
+
+static int scan_prefix(compiler_common *common, PCRE2_SPTR cc, fast_forward_char_data *chars)
 {
-/* Recursive function, which scans prefix literals. */
```
NWilson, line 5675: It took me an embarrassingly long time to work out what a "prefix literal" meant. I _think_ I get it now: you just want to determine the set of characters which could start the pattern, or return an empty set if there's an unknowable set?

So for example, `(x|yz)` the "prefixes" are `x, y` (not z). And in `(x...)?z` the prefixes are `x, z`.

The old code is completely messing with my head - the flow of control is mega-convoluted to follow. So I should perhaps stop worrying about proving whether the behaviour is unchanged, and instead just review whether the new code implements the prefix concept accurately.

src/pcre2_jit_compile.c
```diff
@@ -5888,29 +5969,32 @@ while (TRUE)
...
 
     default:
-    return consumed;
+    chars_end = chars;
+    continue;
     }
 
+  SLJIT_ASSERT(chars < chars_end);
+
   if (any)
```
NWilson, line 5990: Here's a suggestion: as far as I can see, two out of three of the uses of recursion in the original code were completely unnecessary.

There was OP_QUERY, OP_CRSTAR/OP_CRQUERY, and OP_BRA which do recursion. The first two don't need it!

For example, OP_QUERY could be:

```
switch (*cc) {

case OP_QUERY:
add_prefix_chars(*cc, caseless, repeat=1);
cc += len;
continue;
```

Something like that. Basically, take the blocks of code underneath the switch (`if (any) ...` and `if (class) ...` and the code for handling a single character). Lift those out into functions. Then you can process the OP_QUERY immediately, and just carry on to the next opcode, no need for any funky recursion tricks.

NWilson (commented):
Sorry, I've spent quite a lot of time on this, and typed a lot of silly questions. I'm just going to pause reviewing for the day.

zherczeg:
I will try to explain the algorithm using examples. The prefix is a simplified pattern, which has a fixed length, and it is a string of small character sets and dots in regexp terms.

Example:
`/(abc|xbyd)/` prefix is `[ax]b[cy]` (3 characters long)
`/a[a-z]b+c/` prefix is `a.b` (3 characters long) (I decided that `+` may introduce too many combinations, and not worth to continue. Never proved that it is true or false, just my feelings.)
`/ab?cd/` prefix is `a[bc][cd]` (3 characters long)
`/(ab|cd)|(ef|gh)/` prefix is `[aceg][bdfh]` (2 characters long)

Why this prefix is collected: jit can generate a simd code, which can search `...[ab]...[cd]` patterns very fast (currently disabled on Windows though). So in the prefix we search for maximum two character sets (prefer single characters if possible), where the two sets has no common character. In my experiences the chance of such pairs are very low in a real text, so an `ab` is quite rare, but a `  ` (double space) is frequent. We even prefer an ` [ax]` over `  `, although the latter is "simpler".

The scan_prefix scans the pattern in a simple loop. However, if an `|` or `character?` is encountered, we save its location on the stack, and continue to the ending ')'. This way we can parse `/a(bc|cb|xy)e/` as `a[bcx][cby]e`.

I hope this explains how the algorithm works. In jit, everything is rather complex unfortunately. This gives its speed, but quite challenging for understanding.

Let me know if you need more explanation.

src/pcre2_jit_compile.c
```diff
@@ -5687,11 +5693,44 @@ PCRE2_UCHAR othercase[1];
 #endif
 
 repeat = 1;
+stack_ptr = 0;
+rec_count = 10000;
```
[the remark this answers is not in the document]
zherczeg: Updated

NWilson:
Ouch! I knew my understanding of the code's intentions didn't match what the code was doing, but I couldn't work out its purpose from just reading the loop.

That's really helpful explanation. I should be able to match that behaviour to the code now.

zherczeg:
I have added the examples as a comment to the function. Let me know if more changes are needed.

src/pcre2_jit_compile.c
```diff
@@ -5670,11 +5670,17 @@ if (last)
   chars->last_count++;
 }
 
-static int scan_prefix(compiler_common *common, PCRE2_SPTR cc, fast_forward_char_data *chars, int max_chars, sljit_u32 *rec_count)
+#define SCAN_PREFIX_STACK_END 32
```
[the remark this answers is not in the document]
zherczeg: It is a random value. Should be big enough in practice. The stack usage is quite low in most cases, but:
`/(a|){33}b/` or `/(a?){33}/`can exhaust it. In this case no prefix is computed.

zherczeg:
I plan to land this patch soon, then #559, then the rest. Please let me know if this needs more changes.

src/pcre2_jit_compile.c
```diff
@@ -5687,11 +5714,44 @@ PCRE2_UCHAR othercase[1];
...
+  if (--step_count == 0)
     return 0;
-  (*rec_count)--;
+
+  SLJIT_ASSERT(chars <= chars_start + MAX_N_CHARS);
+
+  if (chars >= chars_end)
+    {
+    if (stack_ptr == 0)
+      return chars_end - chars_start;
```
NWilson, line 5729: This `chars_end - chars_start` probably needs an explicit cast to int to silence warnings (even though we know it's always < 12 the compiler won't work it out always).

src/pcre2_jit_compile.c
```diff
@@ -5888,29 +5990,32 @@ while (TRUE)
...
-
-      consumed++;
-      if (--max_chars == 0)
-        return consumed;
       chars++;
       }
-    while (--repeat > 0);
+    while (--repeat > 0 && chars < chars_end);
 
     repeat = 1;
```
NWilson, line 6020: There are several places that do `repeat = 1`, and I think you need to remember to do that in all the places that do a `continue`. That seems error-prone. Couldn't we move all the `repeat = 1` assignments up to the start of the loop, where last/caseless/etc are initialized?

src/pcre2_jit_compile.c
```diff
@@ -5980,19 +6095,18 @@ while (TRUE)
...
-      consumed++;
-      if (--max_chars == 0)
-        return consumed;
       chars++;
       }
-    while (--repeat > 0);
+    while (--repeat > 0 && chars < chars_end);
 
+    repeat = 1;
     switch (*cc)
```
NWilson, line 6103: You could merge this `switch` with the one just above. The switch statement could then set `last`, and do the `if (last) chars_end = chars` assignment at the end. Then the code to handle CRSTAR/CRQUERY/CRRANGE wouldn't be split across two switches, and it would look more similar to the OP_START/OP_QUERY/OP_RANGE code in the main switch.

Totally optional though, and not needed to merge.

src/pcre2_jit_compile.c
```diff
@@ -5687,11 +5714,44 @@ PCRE2_UCHAR othercase[1];
...
+      /* When an alternative is processed, the
+      next alternative is pushed onto the stack. */
+      SLJIT_ASSERT(*cc == OP_ALT);
+      alternative = cc + GET(cc, 1);
+      if (*alternative == OP_ALT)
+        {
+        SLJIT_ASSERT(stack_ptr < SCAN_PREFIX_STACK_END);
+        cc_stack[stack_ptr] = alternative;
+        chars_stack[stack_ptr] = chars;
+        next_alternative_stack[stack_ptr] = 1;
```
NWilson, line 5749: The assignments to `chars_stack` and `next_alternative_stack` should be setting the same values that we just read.

That's just an observation. I don't have a preference between keeping the redundant assignments as they are here, or converting them to a debug assertion.

NWilson (approved):
Sweet, I've spent another hour or two mentally stepping through the old code, and the new code, and I understand them both fully now.

They both appear correct, and should have the same behaviour.

Sorry to take a while. This was a good learning opportunity for me, since it made me learn more about how some of the OP codes worked, for some that I hadn't needed to look at yet.

src/pcre2_jit_compile.c
zherczeg: Unfortunately not.

https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_jit_compile.c#L5883

For combined opcodes, we just record the repeat, and continue the main loop. Then we process the actual opcode with the recorded repeat.

E.g.: OP_TYPEEXACT followed by OP_ANY.

src/pcre2_jit_compile.c
zherczeg: I haven't noticed that. This is a quite good observation.

src/pcre2_jit_compile.c
zherczeg: Maybe a goto could be used here, but I try to not use goto when possible.

PhilipHazel:
> Most of these choices are made based on how it is easier / faster to process them in the interpreter. Philip probably could tell more about it.

When I first wrote PCRE in 1997, Perl regular expressions were a lot simpler than they are today. I'm actually quite gratified to know that we've managed to keep on upgrading the code over nearly 30 years. The original set of opcodes was invented "off the top of my head" but has been changed over the years. For example, at first I had a "string" opcode rather than OP_CHAR. That had its problems; this is from the ChangeLog for 2.07:

> 7. Fixed bug: a zero repetition after a literal string (e.g. /abcde{0}/) was   
> causing the entire string to be ignored, instead of just the last character. 

It was not until release 5.0 (2004) that the string op was abolished in favour of OP_CHAR. (Reading the PCRE1 ChangeLog can be quite instructive.) Note that JIT support didn't come along until 2011 (release 8.20), a little bit after I got rid of tracking options at runtime (there used to be OP_OPT) and introduced OP_CHARI and other caseful/caseless pairs in 8.13. 

So yes, it was the interpreter that influenced all these early decisions.

@zherczeg is also correct in saying that not much attention has been paid to the exact values of error offsets.

PR #562, opened 2024-11-13
WIP: slightly improve substitutions

carenas:
Avoid at least one crash introduced with recent changes to substitute code as well as clarify what the expected offset value should be when overflowing the provided buffer.

While at it, make sure that the returned string is always NUL terminated, and do some minor cleanup.

NOTE: at least truncation is wrong so posting mainly as a FYI with the hopes someone else might give it some love

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -826,12 +826,12 @@ for (;;)
     return;
 
     case META_CAPTURE:
-    fprintf(stderr, "META_CAPTURE %d", meta_arg);
+    fprintf(stderr, "META_CAPTURE %u", meta_arg);
     break;
 
     case META_RECURSE:
     GETOFFSET(offset, pptr);
-    fprintf(stderr, "META_RECURSE %d %zd", meta_arg, offset);
+    fprintf(stderr, "META_RECURSE %u %zu", meta_arg, offset);
     break;
 
     case META_BACKREF:
@@ -839,7 +839,7 @@ for (;;)
       offset = cb->small_ref_offset[meta_arg];
     else
       GETOFFSET(offset, pptr);
-    fprintf(stderr, "META_BACKREF %d %zd", meta_arg, offset);
+    fprintf(stderr, "META_BACKREF %u %zu", meta_arg, offset);
     break;
 
     case META_ESCAPE:
@@ -847,7 +847,7 @@ for (;;)
       {
       uint32_t ptype = *pptr >> 16;
       uint32_t pvalue = *pptr++ & 0xffff;
-      fprintf(stderr, "META \\%c %d %d", (meta_arg == ESC_P)? CHAR_P:CHAR_p,
+      fprintf(stderr, "META \\%c %u %u", (meta_arg == ESC_P)? CHAR_P:CHAR_p,
         ptype, pvalue);
       }
     else
@@ -914,7 +914,7 @@ for (;;)
     case META_LOOKAHEAD_NA: fprintf(stderr, "META (*napla:"); break;
     case META_SCRIPT_RUN: fprintf(stderr, "META (*sr:"); break;
     case META_KET: fprintf(stderr, "META )"); break;
-    case META_ALT: fprintf(stderr, "META | %d", meta_arg); break;
+    case META_ALT: fprintf(stderr, "META | %u", meta_arg); break;
 
     case META_CLASS: fprintf(stderr, "META ["); break;
     case META_CLASS_NOT: fprintf(stderr, "META [^"); break;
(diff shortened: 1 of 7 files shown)


zherczeg:
Could you update this patch?

NWilson:
Could you also split out the unrelated changes into their own PRs? It should be quick to do, and would let us merge the cosmetic changes and behaviour-altering changes in their own commits.

src/pcre2_compile.c
```diff
@@ -1113,7 +1112,7 @@ in the decoded tables. */
 
 if ((code->flags & PCRE2_DEREF_TABLES) != 0)
   {
-  ref_count = (PCRE2_SIZE *)(code->tables + TABLES_LENGTH);
+  PCRE2_SIZE *ref_count = (PCRE2_SIZE *)(code->tables + TABLES_LENGTH);
```
NWilson, line 1115: Personally, I'm very happy with these changes.

I know Philip likes the old style of defining variables high up, at the top of a scope, and with a blank line after variable definitions.

But I don't see any benefit to having variables available for use, but not yet initialised. Much better to define & initialise at the same time (safer).

The compiler will hoist all the variables up to the top anyway (it will bump the stack pointer just once at the start of a block, rather than bump the stack pointer multiple times, when it sees a new variable).

src/pcre2_substitute.c
```diff
@@ -942,7 +940,7 @@ do
           GETCHARINCTEST(ch, subptr);
           if (forcecase != 0)
             {
-            if (mcontext->substitute_case_callout)
+            if (mcontext != NULL && mcontext->substitute_case_callout != NULL)
```
NWilson, line 943: Thanks for fixing this! This could be its own PR, since it's an actual defect.

src/pcre2_substitute.c
```diff
@@ -1194,8 +1192,17 @@ if (!replacement_only)
...
+
+if (lengthleft-- > 0) ++buff_offset;
+else
+  {
+  if ((suboptions & PCRE2_SUBSTITUTE_OVERFLOW_LENGTH) == 0) goto NOROOM;
+  overflowed = TRUE;
+  extra_needed++;
+  lengthleft = 0;
+  }
+if (!overflowed || lengthleft == 0) buffer[buff_offset] = 0; else extra_needed++;
```
NWilson, line 1205: Could you explain why you need to inline the CHECKMEMCPY here for, for the trailing NUL?

What was wrong before? Do you want the returned string to be NUL-terminated, even if the function returns an error?

testdata/testinput2
```diff
@@ -4220,6 +4220,37 @@
...
 
 /abc/
+    123abc123\=replace=XYZ
+    123abc123\=replace=[10]XYZ
+\= Expect error
+    123abc123\=replace=[9]XYZ
+    123abc123\=substitute_overflow_length,replace=[9]XYZ
+    123abc123\=substitute_overflow_length,replace=[6]XYZ
+    123abc123\=substitute_overflow_length,replace=[1]XYZ
+    123abc123\=substitute_overflow_length,replace=[0]XYZ
```
NWilson, line 4230: I'd be curious to run these new tests against the old code, just to see which (if any) of the test outputs have changed.

src/pcre2_compile.c
PhilipHazel: Partly it's because I'm a dinosaur from the age when one had to define variables like that, but partly also I find it makes it easier when looking back up some code to find where a variable is defined. However, I am not going to try to impose my own preferences on the future. I can certainly see the advantage of always initializing at definition time. So please don't worry about me too much.

src/pcre2_substitute.c
carenas: > Do you want the returned string to be NUL-terminated, even if the function returns an error?

Correct, I found the way this function behaves strange and the fact that it will return non NUL terminated strings on overflow, potentially risky.

testdata/testinput2
carenas: None does, but could add split the tests in a "setup" patch of its own so it will be obvious

src/pcre2_compile.c
carenas: Funny is that this change is still valid C89 code and the main motivation wasn't to go against Philip's advice of defining variables at the beginning of blocks, but just reducing the scope of this variable to where it was actually needed/used.

Since we have at least one CI job with `-Wshadow` and I wanted to minimize churn didn't rename the variable to reflect its "temp" holder (might be even optimized out) status.

NWilson:
@carenas I don't know if any of these changes are still relevant? I will close if there's no desire to rebase and update this.

PR #563, opened 2024-11-13
Support small offsets in capture group list

zherczeg:
It will conflicts with some other open PRs.


src/pcre2_compile.c
```diff
@@ -4301,9 +4314,17 @@ while (ptr < ptrend)
                 errorcode = ERR15;
                 goto FAILED;
                 }
-              *parsed_pattern++ = meta ? META_SCS_NEXT_NUMBER : META_SCS_NUMBER;
-              offset = (PCRE2_SIZE)(ptr - cb->start_pattern - 2);
-              PUTOFFSET(offset, parsed_pattern);
+
+              if (meta >= 2 && (next_offset - offset) <= 0xffff)
```
NWilson, line 4318: How can you have `next_offset - offset > 0xffff`? Would that be if you have `(0000000000...0000001)` as the number in read_number, or does `(      'name'     )` allow skipping whitespace? Having >65000 characters of junk inside the scan_substring syntax seems perverse. You _could_ just reject `next_offset - offset > 0xffff` as invalid syntax, rather than having the non-RO versions of the META codes just to support this obscure case.

src/pcre2_compile.c
```diff
@@ -9611,8 +9658,6 @@ for (; *pptr != META_END; pptr++)
     case META_COND_NUMBER:
     case META_COND_RNAME:
     case META_COND_RNUMBER:
-    case META_SCS_NEXT_NAME:
-    case META_SCS_NEXT_NUMBER:
     pptr += 1 + SIZEOFFSET;
     nestlevel++;
```
NWilson, line 9662: Is this a sneaky unrelated bugfix?

carenas (approved):
updating HACKING probably a good idea, though

src/pcre2_compile.c
zherczeg: There is a bit more complex problem here. The `compile_branch` function is recursive. Hence the initial offset of the initial meta code only known by the caller. The callee always use a non-RO variant first to initialize the offset, then use the RO variants.

Avoiding this behavior is difficult:
1) Doing it in the caller level: a loop could process the continuation meta codes. This would require a loop, which duplicates a lot of logic, or a sub-function.
2) Doing it in the callee level: the first opcode should go back in the meta code to find the starting offset. Looks a bit dangerous/fragile.
3) The meta code has an optional offset: the `length[meta]` structure has a fixed size for each element. This is also difficult to change.

I don't want to add space/comment support to the list at this point, because conditional blocks do not support them as well. This can be changed though.

I don't worry too much about the number of meta codes. We should never run out of them.

Anyway, if you have a better idea, let me know.

NWilson: I see, I see. Hmmm. I mainly find it a bit distasteful to see things handled in a non-uniform way.

We already have META_SCS_NAME and META_SCS_NEXT_NAME to handle the first and subsequent conditions separately; now you're creating META_SCS_NAME, META_SCS_N_NAME_RO, and META_SCS_N_NAME to handle the first three conditions differently!

How about this:

META_SCS simply starts the assertion. Contains no record of name or number. But does the recursive call into compile_regex(), which will consume the following META codes.

META_OFFSET always follows META_SCS, and will be consumed to set the offset inside the nested compile_regex().

Finally META_SCS_NUMBER and META_SCS_NAME search for the name/number and emit the appropriate OP_CREF/DNCREF. Both of these would embed a relative offset. If the offset is >0xffff then emit a parse error in the pre-parser.

The benefit would be: don't embed the first name/number into the SCS start, but instead handle the first one in the same way as the following ones. Also, we wouldn't then need more than one version of the anything.

zherczeg: Ok, this would be a bigger rework then. I very much like this idea though. We have more characters here than `(?(` conditional blocks, since `(*scs:(` is at least 7 characters long. START(1), OFFSET(3), NUMBER/NAME(2) is still just 5.

I would still not worry about the > 0xffff case, since `META_OFFSET` could be repeated if needed.

src/pcre2_compile.c
```diff
@@ -4301,10 +4297,18 @@ while (ptr < ptrend)
...
+
+              if (meta == 0 || (next_offset - offset) >= 0x10000)
+                {
+                *parsed_pattern++ = META_OFFSET;
+                PUTOFFSET(next_offset, parsed_pattern);
+                offset = next_offset;
+                }
+
+              *parsed_pattern++ =
+                META_SCS_NUMBER | (uint32_t)(next_offset - offset);
```
NWilson, line 4309: It looks like the old code was writing out the offset of the `ptr` after `read_number`, so pointing to the comma following, whereas the new code is writing out the offset of the start of the number.

I don't especially mind, but I've been trying to make error messages point after the syntax error, in the style of Perl which says "error was left of <-- HERE".

src/pcre2_compile.c
```diff
@@ -4287,11 +4279,15 @@ while (ptr < ptrend)
             }
 
           ptr++;
+          *parsed_pattern++ = META_SCS;
           /* Temporary variable, zero in the first iteration. */
           meta = 0;
```
NWilson, line 4284: You've copy-pasted the code for writing out the META_OFFSET twice, once in each branch (number/name). So maybe do it a third time by getting rid of the `meta` variable, and hoist out the initial write of the META_OFFSET to the top?

src/pcre2_compile.c
```diff
@@ -6078,17 +6101,19 @@ for (;; pptr++)
     case META_COND_RNUMBER:   /* (?(Rdigits) */
     case META_COND_NAME:      /* (?(name) or (?'name') or ?(<name>) */
     case META_COND_RNAME:     /* (?(R&name) - test for recursion */
-    case META_SCS_NAME:       /* (*scan_substring:'name') or (*scan_substring:(<name>)) */
-    case META_SCS_NEXT_NAME:  /* More names for scan substring. */
-    bravalue = meta == META_SCS_NAME ? OP_ASSERT_SCS : OP_COND;
+    case META_SCS_NAME:       /* Name of scan substring */
+    bravalue = OP_COND;
```
NWilson, line 6105: Does it matter that we set `bravalue` to something a bit meaningless, for the `META_SCS_NAME` case?

It looks like `bravalue`is only used in `GROUP_PROCESS:`, and the code here for META_SCS_NAME has already done that, and every branch below which does the jump there checks for META_SCS_NAME to make sure it's special-cased.

NWilson (approved):
Great, it all looks correct and clean to me. I like it!

It makes you wonder, perhaps we should do the same for META_COND_NAME and META_COND_NUMBER (split out the group recursion into a common META_COND_CAPTURE, like META_SCS); and similarly for META_COND_RNAME/NUMBER, which duplicate the handling for the "R".

src/pcre2_compile.c
zherczeg: The old code used `(ptr - cb->start_pattern - 2)` which was copied from conditional blocks and does not make much sense. 

```
  re> /(?(1))/
Failed: error 115 at offset 2: reference to non-existent subpattern
  re> /(?(<A>))/
Failed: error 115 at offset 4: reference to non-existent subpattern
```

I though I just follow the second variant all the time. Anyway, this could be made consistent in another patch. So far nobody cared too much about error offsets, but this is probably important for you.

src/pcre2_compile.c
zherczeg: There are goto-s in the code, and this is something which is easy to forget adding them before every goto (also increase code size). I would not change it, even if it is a dead code in the scs case.

src/pcre2_compile.c
zherczeg: I have reworked the code.

src/pcre2_compile.c
NWilson: To be honest, it's not _important_ to me, I'm just being fussy. If I want to change the error offsets to match Perl, I'll do it another time. Don't worry about it now!

HACKING
```diff
@@ -766,9 +766,16 @@ In ASCII or UTF-32 mode, the character counts in OP_REVERSE and OP_VREVERSE are
 also the number of code units, but in UTF-8/16 mode each character may occupy
 more than one code unit.
 
-The "scan substring" assertion compiles as OP_ASSERT_SCS. What follows takes
-the same form as a conditional subpattern with a back reference condition (see
-next section).
+The "scan substring" assertion compiles as OP_ASSERT_SCS. This opcode is
+followed by a list of arguments (see next section). The first argument
+is always preceded by a META_OFFSET opcode, which sets the starting
```
NWilson, line 771: Can these two paragraphs possibly be right? You are saying that OP_ASSERT_SCS is followed by META_OFFSET?

META_SCS is followed by META_OFFSET; and OP_ASSERT_SCS is followed by OP_(DN)CREF, surely?

src/pcre2_compile.c
```diff
@@ -4322,14 +4315,24 @@ while (ptr < ptrend)
...
+            PCRE2_ASSERT(next_offset > 0);
+            if (offset == 0 || (next_offset - offset) >= 0x10000)
+              {
+              *parsed_pattern++ = META_OFFSET;
+              PUTOFFSET(next_offset, parsed_pattern);
+              offset = next_offset;
               }
 
+            *parsed_pattern++ = meta | (uint32_t)(next_offset - offset);
+            *parsed_pattern++ = namelen;
```
NWilson, line 4333: It might help a future maintainer to just a tiny comment here:

"We need to use a relative offset which fits in the bottom 16 bits (meta_arg), since we cannot output more that 2x uint32_t values; a numbered group reference could be as short as two characters ("1," including the comma or closing parenthesis), and the META codes must be no longer than the input pattern."

NWilson (approved):
Fantastic, the code's correct & clean. Feel free to update HACKING and merge whenever you're happy.

src/pcre2_compile.c
NWilson: Even better now, thank you!

zherczeg:
I have updated the patch. If it is ok, I will land it.

NWilson:
Yes! Go for it.

Issue #561, opened 2024-11-13
Several heap-BoF crashes from pcre2_compile related to parsed_pattern


stevenagy:
We found several crashing files (attached) through fuzzing the following program:

```
#include <stdio.h>
#define PCRE2_CODE_UNIT_WIDTH 8
#include <pcre2.h>

int main(int argc, char *argv[])
{
    FILE *f = fopen(argv[1], "rb");
    fseek(f, 0, SEEK_END);
    long size = ftell(f);
    rewind(f);
    char *fuzzData = (char*)malloc((size_t)size+1);
    fread(fuzzData, (size_t)size, 1, f);
    fuzzData[size] = '\0';

	int v0 = 1;
	size_t v1 = 1;
	pcre2_code_8* pcre2_compile_8val1 = pcre2_compile_8((PCRE2_SPTR8)fuzzData, size, 0, &v0, &v1, NULL);
   	return 0;
}
```

All crashes are heap-buffer-overflows stemming from the following source lines:
* `pcre2_compile.c:4014`
* `pcre2_compile.c:4030`
* `pcre2_compile.c:4304`
* `pcre2_compile.c:4306`
* `pcre2_compile.c:4307`

These all appear to involve the `parsed_pattern` object. 

We were able to reproduce the crashes using the PCRE2 OSS-Fuzz harness here after modifying it to read data from file: https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_fuzzsupport.c. 

Any feedback on these crashes would be greatly appreciated. :)
[crashes.zip](https://github.com/user-attachments/files/17724183/crashes.zip)

carenas:
> Any feedback on these crashes would be greatly appreciated.

just to be clear (and without looking).

are all those crashes and their line numbers associated with a specific revision of the code?, how about the latest released version?

stevenagy:
We originally found them on version  `a67878318cd08eb92e7f3afa2a15b55d46d285e2`.

stevenagy:
(sorry, accidentally hit close)

I just tested on the latest version and they appear there too. Happy to provide any other info on request.

carenas:
crashes on unreleased code are to be expected (indeed I know at least of another one I had yet to fix), it would be very important to know though if you can crash a released version of the code.

NWilson:
I have looked into this, because the code mentioned by @stevenagy is in the area I've been working on recently.

I reproduce with:
[...]

PR #568, opened 2024-11-18
Add debug assertions to ensure we never write too much to parsed_pattern

NWilson:
A follow-up to #561.

The bug was found by fuzzing - but it would have been a serious problem if we'd made a release with this bug in.

I have added deterministic assertions, so that we don't rely on fuzzing, in case we make the same (easy) mistake again in future.

It's a little bit cumbersome, since there are two cases when we _do_ write out more items to `parsed_pattern` than we consume from the input: ")" after `(*atomic_script_run:...` will write out two META_KETs when it consumes the ")", and secondly a quantifier after `(*ACCEPT)` can write out a META_NOCAPTURE/META_KET pair.

The code here is rather inelegant, but it's the best I could manage.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -563,8 +563,8 @@ static const alasitem alasmeta[] = {
   { 30, META_LOOKBEHIND_NA     },
   { 18, META_LOOKAHEADNOT      },
   { 19, META_LOOKBEHINDNOT     },
-  {  3, META_SCS_NUMBER        }, /* placeholder, updated later */
-  { 14, META_SCS_NUMBER        }, /* placeholder, updated later */
+  {  3, META_SCS               },
+  { 14, META_SCS               },
   {  6, META_ATOMIC            },
   {  2, META_SCRIPT_RUN        }, /* sr = script run */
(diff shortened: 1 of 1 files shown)


src/pcre2_compile.c
```diff
@@ -2766,6 +2766,60 @@ return parsed_pattern;
...
+
+Arguments:
+  ptr             points to the start of the pattern
+  ptrend          points to the end of the pattern
+  utf             TRUE in UTF mode
+  options         the options bits
+
+Returns:          the number of uint32_t units for parsed_pattern
+*/
+static ptrdiff_t
```
zherczeg, line 2785: `size_t` ? Or PCRE2_SIZE?

zherczeg (approved):
LGTM

src/pcre2_compile.c
NWilson: This was intentional. There is some arithmetic in the PCRE2_ASSERT lines, and some of the expressions are negative, so it made sense to make all the expressions signed rather than unsigned. Otherwise there would be ugly casts.

PR #569, opened 2024-11-20
Move character matching code into pcre2_jit_char_inc.h

zherczeg:
Large code movement patch. No functionality change. Based on #567


carenas:
I wonder if we shoukd start doing more feature branches to make reviewing this much easier.

IMHO we coud still keep history flat (like it was with subversion) by forcing squash merging at the end, and maybe even have better merge commits with some smallwork in tooling and some standardization.

PR #571, opened 2024-11-22
Add coverage job

### LINKED ISSUE #555: Add lcov coverage report to CI jobs
There are various error cases and stray blocks of code that appear not to be covered by the unit tests.

We should do a little bit more monitoring of the `lcov` report, and close any loopholes.

Noticed while checking for missed lines or branches in the new extended classes code.

NWilson:
Fixes #555

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,3 +129,61 @@ jobs:
       - name: Debug
         if: ${{ failure() && steps.test.outcome == 'failure' }}
         run: type build/Testing/Temporary/LastTest.log
+
+  coverage:
+    name: Code coverage
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - name: Configure
+        run: CC="clang -fprofile-instr-generate -fcoverage-mapping" cmake -DCMAKE_BUILD_TYPE=Debug -DPCRE2_SUPPORT_JIT=ON -DPCRE2_BUILD_PCRE2_16=ON -DPCRE2_BUILD_PCRE2_32=ON -B build
+
+      - name: Build
+        run: cmake --build build
+
+      - name: Test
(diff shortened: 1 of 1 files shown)


zherczeg:
This looks useful.

`pcre2_jit_compile.c.html#L102: 58.9k #define MACHINE_STACK_SIZE 32768`

Interesting that a const has coverage. Some macros has coverage, others don't, so probably not all code is fully "covered" by the tool :)

NWilson:
I'd guess that constants which are built into the binary will be counted as "code" and generate coverage.

Macros which are only used by the preprocessor won't generate coverage. They have to actually write out some text which is seen as C code by the compiler.

I'd be interested in trying out the "Coveralls" or "CodeCov" services, which look like good ways to track coverage over time, coverage diff in a PR, and so on.

.github/workflows/build.yml
```diff
@@ -129,3 +129,61 @@ jobs:
...
+            ../src/ ./ \
+            >> "$GITHUB_STEP_SUMMARY"
+          echo '```' >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Upload report
+        uses: actions/upload-artifact@v4
+        with:
+          name: "Coverage report"
+          path: './build/coverage-report'
+          if-no-files-found: error
```
carenas, line 189: missing "\n" at EOF

carenas (changes requested):

.github/workflows/build.yml
NWilson: Thanks! Fixed

carenas (approved):
it might be better to move this job into their own "tools.yml" in te long run, though

NWilson:
> it might be better to move this job into their own "tools.yml" in te long run, though

I feel like code-scanning tools (like CodeQL, or Clang Static Analyzer) should be in their own jobs.

However, code coverage should be part of the Release "Build" pipeline (not the "Dev" pipeline) because it's actually a fixed, reportable artifact of the release process: "we built the code; we ran the tests; the tests covered 85% of the code; now we release the binaries".

PR #575, opened 2024-11-22
Configure Clang static analyzer CI

NWilson:
In this PR, I'm only fixing trivial/cosmetic issues.

However, Clang's analyzer has found two genuine issues (I believe), which I'll address separately.


src/pcre2_printint.c
```diff
@@ -436,8 +434,6 @@ while (type > 0)
   else if (list_ind == 2) char_list_add = XCL_CHAR_LIST_LOW_32_ADD;
   else char_list_add = XCL_CHAR_LIST_HIGH_32_ADD;
   }
-
-return code + LINK_SIZE;
```
zherczeg, line 440: How is this working without adding the link size?

.github/workflows/scorecards.yml
```diff
@@ -51,6 +51,7 @@ jobs:
 
       # Upload the results to GitHub's code scanning dashboard.
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@5f532563584d71fdef14ee64d17bafb34f751ce5 # v1.0.26
+        uses: github/codeql-action/upload-sarif@v3
         with:
           sarif_file: results.sarif
+          category: ossf-scorecard
```
zherczeg, line 57: Is there a reason to not add newlines? I saw this in another patch.

NWilson: Unintentional

src/pcre2_printint.c
zherczeg: It does not seem to be wrong. But it can be unused, if the xclass global siye used instead.

NWilson: In that case, I will investigate further why the old code seemed not to be working.

zherczeg: Don't spend too much effort on it. An assert could be good, but if it makes the code too complex, it is not worth it. Perhaps the assert could be in the compiler to check that the XCLASS length is filled correctly (no dummy bytes at the end) to reduce possible mistakes in future code. But it can be done later.

NWilson: I've worked it out: your code to advance to the end of the XCL_LIST is correct (my apologies).

The issue was that I didn't realise there is no XCL_END after an XCL_LIST.

In fact - we should just get rid of XCL_END completely. We can do:

```
code_end = code + GET(code, 1);
while (code < code_end) { ... }

// instead of:

while (*code != XCL_END) { ... }
```

It's a waste of a byte to use an XCL_END terminator, when we've already encoded the length of the XCLASS data.

For now, all I have done is add an assertion that the XCL_LIST is the final item in the XCLASS data.

src/pcre2grep.c
```diff
@@ -1539,12 +1539,13 @@ switch(endlinetype)
   for (;;)
     {
     while (p < endptr && *p != '\r') p++;
-    if (++p >= endptr)
+    if (p == endptr)
```
NWilson, line 1542: The code change here should absolutely no effect.

However, it silences the Clang warning. The old code would do `++p` after p had already reached endptr, which is a two-past-the-end pointer.

Clang notices that this is Undefined Behaviour (you can legally construct a one-past-the-end pointer, but no further).

My updated code here simply ensures that `p < endptr` before we increment it.

src/pcre2_match.c
```diff
@@ -5792,6 +5792,8 @@ fprintf(stderr, "++ %2ld op=%3d %s\n", Fecode - mb->start_code, *Fecode,
 
       /* Disable compiler warning. */
       offset = 0;
```
NWilson, line 5794: "Disable compiler warning" is a bit ironic :)

Clang now complains about the fact that we're writing a dead value to a variable, when there are no further reads from it!

src/pcre2_printint.c
zherczeg: This is true. I suspect it was done this way because of the interpreter. If code processing does not become too complex, I am ok to remove the XCL_END.

src/pcre2_match.c
zherczeg: Since multiple of such cases are updated: is there a better way to do this? Separating smart and stupid compilers in some way?

NWilson: Doing `(void)variable` in 5 or ten places is the just the tax you pay to support multiple compilers, and avoid generating warnings on all of them. Ugly, but harmless.

For what it's worth: Clang is being clever, not stupid. Assigning to a variable, but never reading the value back, is a very helpful warning that can catch typos when you have variables with similar names.

src/pcre2_printint.c
NWilson: This PR is complete - no changes to XCL_END right now.

zherczeg (approved):
LGTM

PR #580, opened 2024-11-25
Use [^...] rather than '[...] (neg)' to print NCLASS

NWilson:
This is a cosmetic change in the tests. It looks much better.

I've also fixed a highly confusing comment in printint.c which suggests that the character map for XCLASS behaves differently to NCLASS (it doesn't now - perhaps it did previously).


zherczeg:
It looks better, but we loose the "nclass" information from the output. Could we keep it some way? (Note: classes can contain `^` as a letter).

zherczeg:
```
  re> /[\^a]/B
        Bra
        [^a]
        Ket
        End
```

NWilson:
Well-spotted. This is an existing bug for XCLASS printing - and worse, even backslash isn't escaped currently!

I've fixed those cases too (and added tests).

NWilson:
You're saying that, because we use "[^a]" to represent OP_NOT, you want the output for an OP_NCLASS to be distinguishable. That makes sense. And we're using the "!" notation in eclasses, so that's a consistent notation to use for OP_NCLASS.

I am half-way through the optimisation work. You might be happy to know that the OP_ECLASS_NOT code will disappear when I'm finished (it will always be constant-folded in to atomic classes within the eclass).

zherczeg:
Yes. The purpose of /B is debugging, and less constructs have the same text representation is the better.

Removing OP_ECLASS_NOT is good news indeed, nice work! How do you encode `[^[\pC]&&[\pL]]` without it?

NWilson:
> How do you encode [^[\pC]&&[\pL]] without it?

We know that `!(A && B) = !A || !B`. So we can distribute the NOT over the AND and OR (and SUB and XOR), and flip the operands and flip the operators.

Issue #577, opened 2024-11-24
pcre2grep(1): Bug when context lines are consecutive


alejandro-colomar:
```sh
alx@devuan:~/tmp/grep$ cat foo 
zxc
cvb
bnm
asd
qwe
ert
tyu
uio
ggg
asd
dfg
ghj
jkl
123
345
567
789
alx@devuan:~/tmp/grep$ pcre2grep -nC3 '(ert|jkl)' foo
3-bnm
4-asd
5-qwe
6:ert
7-tyu
8-uio
9-ggg
--
10-asd
11-dfg
12-ghj
13:jkl
14-123
15-345
16-567
alx@devuan:~/tmp/grep$ pcre2grep -nC3 '(ert|ghj)' foo
3-bnm
4-asd
5-qwe
6:ert
7-tyu
8-uio
9-ggg
--
10-asd
11-dfg
12:ghj
13-jkl
14-123
15-345
alx@devuan:~/tmp/grep$ pcre2grep -nC3 '(ert|dfg)' foo
3-bnm
4-asd
5-qwe
6:ert
7-tyu
8-uio
9-ggg
--
10-asd
11:dfg
12-ghj
13-jkl
14-123
alx@devuan:~/tmp/grep$ pcre2grep -nC3 '(ert|asd)' foo
1-zxc
2-cvb
3-bnm
4:asd
5-qwe
6:ert
7-tyu
8-uio
9-ggg
10:asd
11-dfg
12-ghj
13-jkl
```

These lines should be printed without a separator, since they are consecutive.

Here's some reproducers with different -A and -B:

```sh
alx@devuan:~/tmp/grep$ pcre2grep -n -B3 -A3 '(ert|dfg)' foo
3-bnm
4-asd
5-qwe
6:ert
7-tyu
8-uio
9-ggg
--
10-asd
11:dfg
12-ghj
13-jkl
14-123
alx@devuan:~/tmp/grep$ pcre2grep -n -B2 -A4 '(ert|dfg)' foo
4-asd
5-qwe
6:ert
7-tyu
8-uio
9-ggg
10-asd
11:dfg
12-ghj
13-jkl
14-123
15-345
alx@devuan:~/tmp/grep$ pcre2grep -n -B4 -A2 '(ert|dfg)' foo
2-cvb
3-bnm
4-asd
5-qwe
6:ert
7-tyu
8-uio
--
9-ggg
10-asd
11:dfg
12-ghj
13-jkl
```

[...]

PR #581, opened 2024-11-26
Add RunGrepTest.bat to CMake

### LINKED ISSUE #551: RunGrepTest.bat very out-of-date
The Windows builds aren't running & testing RunGrepTest. As a result, it hasn't been kept in sync with changes to the Unix version, and it no longer passes.

* Update the CMake config to actually run it on Windows
* Fix it so that it passes

NWilson:
The Windows builds are using CMake, which wasn't running the grep tests on Windows.

This caused the tests to become very out of date (because no-one was running them).

I have re-synced all the tests from `RunGrepTest` to `RunGrepTest.bat` by keeping the existing batch script boilerplate, but splatting the contents of the Bash script into it, then fixing it up into Batch syntax (with regexes in my editor, of course).

A few newly-added tests in `RunGrepTest` had to be tweaked a bit, since they do things that are hard to do on Windows. For example: constructing a file using `printf(1)` that has some invalid UTF-8; I just checked these inputs in to git instead of trying to work out a Batch spell to build the same file.

Fixes #551


.gitattributes
```diff
@@ -0,0 +1 @@
+testdata/*    -text
```
NWilson, line 1: We really want to disable all the CRLF/LF line conversions that git likes to do. This was responsible for much of the pain in the CI test jobs. We want to treat everything inside `testdata/` as binary files (as far as git is concerned) - please don't mess with anything! Check them out so that they are the exact same bytes on disk on Windows & Unix.

src/pcre2grep.c
```diff
@@ -2528,6 +2528,7 @@ while (length > 0)
 necessary, otherwise assume fork(). */
 
 #ifdef WIN32
+(void)fflush(stdout);
```
NWilson, line 2531: Running the Grep tests on Windows showed that they need the same fflush() fix as Unix.

I believe the Unix code is still wrong however: you're doing `fflush()` after the call to fork()! We need to flush first, before the child process starts writing anything. I'm a bit surprised that you got away with that (probably because most of the printing that's done by pcre2grep is newline-terminated, and stdout is flushed on newline usually).

You didn't need two calls to fflush either on Unix (one before and one after waitpid).

testdata/grepinputUN
```diff
@@ -0,0 +1,2 @@
+abcሴdef
+xyz
```
NWilson, line 2: Deliberate lack of trailing newline. I just checked in the file, as it was being created in the tests.

PhilipHazel:
The fflush()s were probably me doing belt-and-braces stumbling around. That is in any case probably a rarely used bit of pcre2grep. I have prepared a fix for #577 which of course modifies pcre2grep.c. I will hold it off until this PR is merged. I also won't add new tests till then.

.gitattributes
PhilipHazel: Where is this .gitattributes file, please? I doesn't seem to be in the GitHub repo.

PhilipHazel: Oh hang on, I'm being silly. You are proposing to ADD this file. Doh!

PhilipHazel:
Is this PR ready for merging? I am not qualified to judge...

carenas (approved):

zherczeg (approved):
LGTM

PhilipHazel:
Following this, I merged my fix for #577, but because I added to the pcre2grep tests, I seem to have broken the Windows CI. Please can somebody fix? I'm looking at why the Alpine CI failed.

PhilipHazel:
Fixed Alpine. I think I can fix Windows, so don't spend time (yet).

PR #582, opened 2024-11-26
Install zlib in CI builds

NWilson:


.github/workflows/build.yml
```diff
@@ -16,7 +21,7 @@ jobs:
         run: ./autogen.sh
 
       - name: Configure
-        run: ./configure --enable-jit --enable-pcre2-16 --enable-pcre2-32
+        run: ./configure --enable-jit --enable-pcre2-16 --enable-pcre2-32 --enable-pcre2grep-libz --enable-pcre2grep-libbz2 --enable-pcre2test-libreadline
```
carenas, line 24: we also have `libeditline` support (which is preferred due to license in the BSD)

NWilson: OK, I can use that.

I really wanted the CI to be testing the zlib/bzip2 support; and when I added that, I thought it would be reasonable to throw in a test for readline too (although we're only testing it builds & links, and not really exercising the readline support).

The readline support is not really important, since it's just test-only. It would be OK if it broke. But the zlib support can't break.

NWilson: It would be nicer if other people could maintain the frontend (like even GNU grep has `-P` now for PCRE!)

But I guess if people are using pcre2grep, then it needs to be tested.

I'm happy with this PR. We only need a minimal effort (smoke test) to exercise these external library integrations.

carenas (approved):

zherczeg (approved):
LGTM

PR #583, opened 2024-11-26
JIT compiler update

zherczeg:
Testing on some non-x86 cpus could be added to CI at some point (ARM, x390x)

### DIFF
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -13223,7 +13223,7 @@ common->compiler = compiler;
 
 /* Main pcre2_jit_exec entry. */
 SLJIT_ASSERT((private_data_size & (sizeof(sljit_sw) - 1)) == 0);
-sljit_emit_enter(compiler, 0, SLJIT_ARGS1(W, W), 5 | SLJIT_ENTER_FLOAT(SLJIT_NUMBER_OF_SCRATCH_FLOAT_REGISTERS), 5, private_data_size);
+sljit_emit_enter(compiler, 0, SLJIT_ARGS1(W, W), 5 | SLJIT_ENTER_VECTOR(SLJIT_NUMBER_OF_SCRATCH_VECTOR_REGISTERS), 5, private_data_size);
 
 /* Register init. */
 reset_ovector(common, (re->top_bracket + 1) * 2);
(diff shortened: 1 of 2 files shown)


carenas:
funny enough we are already testing in 64bit ARM, which is what the macOS runner has.

for the rest we are still relying in the "mostly emulated" SLJIT "CI", but agree setting up our own workers for the rest would be useful.

src/pcre2_jit_compile.c
```diff
@@ -13223,7 +13223,7 @@ common->compiler = compiler;
 
 /* Main pcre2_jit_exec entry. */
 SLJIT_ASSERT((private_data_size & (sizeof(sljit_sw) - 1)) == 0);
-sljit_emit_enter(compiler, 0, SLJIT_ARGS1(W, W), 5 | SLJIT_ENTER_FLOAT(SLJIT_NUMBER_OF_SCRATCH_FLOAT_REGISTERS), 5, private_data_size);
+sljit_emit_enter(compiler, 0, SLJIT_ARGS1(W, W), 5 | SLJIT_ENTER_VECTOR(SLJIT_NUMBER_OF_SCRATCH_VECTOR_REGISTERS), 5, private_data_size);
```
carenas, line 13226: is SLJIT SIMD really being used now, or this is an abstraction that leaked?

carenas (commented):
how is this going to affect RISCV, knowing we have a problem there with SIMD that could result in crashes?

src/pcre2_jit_compile.c
zherczeg: The x86 uses the simd interface. Now the simd interface has correct vector register handling.

I did not have time to merge the others and simplify those platforms.

carenas (approved):

PhilipHazel:
I seem to have a problem here. When I updated with this PR, I got this error:

```
src/pcre2_jit_compile.c: In function 'jit_compile':
src/pcre2_jit_compile.c:13226:54: error: implicit declaration of function 'SLJIT_ENTER_VECTOR'; did you mean 'SLJIT_ENTER_KEEP'? [-Wimplicit-function-declaration]
13226 | sljit_emit_enter(compiler, 0, SLJIT_ARGS1(W, W), 5 | SLJIT_ENTER_VECTOR(SLJIT_NUMBER_OF_SCRATCH_VECTOR_REGISTERS), 5, private_data_size);
      |                                                      ^~~~~~~~~~~~~~~~~~
      |                                                      SLJIT_ENTER_KEEP
src/pcre2_jit_compile.c:13226:54: warning: nested extern declaration of 'SLJIT_ENTER_VECTOR' [-Wnested-externs]
```
I can't find a definition for SLJIT_ENTER_VECTOR anywhere in the source. What am I doing wrong?

zherczeg:
what `git status` says? Is it possible that you did not fetch the `deps/sljit` and moved to the expected hash?

PhilipHazel:
Thank you! I keep forgetting that I have to fetch deps/sljit separately.

PR #586, opened 2024-11-27
Add folding and simplication for OP_ECLASS

### LINKED ISSUE #537: Extended character classes: optimisation
As a follow-up to my PR, I will refactor the parser which consumes META codes.

It should do constant-folding on simple cases, such as `[a-z -- m]` or `[^[^a-z]]`. We do still need the OP_ECLASS code, for tricky cases that can't be folded (complex logical ops over XCLASS), but doing AND/OR operations on an OP_CLASS can be folded at compile-time.

NWilson:
Fixes #537

Whew! This was a bit of a brain-bending PR, for me at least.

OP_ECLASS now _only_ contains ECS_XCLASS, and ECL_AND...ECL_XOR codes. No more CLASS, NCLASS nested inside; we fully-fold those, so anything like `(?[  [a-z] - [aeiou] ])` will no longer generate an OP_ECLASS. You have to actually put two XCLASSes inside the extended character class, in order to something that we can't fully fold.

In addition, for the cases that do result in an OP_ECLASS (mainly, logical operations on Unicode classes), we hoist up the bitmap so that ASCII characters can be matched without needing to consult any of the unicode properties.


HACKING
```diff
@@ -633,6 +633,24 @@ When XCL_NOT is set, the bit map, if present, contains bits for characters that
...
+containing the total length of the opcode and its data. That is followed by a
+code containing flags: currently just ECL_MAP indicating that a bit map is
+present. There follows the bit map, if ECL_MAP is set. Finally a sequence of
+items that are either an operand or operator. Each item starts with a single
+code unit containing its type:
+
+  ECL_AND       AND; no additional data
+  ECL_OR        OR; no additional data
+  ECL_XOR       XOR; no additional data
+  ECL_NOT       NOT; no additional data
```
zherczeg, line 649: I think you mentioned that not is removed. Is there a special type (sequence) for all/nothing?

src/pcre2_compile.c
```diff
@@ -782,12 +782,16 @@ static const uint8_t opcode_possessify[] = {
...
-  0, 0, 0,                 /* CLASS, NCLASS, XCLASS */
+  0, 0, 0, 0,              /* CLASS, NCLASS, XCLASS, ECLASS */
   0, 0,                    /* REF, REFI */
   0, 0,                    /* DNREF, DNREFI */
-  0, 0                     /* RECURSE, CALLOUT */
+  0, 0,                    /* RECURSE, CALLOUT */
 };
 
+/* Compile-time check that the table has the correct size. */
+typedef int check_opcode_possessify[
```
zherczeg, line 792: Maybe we should introduce static asserts.

src/pcre2_compile.c
```diff
@@ -6169,21 +6176,163 @@ for (;; pptr++)
...
+        code = previous;
+        /* (*lengthptr - previous_length) now holds the amount of buffer that
+        we require to make the call to compile_class_nested() with
+        lengthptr = NULL, and including the (1+LINK_SIZE+1) that we write out
+        before that call. */
+        }
 
-      break;   /* We are finished with this class */
+      /* Do some useful counting of what's in the bitmap. */
+      while (classones < 32 && op_info.bits[classones] == 0xff) classones++;
```
zherczeg, line 6203: If you use uint32_t type (can be a union), processing these is much faster.

src/pcre2_compile.c
```diff
@@ -6169,21 +6176,163 @@ for (;; pptr++)
...
+      no-Unicode mode the matcher is compiled without support for OP_ECLASS. */
+
+#ifndef SUPPORT_WIDE_CHARS
+      PCRE2_ASSERT(op_info.op_single_type != 0);
+#else
+      if (op_info.op_single_type != 0)
+#endif
+        {
+        /* Rewind back over the OP_ECLASS. */
+        code = previous;
```
zherczeg, line 6218: Can we compute this at length computation time, store the bit in the META code, and use this knowledge when the byte code is generated. We might do a buffer overrun otherwise.

src/pcre2_compile.c
```diff
@@ -6169,21 +6176,163 @@ for (;; pptr++)
...
+
+        /* If the bits are all ones, and the "high characters" are all matched
+        too, we use a special-cased encoding of OP_ALLANY. */
+
+        if (op_info.op_single_type == ECL_ANY && classones == 32)
+          {
```
zherczeg, line 6224: No need braces for this

src/pcre2_compile.h
```diff
@@ -218,11 +218,26 @@ therefore no need for it to have a length entry, so use a high value. */
...
+typedef struct {
+  /* The position of the operand - or NULL if (lengthptr != NULL). */
+  PCRE2_UCHAR *code_start;
+  PCRE2_SIZE length;
+  /* The operand's type if it is a single code (ECL_XCLASS, ECL_ANY, ECL_NONE);
+  otherwise zero if the operand is not atomic. */
+  uint8_t op_single_type;
+  /* Regardless of whether it's a single code or not, we fully constant-fold
+  the bitmap for code points < 256. */
+  uint8_t bits[32];
```
zherczeg, line 234: I mentioned this could be an uint32_t array as well using an union, and merging / counting 32 bit values are 4 times faster than bytes.

src/pcre2_compile_class.c
```diff
@@ -1961,38 +2360,83 @@ return TRUE;
...
+  eclass_op_info rhs_op_info;
+  
+  if (negated)
+    {
+    /* The whole expression is being negated; we respond by unconditionally
+    negating the LHS A, before seeing what follows. And hooray! We can recover,
+    no matter what follows. */
+    /* !(A || B)   ->  !A && !B                     */
+    /* !(A -- B)   ->  !(A && !B)    ->  !A || B    */
+    /* !(A XOR B)  ->  !(!A XOR !B)  ->  !A XNOR !B */
```
zherczeg, line 2409: XNOR?

zherczeg (commented):
Nice progress! I will check the rest later.

src/pcre2_compile_class.c
NWilson: Yes, I think XNOR exists.

`A XOR B ⇔ (A ∧ ¬B) ∨ (¬A ∧ B)`
`A XNOR B ⇔ (A ∧ B) ∨ (¬A ∧ ¬B)`

HACKING
NWilson: I hadn't mentioned them in the HACKING document, because they are not part of a fully-formed OP_ECLASS, but yes, there are two values ECL_ANY and ECL_NONE.

When one of these appears as the LHS/RHS of a binary operator, it can be constant-folded away. Therefore, in the final return value from `compile_class_nested()`, the result either has no instances of ECL_ANY or ECL_NONE, or the result is a single ECL_ANY or ECL_NONE item. In this case, we don't need to wrap it in an OP_ECLASS, so at match-time, an OP_ECLASS will only ever contain ECL_XCLASS items plus operators.

src/pcre2_compile.c
NWilson: This _is_ a static assert. I guess you mean, you'd prefer to wrap it in a macro? We could do that.

src/pcre2_compile.c
NWilson: Hmm. There are existing uses of `uint8_t[8]`, such as `pcre2_real_code.start_bitmap` (which has a few 32-count loops).

But that makes sense, I'll replace with a `class_bits_storage`.

src/pcre2_compile.c
NWilson: There shouldn't be a buffer overrun?

This is the rule I'm following: "don't do any conditional logic / branches inside "`if (lengthptr == NULL)`" blocks". It's totally OK to check whether lengthptr == NULL, but you need to absolutely sure when you do that, that the code will behave the same with and without lengthptr.

What would be terrible, would be to do "better" optimisation when `lengthptr == NULL`, or to do anything different at all (apart from actually writing the bytecode itself). We have to make exactly the same optimisation decisions in each case. The `if (lengthptr == NULL)` blocks need to extremely short, and careful, so that they don't affect control flow.

That way, we can be sure we'll have enough buffer space - because we followed all the same code paths each time.

NWilson:
I know the tests are failing. I'll sort that out!

I'm also going to do additional testing.

I plan to write a grammar-driven fuzzing script, to generate millions of random extended classes which have valid syntax, and deep, complex expressions. Then I'll run pcre2test using the OLD code, to generate a reference answer for each.

If pcre2test produces the same output with the NEW code, then we can be almost certain that the optimisations don't have any effect on behaviour.

NWilson:
Actually, _don't_ start reviewing this quite yet.

A big chunk of the core logic is all present & correct, but there are some bits of work-in-progress. The code is close, but there are some very fiddly details around XCLASS handling I'm still wrestling with.

zherczeg:
Just a question: if multiple XCLASS-es uses property accesses, they each read the ucd data structure separately. On the long run, it would be good to do it only once, and share the ucd data pointer. A flag bit could be useful in ECLASS, which tells, that at east one XCLASS needs ucd data. Then ECLASS reads the data, and shares with XCLASS-es. Adding such flag is easy or difficult in your current implementation?

NWilson:
Great! There are some minor bugs still to fix (CI tells me that the no-Unicode build has some failing tests).

However, is working nearly perfectly, and I don't expect any substantial changes from here on.

The code which builds XCLASSes (using char lists etc) is incredibly complex, and it would be rather nice to try hard to simplify it at some point. There are two very separate codepaths:
- The case for 16-bit/32-bit, or 8-bit with UTF set. These go via cranges, but also share code with the pure-8-bit-chars codepath, in a bit of a messy arrangement
- The pure-8-bit codepath, with no cranges, for 8-bit library without UTF set. This basically has its own special-cased code, but it's mingled in with other cases.

Fortunately... this PR really doesn't go into any of that stuff, even though it builds on top of it, so I've simply left it alone for now. However, I am very tempted to do a follow-up refactor of `PRIV(compile_class_not_nested)()`.

zherczeg:
Let us know when this patch is ready for review.

NWilson:
I have some additional testing to do (eg add some tests where we don't have 100% coverage of all the optimisation paths). So it's not ready to _merge_.

But the code is stable and ready for review.

src/pcre2_compile.c
```diff
@@ -6139,13 +6142,16 @@ for (;; pptr++)
...
+      memset(code, 0, 32);
       code += 32 / sizeof(PCRE2_UCHAR);
       }
 
     if (firstcuflags == REQ_UNSET) firstcuflags = REQ_NONE;
     zerofirstcu = firstcu;
     zerofirstcuflags = firstcuflags;
+    // XXX revisit why these flags here are different to when we process a
+    // non-empty class - is this a problem? You can have '[^\d\D]' is basically
+    // the same as '[]' so they should surely set the same flags.
```
NWilson, line 6154: This is a question for reviewers.

The handling for META_CLASS_EMPTY_NOT emits OP_ALLANY, and sets the flags:

>    if (firstcuflags == REQ_UNSET) firstcuflags = REQ_NONE;
>    zerofirstcu = firstcu;
>    zerofirstcuflags = firstcuflags; 

And, the handling for `[\d\D]` also emits OP_ALLANY, but sets the different flags:

>    if (firstcuflags == REQ_UNSET) firstcuflags = REQ_NONE;
>    zerofirstcu = firstcu;
>    zerofirstcuflags = firstcuflags;
>    zeroreqcu = reqcu;
>    zeroreqcuflags = reqcuflags;

Is this a problem? Is there any impact?

HACKING
```diff
@@ -633,6 +633,41 @@ When XCL_NOT is set, the bit map, if present, contains bits for characters that
 are allowed (exactly as for OP_NCLASS), but the list of items that follow it
 specifies characters and properties that are not allowed.
 
+The meaning of the bitmap indicated by XCL_MAP is that, if one is present, then
+it fully describes which code points < 256 match the class (without needing to
+invert the check according to XCL_NOT); the other items in the OP_XCLASS need
+not be consulted. However, if a bitmap is not present, then code points < 256
+may still match, so the other items in the OP_XCLASS must be consulted.
```
NWilson, line 640: Would someone mind checking I've understood correctly the behaviour of the bitmap for OP_XCLASS?

I made a bug in an earlier version of the PR, where I had assumed that, if an XCLASS had no bitmap, then it wouldn't match any ASCII characters. I now realise that is wrong.

If we're willing to spend 32 bytes though, we could speed up matching of things like `\pL` if we _did_ include the bitmap for all cases of XCLASS which match in the ASCII range.

NWilson:
I have added additional tests, until the code-coverage report reached 100% branch coverage for all the (reachable) branches in the ECLASS code.

I will do further testing by fuzz-testing the new (optimised) ECLASS behaviour against the old reference implementation.

HACKING
zherczeg: Bitmap is generated, if cranges has a range, which start is less than 256. The \p or \P properties might match to characters < 256 even if there is no bitmap.

NWilson: Great! That matches my understanding now.

NWilson:
I have completed my testing.

I wrote a script, which generates random extended character classes, UTS#18 and Perl-style, in order to fuzz-test the optimisations in this PR. I generated millions of test cases (GB of testinput), and ran the test cases using pcre2test, comparing master to this branch. The output files were identical.

Then I ran my fuzzed inputs using the Clang's coverage check... and found I'd forgotten to generate cases for one of the operators! Oops. So I re-ran... all good. The coverage verifies that the randomised test cases are covering all the branches of the optimiser, and probably covering far more paths through the code than the manually-written test cases.

src/pcre2_compile.c
```diff
@@ -6170,21 +6176,186 @@ for (;; pptr++)
...
+          *code++ = (op_info.op_single_type == ECL_ANY)? OP_NCLASS : OP_CLASS;
+          memcpy(code, op_info.bits.classbits, 32);
+          code += 32 / sizeof(PCRE2_UCHAR);
+          }
+
+        /* Otherwise, we have an ECL_XCLASS, so we have the OP_XCLASS data
+        there, but, we pulled out its bitmap into op_info, so now we have to
+        put that back into the OP_XCLASS. */
+
+#ifndef SUPPORT_WIDE_CHARS
```
zherczeg, line 6258: You can move the entire `#ifndef` into the `else` below.

src/pcre2_compile_class.c
```diff
@@ -1783,85 +1787,421 @@ return pptr - 1;
...
+
+
+/* This function folds together two operands using a binary operator.
+The new, combined chunk of stack code is written out to *lhs_op_info. */
+
+static void
+fold_binary(int op, eclass_op_info *lhs_op_info, eclass_op_info *rhs_op_info,
+  PCRE2_SIZE *lengthptr)
+{
+switch (op) {
```
zherczeg, line 1852: I think this is the format:
```
switch (op)
  {
  case X:
  break;
  }
```

src/pcre2_compile_class.c
zherczeg: I mean is it the combination of ECL_XOR / ECL_NOT?

testdata/testoutput2
```diff
@@ -20047,11 +19972,7 @@ No match
 /[z||[^\dAC-E[:space:]]]/B,alt_extended_class
 ------------------------------------------------------------------
         Bra
-        eclass[
-          cls:[z]
-          cls:[^\x09-\x0d 0-9AC-E]
-          op: ||
-        ]
+        [^\x09-\x0d 0-9AC-E]
```
zherczeg, line 19975: Nice!

testdata/testoutput2
```diff
@@ -20270,13 +20138,7 @@ No match
...
 ------------------------------------------------------------------
         Bra
-        eclass[
-          cls:[ ^a]
-          cls:[ ^b]
-          op: ||
-          cls:[ ]
-          op: ||
-        ]
+        [ ^ab]
```
zherczeg, line 20141: Is there tests with /x and comments / spaces?

testdata/testoutput5
```diff
@@ -6238,9 +6238,10 @@ Failed: error 115 at offset 52: reference to non-existent subpattern
 ------------------------------------------------------------------
         Bra
         eclass[
-          cls:[\p{Lu}]
-          cls:[\p{Nd}]
-          op: ||
+          bitmap: [0-9A-Z\xc0-\xd6\xd8-\xde]
```
zherczeg, line 6241: I think this is exactly the case where no bitmap is needed.

zherczeg (changes requested):
Nice progress!

src/pcre2_compile.c
NWilson: Whenever I rewind the `code` pointer, I increment the `lengthptr` by the same amount.

* Say, we write two bytes to `code`, and request another 5 via `lengthptr`. (A total of 7 bytes requested)
* Then we rewind `code` (by 2 bytes), and increment `lengthptr` to compensate. (Still 7 bytes requested)
* Now we write more bytes to `code` - can initially decrement `lengthptr` (if writing 7 bytes or less), but then if we are requesting more bytes, we'll increase the size of the allocation

src/pcre2_compile_class.c
NWilson: Yes, I have implemented XNOR by writing an XOR followed by NOT, rather than giving it a dedicated ECL_ code.

.github/workflows/dev.yml
```diff
@@ -68,7 +68,7 @@ jobs:
 
       - name: Test (main test script)
         run: |
-          ulimit -S -s 32768 # Raise stack limit; ASAN with -O0 is very stack-hungry
+          ulimit -S -s 49152 # Raise stack limit; ASAN with -O0 is very stack-hungry
```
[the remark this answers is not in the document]
NWilson: Yes. The fuzz tests generate very deep stacks. The stack usage for ordinary builds is quite sensible (15 * 32 bytes), but the stack frames and stack allocations have absurdly-large amounts of guard space in the ASAN builds, so I had to bump this higher.

The eclass code (with a 15-deep parenthesis limit) uses similar or less stack than the main regex parser, which is also recursive-descent with a 255-deep limit.

src/pcre2_compile.c
```diff
@@ -6139,13 +6142,16 @@ for (;; pptr++)
     else
       {
       *code++ = OP_CLASS;
-      memset(code, 0, 32 * sizeof(uint8_t));
+      memset(code, 0, 32);
```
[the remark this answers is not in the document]
NWilson: Yes. I just noticed a load of `sizeof(uint8_t)` and thought it was simpler to just remove (personal style).

src/pcre2_compile_class.c
NWilson: Thanks! I'll fix these formatting issues.

testdata/testoutput5
NWilson: Yes, sadly. Because ECLASS combines bitmaps in ways that are harder to reason about, I made the ECLASS bitmap behaviour a bit different to XCLASS (see the changes in HACKING). For ECLASS, I made it so that there is a bitmap if and only if the ECLASS matches any characters < 256. So an ECLASS will have at most one bitmap, but it will be preserved for cases like `\pL & \pN`.

(This example is even worse, because additional optimisation would actually mean it could be simplified to a simple XCLASS.)

zherczeg: Ok. My idea was to detect that bitmap is needed in the `lengthptr != NULL` phase, and use this info later. I will try to do it after this patch is landed.

src/pcre2_compile.c
zherczeg: I haven't spent time to understand these constructs, so this is also magic to me.

src/pcre2_compile.c
```diff
@@ -6170,21 +6176,193 @@ for (;; pptr++)
...
+        /* (*lengthptr - previous_length) now holds the amount of buffer that
+        we require to make the call to compile_class_nested() with
+        lengthptr = NULL, and including the (1+LINK_SIZE+1) that we write out
+        before that call. */
+        }
+
+      /* Do some useful counting of what's in the bitmap. */
+      for (int i = 0; i < 8; i++)
+        if (op_info.bits.classwords[i] != 0xffffffff)
+          { allbitsone = FALSE; break; }
```
zherczeg, line 6205: The syntax for these is not changed.

zherczeg (approved):
Please squash the commits, and fix the minor things. Good progress.

zherczeg:
Is there anybody who plans to review it later? If not, I plan to land it after a few changes.

src/pcre2_compile.c
NWilson: I did change the indentation. Is it wrong to put it all on one line?

src/pcre2_compile.c
NWilson: Perhaps @PhilipHazel could comment on this one specific question.

NWilson:
Squashed. This should be good to go!

src/pcre2_compile.c
NWilson: I have split the code onto multiple lines.

src/pcre2_compile.c
PhilipHazel: Always happy to hepple. :-)

PhilipHazel:
I tried merging this ... after a while the merge button came back so (stupidly) I clicked it again. It now says merge already in progress... has it done it, I wonder? I hope I haven't broken anything.

zherczeg:
I still see the "squash and merge" button, but I don't click on it to avoid even more chaos.

PhilipHazel:
I now see that too, but likewise, I'm holding back. @NWilson what do you think? Is it safe to try merging again?

NWilson:
I can't see what buttons you're seeing... but why not? It can't hurt to keep clicking until it merges.

PR #591, opened 2024-12-02
Improve coverage of escapes in character classes

NWilson:
One switch case in pcre2_compile.c is not covered by any tests (the ERR7 default).

Add tests covering all the possible escapes, and fix a couple that weren't working (ESC_k and ESC_ub).

Also fix some empty statements causing coverage failures in pcre2_match.c.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -1628,6 +1628,7 @@ else
             hptr >= ptrend ||    /* Hit end of input */
             *hptr != CHAR_RIGHT_CURLY_BRACKET)  /* No } terminator */
           {
+          if (isclass) break; /* In a class, just treat as '\u' literal */
           escape = ESC_ub;    /* Special return */
           ptr++;              /* Skip { */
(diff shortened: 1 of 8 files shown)


src/pcre2_compile.c
```diff
@@ -4342,9 +4348,20 @@ while (ptr < ptrend)
 #endif
           break;  /* End \P and \p */
 
-          default:    /* All others are not allowed in a class */
+          /* All others are not allowed in a class */
+
+          default:
+          PCRE2_DEBUG_UNREACHABLE();
```
zherczeg, line 4354: Not allowed or not possible?

NWilson: Not possible (as far as I can tell). I have gone through `check_escape()` and made a list of every ESC_ code it can return (when passed `inclass=TRUE`), and added a case statement and a test for all of those. There should be no other escapes allowed, so I added a catch-all default which only triggers in Debug builds.

zherczeg (approved):
LGTM

PR #592, opened 2024-12-03
Auto-format and minimal cleanup to CMake

NWilson:
I haven't tackled any controversial steps in this PR - simply tidying the formatting.

I have used the `gersemi` tool, which simply "does its thing". I have additionally renamed a few variables to match standard casing conventions (but I am aware that some lowercased variables are used, for example in package-config files, and have left those alone).


NWilson:
It's unpleasantly inconsistent to look at. Previous tickets (such as #115) have mentioned that the style and capitalisation don't match CMake standards.

The history is still all there. It's not too hard to make `git blame` skip the commit ... if anyone does ever need to do any archaeology.

zherczeg:
I am neutral for this change. @carenas what do you think?

zherczeg (approved):
LGTM

Issue #576, opened 2024-11-22
Two minor correctness bugs found by Clang Analyzer


NWilson:
I have a PR open which adds Clang's very good static analyzer to our reports.

This has flagged two issues, one of which appears to be a genuine bug:

* In pcre2grep, the `end_of_line` function is naughty! It looks at the leading byte of a UTF-8 sequence to find out how many bytes to read, then reads them without checking whether the buffer contains enough bytes! It appears to be a read-past-the-end bug if the input ends in a truncated UTF-8 sequence.
* In pcre2posix, the `pcre2_regerror` has a NULL check for `preg`, and the function does support passing in a NULL `preg`. But further down in the function, it's used unconditionally in `message_len(message, (int)preg->re_erroffset)`, and clang reckons that a NULL dereference (might) be possible there. I believe that this is bogus, however, and clang just isn't quite clever enough to work it out; but someone should probably double-check that reasoning.

PR #593, opened 2024-12-03
Suppress some minor analysis warnings

NWilson:
One of these appears in Coverity's dashboard; the rest are from clang-scan.

See #576

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -7285,7 +7285,6 @@ for (;; pptr++)
     single-char opcodes. */
 
     reqvary = (repeat_min == repeat_max)? 0 : REQ_VARY;
-    op_type = 0;
 
     /* Adjust first and required code units for a zero repeat. */
 
@@ -7848,7 +7847,8 @@ for (;; pptr++)
           }
         else
           {
-          /* Come here from just above with a character in mcbuffer/mclength. */
+          /* Come here from just above with a character in mcbuffer/mclength.
+          You must also set op_type before the jump. */
           OUTPUT_SINGLE_REPEAT:
           prop_type = prop_value = -1;
           }
--- a/src/pcre2_compile_class.c
+++ b/src/pcre2_compile_class.c
@@ -245,6 +245,7 @@ while (c <= end)
     buffer[0] = new_start;
     buffer[1] = new_end;
     buffer += 2;
+    (void)buffer;
     }
   return result;
 }
@@ -330,6 +331,7 @@ while (*p != NOTACHAR)
     buffer[0] = start;
     buffer[1] = get_highest_char(options);
     buffer += 2;
+    (void)buffer;
     }
 
   return result;
@@ -745,6 +747,7 @@ for (c = 0; c < 256; c++)
   {
   prop = GET_UCD(c);
   set_bit = FALSE;
+  (void)set_bit;
 
   switch (ptype)
     {
--- a/src/pcre2_dfa_match.c
+++ b/src/pcre2_dfa_match.c
@@ -3070,7 +3070,7 @@ for (;;)
         if (codevalue == OP_BRAPOSZERO)
           {
           allow_zero = TRUE;
-          codevalue = *(++code);  /* Codevalue will be one of above BRAs */
+          ++code;  /* The following opcode will be one of the above BRAs */
           }
         else allow_zero = FALSE;
 
--- a/src/pcre2grep.c
+++ b/src/pcre2grep.c
@@ -3059,7 +3059,6 @@ while (ptr < endptr)
         if (group_separator != NULL)
           fprintf(stdout, "%s%s", group_separator, STDOUT_NL);
         hyphenpending = FALSE;
-        hyphenprinted = TRUE;
         }
 
       /* Now print the matching line(s); ensure we set hyphenpending at the end
@@ -3874,8 +3873,12 @@ else
   filename = name;
   }
 
-while ((patlen = sizeof(buffer)) && read_pattern(buffer, &patlen, f))
+while (TRUE)
   {
+  patlen = sizeof(buffer);
+  if (!read_pattern(buffer, &patlen, f))
+    break;
+
   if (!posix_pattern_file)
    {
    while (patlen > 0 && isspace((unsigned char)(buffer[patlen-1]))) patlen--;
--- a/src/pcre2posix.c
+++ b/src/pcre2posix.c
@@ -94,6 +94,7 @@ changed. This #define is a copy of the one in pcre2_internal.h. */
 
 #include "pcre2.h"
 #include "pcre2posix.h"
+#include "pcre2_util.h"
 
 /* Table to translate PCRE2 compile time error codes into POSIX error codes.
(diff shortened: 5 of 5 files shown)


CMakeLists.txt
```diff
@@ -190,7 +190,7 @@ CHECK_C_SOURCE_COMPILES(
 
 CHECK_C_SOURCE_COMPILES(
   "#include <stddef.h>
-   int main(void) { int a,b; size_t m; __builtin_mul_overflow(a,b,&m); return 0; }"
+   int main(void) { int a=0, b=0; size_t m; __builtin_mul_overflow(a,b,&m); return 0; }"
```
NWilson, line 193: Harmless. Stop a spammy warning about using uninitialised values.

src/pcre2_compile.c
```diff
@@ -7102,7 +7102,6 @@ for (;; pptr++)
     single-char opcodes. */
 
     reqvary = (repeat_min == repeat_max)? 0 : REQ_VARY;
-    op_type = 0;
```
NWilson, line 7105: Clang is correct here. The assignment to `op_type` is provably a dead assignment. What's more... I reckon it hurts rather than helps, because if there's a branch in the code that _should_ be assigning to op_type but forgets, this would suppress warnings about use-of-uninitialised!

src/pcre2_compile_class.c
```diff
@@ -245,6 +245,7 @@ while (c <= end)
     buffer[0] = new_start;
     buffer[1] = new_end;
     buffer += 2;
+    (void)buffer;
```
NWilson, line 248: We have some clearly useless assignments here, but they're OK to keep, because they ensure the buffer is correctly tracked if someone extends the function later. Easy suppression.

src/pcre2_dfa_match.c
```diff
@@ -3076,7 +3076,7 @@ for (;;)
         if (codevalue == OP_BRAPOSZERO)
           {
           allow_zero = TRUE;
-          codevalue = *(++code);  /* Codevalue will be one of above BRAs */
+          ++code;  /* The following opcode will be one of the above BRAs */
```
NWilson, line 3079: Clang is correct. We are writing a value to `codevalue` which is provably never used.

src/pcre2grep.c
```diff
@@ -3874,7 +3873,7 @@ else
   filename = name;
   }
 
-while ((patlen = sizeof(buffer)) && read_pattern(buffer, &patlen, f))
+while ((patlen = sizeof(buffer), read_pattern(buffer, &patlen, f)))
```
NWilson, line 3876: Stupid coverity. It seems unable to reason about the `(patlen = sizeof(buffer)) &&` expression, which assigns to patlen, and then unconditionally continues (because `sizeof(buffer)` is always > 0).

src/pcre2posix.c
```diff
@@ -210,6 +211,8 @@ else
   ret = (int)len;
   }
 
+PCRE2_ASSERT(len > 0 || preg != NULL);
```
NWilson, line 214: Adding an assertion here lets Clang reason about when `preg` is NULL. There was no bug, but static analysis wasn't able to work out that checking `len` was equivalent to checking `preg != NULL`.

CMakeLists.txt
carenas: Not harmless; they were intentionally uninitialized so they wouldn't be eliminated by an over eager optimizing compiler.

NWilson: ? But this file is never executed. We are simply testing whether it compiles to an object file.

src/pcre2grep.c
NWilson: `-Wcomma` allows assignments on the LHS of a comma operator, so this should be accepted.

I'm not sure the original code _was_ clearer anyway. We're not logically doing an `&&`: we want to do something with side-effects on the left, ignore whatever it evaluates to, and then execute the right.

CMakeLists.txt
carenas: AFAIK dead code elimination happens at compile time. If I remember correctly at keast with AIX's xlc it will result in a false positive for having `__builtin_mul_overflow` which then will result in a FTBFS problem.

src/pcre2grep.c
zherczeg: I would rewrite it to `while (TRUE)`, set variable, break if read_pattern fails. This is not a nice code.

NWilson: Good point; I've done Zoltan's improvement.

zherczeg:
@carenas are the changes ok?

NWilson:
I think it's OK to merge. Carlo's only request was that there (might be, unconfirmed) a `-Wcomma` warning, but I've removed the comma.

carenas:
Apologies for the delay, validating the changes was indeed made more difficult by the rebasing of these changes, but it seems that another concern I had might had also sneaked in.

It was previously a regression that broke building in AIX with xlc, because it will falsely recognize the compiler as supporting a builtin that it doesn't have because the detection code was optimized out because it was deemed uneeded and without side effects.

carenas:
got an AIX 7.1 system to validate, and seems my concerns were unfunded, eventhough there might be still problems with `make dist` as a snapshot from HEAD failed with:

```
carenasgm@gcc111:[/home/carenasgm/pcre2-10.45-DEV]cat RunGrepTest.log
Testing pcre2grep version 10.45-DEV 2024-06-09
Testing pcre2grep main features
--- ./testdata/grepoutput       2024-12-08 04:13:29.000000000 -0800
+++ testtrygrep 2024-12-08 04:22:23.000000000 -0800
@@ -496,10 +496,7 @@
  ./testdata/grepinput:456
  ./testdata/grepinput3:0
  ./testdata/grepinput8:0
- ./testdata/grepinputBad8:0
- ./testdata/grepinputBad8_Trail:0
  ./testdata/grepinputM:0
- ./testdata/grepinputUN:0
  ./testdata/grepinputv:1
  ./testdata/grepinputx:0
  RC=0
@@ -728,8 +725,6 @@
  ---------------------------- Test 96 -----------------------------
  ./testdata/grepinput3
  ./testdata/grepinput8
- ./testdata/grepinputBad8
- ./testdata/grepinputBad8_Trail
  ./testdata/grepinputx
  RC=0
  ---------------------------- Test 97 -----------------------------
@@ -841,10 +836,7 @@
  testdata/grepinput:469
  testdata/grepinput3:0
  testdata/grepinput8:0
- testdata/grepinputBad8:0
- testdata/grepinputBad8_Trail:0
  testdata/grepinputM:2
- testdata/grepinputUN:0
  testdata/grepinputv:3
  testdata/grepinputx:6
  TOTAL:480
@@ -863,10 +855,7 @@
  469
  0
  0
- 0
- 0
  2
- 0
  3
  6
  480
@@ -874,9 +863,6 @@
  ---------------------------- Test 118 -----------------------------
  testdata/grepinput3
  testdata/grepinput8
- testdata/grepinputBad8
- testdata/grepinputBad8_Trail
- testdata/grepinputUN
  RC=0
  ---------------------------- Test 119 -----------------------------
  123
FAIL RunGrepTest (exit status: 1)
```

PR #595, opened 2024-12-04
Apply some security hardening, as recommended by OSSF scorecard

NWilson:
* Use hashes to pin versions of third-party Actions used in CI
* Use GitHub's dependabot to do a monthly update of the Actions
* Restrict all Workflows to use a read-only token (except for the steps which need to upload a SARIF file)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: monthly
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,8 @@
-
 name: Build
 on: [push, pull_request]
+
+permissions:
+  contents: read
 
 jobs:
   linux:
@@ -13,7 +15,7 @@ jobs:
           sudo apt-get -qq install zlib1g-dev libbz2-dev
 
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           submodules: true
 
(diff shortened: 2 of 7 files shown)


zherczeg:
Will this make our life easier or more difficult? Is this needed for CI security, or what is the actual gain here?

NWilson:
It's needed for CI security. As it stands, we depend on some third party github actions, and if someone pushed malicious code to those repos, they would be able to perform pretty much arbitrary actions on our PCRE2 repo, the next time CI runs. 

It should be much more locked down.

NWilson:
These changes won't make our life more difficult (nor any easier).

zherczeg:
This is disturbing. Can a CI write the repository in any way? Where can I disable that?

NWilson:
Yes, sadly. The GitHub Actions have a very permissive token by default - a CI job can happily check out the code, make some changes, and push them to the master branch (and even rewrite history on master). The token that's used by the git checkout job can happily push code too.

And worse, it can even perform other actions on the GitHub organisation using the HTTP API (or GitHub CLI).

Adding `permissions: contents: read` locks down the runner to only have read access to the "contents" (code and issues) of the repo.

zherczeg:
Thank you for the explanation.

I found this in Settings/Actions/General

Workflow permissions

Choose the default permissions granted to the GITHUB_TOKEN when running workflows in this repository. You can specify more granular permissions in the workflow using YAML. [Learn more about managing permissions.](https://docs.github.com/actions/reference/authentication-in-a-workflow#modifying-the-permissions-for-the-github_token)

Read and write permissions
Workflows have read and write permissions in the repository for all scopes.

Read repository contents and packages permissions
Workflows have read permissions in the repository for the contents and packages scopes only.

@PhilipHazel may I change this to read-only? Honestly, I am shocked that writing is the default.

NWilson:
I've broken the Scorecard scanning - because it's designed to run on master but not on PRs! So it wasn't really possible to discover problems before merging. Hmm.

I'll sort it out, don't worry.

PR #596, opened 2024-12-05
Optimize out unnecessary eclass bitmaps

zherczeg:
Large patch, contains code moves and reworks.


src/pcre2_compile_class.c
```diff
@@ -2276,12 +2288,11 @@ return TRUE;
 /* This function consumes unary prefix operators. */
 
 static BOOL
-compile_class_unary(uint32_t options, uint32_t xoptions, BOOL negated,
+compile_class_unary(eclass_context *context, BOOL negated,
```
zherczeg, line 2291: This could be merged with compile_class_juxtaposition.

src/pcre2_compile_class.c
```diff
@@ -2151,40 +2182,22 @@ switch (meta)
...
 
     PCRE2_ASSERT(*code_start == OP_XCLASS);
     *code_start = pop_info->op_single_type = ECL_XCLASS;
 
     PCRE2_ASSERT(code - code_start >= 1 + LINK_SIZE + 1);
     flags = code_start[1 + LINK_SIZE];
 
-    if ((flags & XCL_MAP) != 0)
+    if ((flags & XCL_MAP) != 0 ||
+        (*context->eclass_start & CLASS_ECLASS_HAS_BITMAP) != 0)
```
NWilson, line 2193: I don't think you can check the CLASS_ECLASS_HAS_BITMAP flag yet.

You haven't yet scanned the entire extended class, so you don't know the final value of this. You can only check that flag at the very very end.

What you want to do is to always copy the bits from `cb->classbits` to `pop_info` (unconditionally), and always clear the XCL_MAP flag; but conditionally do `if (flags & XCL_MAP) eclass_start |= CLASS_ECLASS_HAS_BITMAP`

src/pcre2_compile_class.c
NWilson: It's clear as it is. I don't mind either way.

src/pcre2_compile_class.c
```diff
@@ -2467,33 +2475,229 @@ applications.
...
+             META_CODE(*ptr) == META_CLASS_NOT);
 
-if (*ptr++ == (META_CLASS_NOT | CLASS_IS_ECLASS))
+if (META_CODE(*ptr) == META_CLASS_NOT)
   negated = !negated;
 
 /* Because it's a non-empty class, there must be an operand at the start. */
-if (!compile_class_binary_loose(options, xoptions, negated, &ptr, &code,
-                                pop_info, errorcodeptr, cb, lengthptr))
+(*pptr)++;
```
NWilson, line 2496: This `(*pptr++)` is placed confusingly under the comment now. But this skip has nothing to do with the operand, it's the META_CLASS we're skipping.

src/pcre2_compile_class.c
```diff
@@ -2292,15 +2303,13 @@ while (*ptr == META_ECLASS_NOT)
   negated = !negated;
   }
 
+*pptr = ptr;
```
NWilson, line 2306: Moving up the pptr assignment breaks the pattern with the other functions. Although, it looks like it behaves the same.

src/pcre2_compile_class.c
```diff
@@ -2467,33 +2475,229 @@ applications.
...
+  compile_block *cb, PCRE2_SIZE *lengthptr)
+{
+eclass_context context;
+eclass_op_info op_info;
+PCRE2_SIZE previous_length = (lengthptr != NULL)? *lengthptr : 0;
+PCRE2_UCHAR *code = *pcode;
+PCRE2_UCHAR *previous;
+BOOL allbitsone = TRUE;
+
+context.eclass_start = *pptr;
```
NWilson, line 2518: Storing this bit in the META code seems very dangerous. It carries the risk that different branches will be taken on the two passes through the code.

The flag you want should just be a BOOL on the eclass_context, so it's got a fresh state on each pass.

src/pcre2_compile_class.c
```diff
@@ -2467,33 +2475,229 @@ applications.
...
+context.cb = cb;
+
+previous = code;
+*code++ = OP_ECLASS;
+code += LINK_SIZE;
+*code++ = 0;  /* Flags, currently zero. */
+if (!compile_eclass_nested(&context, FALSE, pptr, &code, &op_info, lengthptr))
+  return FALSE;
+
+if (lengthptr != NULL)
```
NWilson, line 2531: Moving this code in here is great. I like the way you did it.

Are there any changes here in the code you moved?

I can't see any changes (apart from changing need_map to use CLASS_ECLASS_HAS_BITMAP)

NWilson (commented):
Nice! That much easier than I'd feared it would be. Thank you.

src/pcre2_compile_class.c
```diff
@@ -2060,6 +2086,7 @@ switch (meta)
     {
     *code++ = pop_info->op_single_type = ECL_ANY;
     memset(pop_info->bits.classbits, 0xff, 32);
+    *context->eclass_start |= CLASS_ECLASS_HAS_BITMAP;
```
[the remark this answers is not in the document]
zherczeg: I did this for the safe side. It is better if we don't need it.

src/pcre2_compile_class.c
zherczeg: It is an optimization. There is no need to cache the pptr.

src/pcre2_compile_class.c
zherczeg: That was the whole point of the patch. Although I realized you inject the bitmap at the end, so caching this value is likely unnecessary.

src/pcre2_compile_class.c
zherczeg: No. The context is initialized at the beginning.

src/pcre2_compile_class.c
```diff
@@ -2143,6 +2173,7 @@ switch (meta)
     if (lengthptr != NULL)
       *lengthptr += code - (code_start + 1);
     code = code_start + 1;
+    *context->eclass_start |= CLASS_ECLASS_HAS_BITMAP;
```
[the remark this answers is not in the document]
zherczeg: I am not sure I follow. The OP_ALLANY is handled above, and the OP_CLASS with all zero bits can be optimized. Nothing else can be optimized. Am I right?

src/pcre2_compile_class.c
```diff
@@ -2151,40 +2182,22 @@ switch (meta)
...
-    PCRE2_SIZE put_length;
 
     PCRE2_ASSERT(*code_start == OP_XCLASS);
     *code_start = pop_info->op_single_type = ECL_XCLASS;
 
     PCRE2_ASSERT(code - code_start >= 1 + LINK_SIZE + 1);
     flags = code_start[1 + LINK_SIZE];
 
-    if ((flags & XCL_MAP) != 0)
+    if ((flags & XCL_MAP) != 0 ||
```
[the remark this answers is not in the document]
zherczeg: I have replaced it with a BOOL* pointer.

src/pcre2_compile_class.c
zherczeg: It was designed with the bit is preserved in the META_ECLASS in mind. The new method requires the copy all the time as you said.

zherczeg:
I have resolved the some conversions, but feel free to reopen them if they are not resolved. There is one comment I don't understand. Let me know if other changes are needed.

src/pcre2_compile_class.c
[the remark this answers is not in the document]
NWilson: The condition to leave out the map is: if you emit ECL_ANY, does the pop_info have an all-ones bitmap? And if you emit an ECL_NONE, does the pop_info have an all-zeros bitmap?

One of the OP_NCLASS branches could hit that condition, so doesn't always need a bitmap.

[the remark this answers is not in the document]
zherczeg: If ECL_ANY is set, and all bits are set, the compile_class_not_nested already converts the XCLASS to OP_ALLANY, and handled earlier. Hence the ECL_NONE and all bits are cleared are the remaining case. Then my question was correct.

zherczeg:
I hope I fixed everything. I noticed that it is not possible to tell the difference in the byte code dump that if an xclass has no bitset, or all bits in the bitset are zero. Maybe an `(allzero)` could be added after `[]` if a bitset is present, and everything is 0. I can do this in another patch.

src/pcre2_compile_class.c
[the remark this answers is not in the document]
NWilson: Great, we agree! You've special-cased the OP_NCLASS case, which (out of OP_CLASS & OP_NCLASS) is the only one where it's possible to _not_ set need_bitmap.

src/pcre2_compile_class.c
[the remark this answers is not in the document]
NWilson: Thanks, that makes me happy!

src/pcre2_compile_class.c
```diff
@@ -2125,6 +2155,7 @@ switch (meta)
     pop_info->length = 1;
     *code_start = pop_info->op_single_type = ECL_ANY;
     memset(pop_info->bits.classbits, 0xff, 32);
+    context->needs_bitmap = TRUE;
```
NWilson, line 2158: I believe that this OP_ALLANY → ECL_ANY branch does not need to set needs_bitmap.

src/pcre2_compile_class.c
```diff
@@ -1752,6 +1774,8 @@ whole class was negated and whether there were negative specials such as \S
 (non-UCP) in the class. Then copy the 32-byte map into the code vector,
 negating it if necessary. */
 
+PCRE2_ASSERT(classbits == cb->classbits.classbits);
```
NWilson, line 1777: We could lose this assertion if you want. We only ever assign once to the classbits variable (it's const) so there's really not much to check.

NWilson:
> I tried one of your suggestions. Original code:
> 
> ```
>   re> /[\pL&&[^\x00-\xFF]]/B,utf,alt_extended_class
> ------------------------------------------------------------------
>         Bra
>         [\p{L}]
>         Ket
>         End
> ------------------------------------------------------------------
> data> a
>  0: a
> ```
> 
> This should not happen. Interestingly, this patch fixes it. Not sure what is happening.

Ouch! I have diagnosed the cause of this, in the code that's currently on `master`.

There are two cases where we emit a bitmap: there's the branch (previously in pcre2_compile.c) which emits an actual OP_ECLASS, and there's the branch that converts a single ECL_XCLASS into an actual OP_XCLASS. Both of these were using the same check: `need_map = !allbitszero;` 

For OP_ECLASS, that (was) correct - no bitmap meant the same as all-bits-zero. But for OP_XCLASS it was not true. So there was an existing bug in my code you fixed.

The new logic checks `context.needs_map` in both branches, which is good - because OP_ECLASS and OP_XCLASS now have the same semantics for when a map is not provided.

Great! I'm glad I understand the bug, and can verify that you've properly fixed it, not just hidden it.

NWilson (approved):
I've left two additional small comments. Thank you for working on this! You can merge whenever you're ready (with or without those small suggestions).

src/pcre2_compile_class.c
zherczeg: True!

src/pcre2_compile_class.c
zherczeg: Ok. I have added a comment about it as well.

zherczeg:
I am happy that one less bug in the code. The test case was your suggestion, I just tried it.

PR #602, opened 2024-12-07
Change READ_CHAR_MAX to maximum unsigned value

zherczeg:
This is just one line change, but I want to make sure it does not break anything.

### DIFF
--- a/src/pcre2_jit_char_inc.h
+++ b/src/pcre2_jit_char_inc.h
@@ -639,7 +639,7 @@ if (category_list == UCPCAT_ALL)
 if (*cc != XCL_END)
   {
 #if defined SUPPORT_UNICODE && (PCRE2_CODE_UNIT_WIDTH == 8 || PCRE2_CODE_UNIT_WIDTH == 16)
-  if (common->utf && compares == 0 && (status & XCLASS_IS_ECLASS))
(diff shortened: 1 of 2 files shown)


NWilson:
I don't understand the implications of this. Did it affect the 32-bit library?

zherczeg:
No. It was wrongly defined (probably to fit to an int), but it was harmless since there is no optimization characters > 0x7fffffff. However it is better this way.

src/pcre2_jit_char_inc.h
```diff
@@ -639,7 +639,7 @@ if (category_list == UCPCAT_ALL)
 if (*cc != XCL_END)
   {
 #if defined SUPPORT_UNICODE && (PCRE2_CODE_UNIT_WIDTH == 8 || PCRE2_CODE_UNIT_WIDTH == 16)
-  if (common->utf && compares == 0 && (status & XCLASS_IS_ECLASS))
+  if (common->utf && compares == 0 && !(status & XCLASS_IS_ECLASS))
```
zherczeg, line 642: It was a typo. Luckily, it has no other effect than a deoptimization (xclass case is not optimized, the eclass case always ignored the computed value). Fortunately I noticed it.

NWilson (approved):

PR #605, opened 2024-12-09
Remove PT_ANY support from matchers

zherczeg:
Advantage: code size reduction / code simplification.
Disadvantage: `\P{Any}` is converted to `[]` which uses more space. I think `\P{Any}` is very rarely used. `(*F)` is better and shorter. If empty brackets are enabled `[]` is even shorter.
Note: we have already convert many empty constructs to `[]`, from now `\P{Any}` just follows them.
Note2: since `\P{Any}` uses less space, all `[]` could be converted to `\p{Any}`, but it requires unicode support. My opinion is that empty classes does not worth much optimizations.


src/pcre2_compile_class.c
```diff
@@ -408,7 +408,7 @@ while (TRUE)
         case ESC_p:
         case ESC_P:
         ptr++;
-        if (meta_arg == ESC_p && *ptr == PT_ANY)
+        if (meta_arg == ESC_p && (*ptr >> 16) == PT_ANY)
```
zherczeg, line 411: This was a bug in master.

testdata/testoutput5
```diff
@@ -1910,6 +1910,24 @@ No match
...
+        #
+        []
+        #
+        AllAny
+        #
+        []
+        #
+        AllAny
+        #
+        [c]
```
zherczeg, line 1926: This is not caught by the `[c]` -> `c` optimization, but I think it is not worth to make it more complex.

src/pcre2_compile.c
```diff
@@ -8055,9 +8055,16 @@ for (;; pptr++)
       /* The special case of \p{Any} is compiled to OP_ALLANY so as to benefit
       from the auto-anchoring code. */
```
NWilson, line 8056: Could update the comment for `\P{Any} → OP_CLASS` too...

testdata/testoutput5
NWilson: That seems fair. I also deliberately skipped those optimisations too for things like `(?[ [ab] - [b] ])` because I felt it was just too obscure.

NWilson (approved):
Nice. I am happy with using empty `OP_CLASS` (with zero bits). If we care about the wasted bytes, we can always add an "OP_NONE" in just a few lines of code - but there's no real need.

zherczeg:
Patch is updated, and added more tests.

Introducing an OP_NONE is not a trivial task, since it needs specialized code paths in the normal/dfa/jit matchers. I don't see any reasonable purpose to use empty classes, so I think they are very rarely used and not worth any specializations in the matchers. The `(*F)` is always a better option.

PR #618, opened 2024-12-10
Fix null-dereference bug in pcre2_substitute

NWilson:
Avoid one crash introduced with recent changes to substitute code as well as clarify what the expected offset value should be when overflowing the provided buffer.


NWilson:
@carenas I took your PR and split it in half for review. These are your changes.

Are you happy for me to merge this half of the PR first?

carenas (approved):
it might be a good idea to update the commit message after squashing though to avoid confusion with the missing NUL

NWilson:
Yes, I always set the commit message to match the PR's "description" at the top. I find it much quicker and easier to add to the PR message as I go, than to re-push a PR just to update the commit message.

Thanks for discovering this bug Carlo!

PR #619, opened 2024-12-11
Rework memory management

zherczeg:
Small code improvement, limit checks.

This patch is another preparation patch, since `private_data_ptrs` is available earlier than before.


src/pcre2_jit_compile.c
```diff
@@ -13107,9 +13104,25 @@ ccend = bracketend(common->start);
...
+/* Allocate space for temporary data structures. */
+private_data_length = ccend - common->start;
+/* The chance of overflow is very low, but might happen on 32 bit. */
+if (private_data_length > ~(sljit_uw)0 / sizeof(sljit_s32))
+  return PCRE2_ERROR_NOMEMORY;
+
+private_data_length *= sizeof(sljit_s32);
+/* Align to 32 bit. */
+total_length = ((re->top_bracket + 1) + (sljit_uw)(sizeof(sljit_s32) - 1)) & ~(sljit_uw)(sizeof(sljit_s32) - 1);
+if (~(sljit_uw)0 - private_data_length <= total_length)
```
NWilson, line 13116: nit: I think this could be `<` rather than `<=`

src/pcre2_jit_compile.c
```diff
@@ -13202,22 +13210,28 @@ SLJIT_ASSERT(common->early_fail_start_ptr <= common->early_fail_end_ptr);
...
   }
 
 compiler = sljit_create_compiler(allocator_data);
 if (!compiler)
   {
-  SLJIT_FREE(common->optimized_cbracket, allocator_data);
   SLJIT_FREE(common->private_data_ptrs, allocator_data);
+  if (common->has_then)
+    SLJIT_FREE(common->then_offsets, allocator_data);
   return PCRE2_ERROR_NOMEMORY;
```
NWilson, line 13235: With all these repeated FREE statements, it's common to have a block at the end of the function with the cleanup code, which you can jump to with goto. Saves duplicating the cleanup. The cleanup block can have multiple jump targets, with fallthrough, so it cleans up stuff in reverse order to what was allocated. Or it can just check each pointer for NULL and clean up all the ones that have been alloced.

NWilson (approved):
Looks fine. The comments are just a suggestion.

src/pcre2_jit_compile.c
```diff
@@ -13202,22 +13210,28 @@ SLJIT_ASSERT(common->early_fail_start_ptr <= common->early_fail_end_ptr);
 if (private_data_size > 65536)
```
NWilson, line 13210: I noticed on one of your branches, you have modified the `if (private_data_size > 65536)` check.

On master, it returns `PCRE2_ERROR_NOMEMORY` but at one time you changed it to `PCRE2_ERROR_JIT_UNSUPPORTED`.

It seems odd to report "not enough memory" when you reach that private_data_size limit, since it's not an issue with the machine's RAM, but it appears to be an implementation limitation.

src/pcre2_jit_compile.c
zherczeg: JIT does not use goto at the moment. But it is reasonable here. I expect compilers to do CSE. But you cannot trust in all compilers. Not sure what is the best. Shall we introduce goto?

src/pcre2_jit_compile.c
zherczeg: Since max is odd number, and the other two are even, the == cannot happen. Ok, lets change it.

src/pcre2_jit_compile.c
zherczeg: I don't remember. I have change it.

zherczeg: It changes the expected file:
```
-JIT compilation was not successful (no more memory)
+JIT compilation was not successful (feature is not supported by the JIT compiler)
```
Both seem not entirely correct.

src/pcre2_jit_compile.c
NWilson: I don't mind; if you don't want to use goto, then duplicating the FREE calls is OK. It's your code - it's only a maintenance concern.

src/pcre2_jit_compile.c
zherczeg: That happens regardless. Introducing a new error code would be the best.

NWilson: If it falls back regardless, then I guess it doesn't matter. Just change it to "PCRE2_ERROR_JIT_UNSUPPORTED" if it's simply a size limitation in the JIT engine.

zherczeg: Ok. Doing these things could be low hanging fruits for somebody.

NWilson (approved):

PR #620, opened 2024-12-11
Add Solaris and FreeBSD to our CI

NWilson:
These are only run after a PR is merged, not every time you push to a PR.

They use a third-party GitHub Action that starts up a VM using Qemu.

The Solaris one is cheating a little - I'm using GCC rather than Oracle Developer Studio's `cc` compiler, because that's a real pain to install. On the other hand, most people probably use GCC on Solaris now anyway, so maybe it is a fair test of whether our Solaris support is OK.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,7 +133,7 @@ jobs:
         run: ./autogen.sh
 
       - name: Build & test
-        uses: vmactions/freebsd-vm@v1
+        uses: vmactions/freebsd-vm@debf37ca7b7fa40e19c542ef7ba30d6054a706a4 # v1.1.5
         with:
           usesh: true
           run: |
@@ -156,7 +156,7 @@ jobs:
         run: ./autogen.sh
 
       - name: Build & test
-        uses: vmactions/solaris-vm@v1
+        uses: vmactions/solaris-vm@a89b9438868c70db27e41625f0a5de6ff5e90809 # v1.1.0
         with:
           usesh: true
           prepare: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -122,7 +122,7 @@ jobs:
   freebsd:
     name: FreeBSD
     runs-on: ubuntu-latest
-    if: github.event_name != "pull_request"
+    if: github.event_name != 'pull_request'
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -119,6 +119,55 @@ jobs:
       - name: Test
         run: cd build && ctest -C Release -j3 --output-on-failure
 
+  freebsd:
+    name: FreeBSD
+    runs-on: ubuntu-latest
+    if: github.event_name != "pull_request"
```
carenas, line 125: presume trying to save some CPU time?

FWIW we could also use `cirrus CI` which provides native FreeBSD runners (hosted in google cloud) instead

.github/workflows/build.yml
```diff
@@ -119,6 +119,55 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
carenas, line 128: why this version instead of "latest"?

carenas (approved):
I have a more convoluted version of this in the "solaris" branch in my fork that I didn't propose because it is too inefficient CPU wise, and we don't really know if Microsoft might eventually tighten the screws in their "free" plan.

This is IMHO a good place were a corporate sponsor and alternative CI providers might help, but will likely require more infrastructure work (including maybe setting a "non profit" or piggybacking on the ones that are around to help projects like us)

.github/workflows/build.yml
NWilson: True. But I don't see a compelling need to switch, for now.

We'd still need a VM for Solaris.

.github/workflows/build.yml
NWilson: In a previous PR, I moved to using pinned dependency versions, with Dependabot updating them periodically. This is a more secure, locked-down use of third-party code.

.github/workflows/build.yml
```diff
@@ -119,6 +119,55 @@ jobs:
...
+        uses: vmactions/solaris-vm@a89b9438868c70db27e41625f0a5de6ff5e90809 # v1.1.0
+        with:
+          usesh: true
+          prepare: |
+            pkg install developer/gcc system/header
+
+          run: |
+            set -e
+            ./configure --enable-jit --enable-pcre2-16 --enable-pcre2-32
+            gmake -j3 CPPFLAGS='-Wall -Wextra -Werror'
```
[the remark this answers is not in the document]
NWilson: It failed with Solaris `make`, so I just used gmake. I didn't consider it important what make system works - as long as the code compiles & runs, we're fine.

But if you prefer using Solaris make, I'll try and fix it.

I've also decided to have an attempt at using Solaris `cc` rather than GCC.

NWilson:
OK, ready for review again. It's now using Oracle's SunStudio compiler (Sun `cc`), as well as their Solaris `make`.

NWilson:
@diizzyy Daniel, just to let you know, I'm adding a FreeBSD runner to our master CI.

diizzyy:
@NWilson 
That's great, thanks! Are there any plans on adding CMake too?

NWilson:
I guess we could. However, we already have plenty of CI jobs using CMake, so we should be OK on that system. For the rarer unix platforms, it's simpler just to have one CI job that tests the lowest-common-denominator for the build; hence why I just tested the autotools build.

.github/workflows/build.yml
carenas: cirrusci provides their own VMs which perform better than GitHub's

PR #623, opened 2024-12-12
Add Bazel build to CI

NWilson:
I know absolutely no Bazel!

The existing Bazel build isn't implementing any of the Autoconf/CMake at all - not testing for header files, or setting any of the platform-specific macros.

That seems... less than ideal. But anyway, if users really want to use Bazel, then can do so. But perhaps the Bazel file should instead shell out to a "proper" build system like CMake, if that's what is required for actually performing system-specific configuration?

I have attempted to update the Bazel file the minimal amount that enables us to at least run the unit tests, and confirm that the build was successful.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,3 +1,0 @@
-common --experimental_enable_bzlmod
-build --incompatible_enable_cc_toolchain_resolution
-build --incompatible_strict_action_env
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -277,11 +277,7 @@ jobs:
   bazel:
     # Tests with: Bazel build system
     name: Bazel
-    strategy:
-      fail-fast: false
-      matrix:
-        os: ["ubuntu-latest", "windows-latest"]
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -279,22 +279,13 @@ jobs:
     name: Bazel
     runs-on: ubuntu-latest
     steps:
-      - name: Setup
-        run: |
-          sudo apt-get -qq install apt-transport-https curl gnupg -y
-          curl -fsSL https://bazel.build/bazel-release.pub.gpg | gpg --dearmor >bazel-archive-keyring.gpg
-          sudo mv bazel-archive-keyring.gpg /usr/share/keyrings
-          echo "deb [arch=amd64 signed-by=/usr/share/keyrings/bazel-archive-keyring.gpg] https://storage.googleapis.com/bazel-apt stable jdk1.8" | sudo tee /etc/apt/sources.list.d/bazel.list
-
-          sudo apt-get -qq update && sudo apt-get -qq install bazel
-
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           submodules: true
 
       - name: Build
-        run: bazel build :all
+        run: bazelisk build :all
 
       - name: Test
         run: srcdir=`pwd` pcre2test=`pwd`/bazel-bin/pcre2test ./RunTest
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -277,7 +277,11 @@ jobs:
   bazel:
     # Tests with: Bazel build system
     name: Bazel
-    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest", "windows-latest"]
+    runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
@@ -285,10 +289,10 @@ jobs:
           submodules: true
 
       - name: Build
-        run: bazelisk build :all
+        run: bazelisk build //...
 
       - name: Test
-        run: srcdir=`pwd` pcre2test=`pwd`/bazel-bin/pcre2test ./RunTest
+        run: bazelisk test //... --test_output=all
 
   heron:
     # Job to verify that the tasks performed by PrepareRelease have been done. It is
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -210,8 +210,8 @@ jobs:
     steps:
       - name: Setup
         run: |
-          sudo apt-get update
-          sudo apt-get install -y valgrind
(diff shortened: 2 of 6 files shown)


NWilson:
@zaucy @keith

I have absolutely no idea if these changes are OK.

.github/workflows/dev.yml
```diff
@@ -274,6 +274,31 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup
+        run: |
+          sudo apt-get -qq install apt-transport-https curl gnupg -y
+          curl -fsSL https://bazel.build/bazel-release.pub.gpg | gpg --dearmor >bazel-archive-keyring.gpg
+          sudo mv bazel-archive-keyring.gpg /usr/share/keyrings
+          echo "deb [arch=amd64 signed-by=/usr/share/keyrings/bazel-archive-keyring.gpg] https://storage.googleapis.com/bazel-apt stable jdk1.8" | sudo tee /etc/apt/sources.list.d/bazel.list
+
+          sudo apt-get -qq update && sudo apt-get -qq install bazel
```
zaucy, line 289: [`bazelisk`](https://github.com/bazelbuild/bazelisk) should be available on all github runners, no need to install! Usually I add a `.bazelversion` to the root of the repository and bazelisk will use that, but the [`USE_BAZEL_VERSION`](https://github.com/bazelbuild/bazelisk?tab=readme-ov-file#how-does-bazelisk-know-which-bazel-version-to-run) env var also works.

BUILD.bazel
```diff
@@ -53,24 +54,95 @@ cc_library(
...
+        "src/pcre2_printint.c",
+        "src/pcre2_tables.c",
+        "src/pcre2_ucd.c",
+        "src/pcre2_valid_utf.c",
+    ],
+    strip_include_prefix = "src",
+    visibility = ["//visibility:private"],
+)
+
 cc_binary(
```
zaucy, line 129: typically a test would be done through a `cc_test` and then ran with `bazel test //...`.

zaucy (commented):
It's been quite awhile since I added the initial bazel files. I'm happy to help out where I can!

.github/workflows/dev.yml
NWilson: OK, that's good to know. I've switched to using `baselisk`, which seems to work fine. It loads v8. I haven't pinned that version, we can just let it auto-update.

Alternatively, if you want to pin a specific version, it will be your job (or someone else) to keep it updated, rather than me.

NWilson:
I notice this build warning:

> WARNING: Option 'experimental_enable_bzlmod' is deprecated: Use --enable_bzlmod instead

Do you need any of the stuff in `.bazelrc`? Or is it OK to remove that file now?

I also wonder if the WORKSPACE file can be removed now.

BUILD.bazel
NWilson: I have granted you access to the repo.

Please, go ahead and make it work, if you can. I spent quite a while yesterday trying to do that, and totally gave up.

What we need to do is really, really simple, and would be a one-liner in any build system other than Bazel. We just need to run `pcre2test=PATH_TO_OUTPUT ./RunTest` in the source directory. The test passes if the RunTest script has a zero exit status. The RunTest script can be passed the path to the `pcre2test` binary, and it looks in the current directory for the `testdata/` folder. However, this seems super-hard to do in Bazel.

.github/workflows/dev.yml
zaucy: Using latest should be fine!

BUILD.bazel
zaucy: Thanks for letting me know how it works! A `sh_test` would be correct in that case. Bazel also needs to know about all a tests inputs when it runs and it does that throught the `data` attribute. I just globbed all the data in the `testdata` directory and it seems to work. Usually you would do it through bazel runfiles, but with how pcre2's file structure it _should_ work as is.

NWilson: Thank you. I tried for a bit to get it work on Windows too. We have a `RunTest.bat` file for Windows (so it shouldn't need `bash` or MSYS at all). But I couldn't work out how to execute a batch file in bazel! There's a `native_test` helper in Skylib, which will run an exe as a test, but that can't run a batch script.

I find this all very weird, that a supposedly mature build system won't let you define your tests as a script, or give you good tools for just "execute this".

I will merge this PR, if you are happy with the lack of testing on Windows.

zaucy: the `native_test` can actually run a `.bat` file (you just have to set the `out` attribute). I gave it a try, but it failed to find the `pcre2test.exe`. I'd have to look into it more to see how to properly find it.

If I get the time I could do a Windows follow-up, yeah. This is already a great improvement.

zaucy (approved):

BUILD.bazel
zaucy: the `sh_test` is _supposed_ to work for windows batch files as well, but I ran into this issue https://github.com/bazelbuild/bazel/issues/20345

.bazelrc
```diff
@@ -1,3 +0,0 @@
-common --experimental_enable_bzlmod
-build --incompatible_enable_cc_toolchain_resolution
-build --incompatible_strict_action_env
```
keith, line 3: deleting this last line does affect things, the first 2 probably don't matter if you updated bazel, since they both default to true as of bazel 7.x.

based on the warning you posted below you just needed to remove `experimental_` from the first one to solve that warning

MODULE.bazel
```diff
@@ -1,8 +1,9 @@
 module(
     name = "pcre2",
-    version = "10.40",
+    version = "10.45-DEV",
```
keith, line 3: note that in this repo the version here doesn't really matter, most repos just set them to `0` and then when submitting to the bazel-central-registry patch in the correct version number, that way you don't have to worry about updating this as part of the release process

.bazelrc
zaucy: the `.bazelrc` doesn't affect upstream users of pcre2 so it shouldn't matter since we're only testing against 8.x anyways

keith: yea, but this option does apply to that case. the last flag here reduces the chances of the local env affecting the build for example

MODULE.bazel
NWilson: OK, good to know. I don't mind setting it to the actual version number however.

BUILD.bazel
NWilson: OK. These Bazel issues are taking too much time for me to fix. I will merge this, without Windows support, and if you wish to add it later you are very welcome to open a PR for that.

NWilson: Thank you very much for your help in this PR!

NWilson:
Actually... I worked out the problem! I got the Windows tests working by adding `--enable_runfiles` to the Bazel commandline.

This is important because otherwise, you don't get the files created in the `testdata/` directly, even if you declare them in the `data= [...]` attribute. It defaults to true on Unix, false on Windows.

PR #626, opened 2024-12-14
Improve repeat support for plus

zherczeg:
Another patch to reduce char1 compilations, which can be quite large for xclasses/eclasses.

### DIFF
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -318,9 +318,9 @@ typedef struct char_iterator_backtrack {
     struct {
       unsigned int othercasebit;
       PCRE2_UCHAR chr;
-      BOOL enabled;
     } charpos;
   } u;
+  BOOL charpos_enabled;
 } char_iterator_backtrack;
 
 typedef struct ref_iterator_backtrack {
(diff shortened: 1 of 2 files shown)


NWilson (approved):
Well, I don't really follow all this logic, because I haven't taken the time to study it, but the changes look reasonable.

zherczeg:
Is there a way to see only the new differences? It looks to me the problem is that the new ASSERT-s are not failed.

NWilson:
Ooh, interesting. There is a Chrome & Firefox browser extension which is supposed to let you look at line-by-line changes, although I admit I haven't tried it yet.

There is this page here (linked from the CI status): https://app.codecov.io/gh/PCRE2Project/pcre2/pull/626?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=PCRE2Project

It suggests that although the patch coverage is "only" 95%, there are no actual changes in coverage - all the lines in the patch which are uncovered, were previously uncovered.

It probably is those ASSERTs. Those are never reachable (or the tests would fail, clearly), so I'll try to adjust the coverage builds to exclude all those lines.

I apologise again for the coverage spam. I still think it will be useful in future. One day we'll forget some important test cases, and it'll help us out.

PR #630, opened 2024-12-16
Add CI test to ensure installation manifest is correct

### LINKED ISSUE #617: Use CI to validate tarball & installation targets
CI should run `make distcheck`.

CI should build a tarball, unpack it, and check the file list against a "distribution manifest" of file names in `testdata/`.

CI should do both `./configure && DESTDIR=... make install` and `cmake --install . --prefix=...`, and again check the list of files which were actually installed against an "installation manifest" in `testdata/`.

I will do this before the 10.45 release. For the time being, I'm happy to do these checks manually pre-release, so it's not vital.

NWilson:
Fixes #617


.github/workflows/build.yml
```diff
@@ -192,6 +217,48 @@ jobs:
...
+          echo "XXX libtool"
+          which libtoolize
+          /usr/bin/libtoolize --version
+
+          ./autogen.sh
+
+          ls -l m4/
+
+          echo "XXX FIXING"
+          chmod ugo-x m4/* ltmain.sh
```
NWilson, line 249: I'm debugging something weird here.

On the GitHub Actions workflow, the files created by libtoolize are executable. They have some weird setup going on in a Docker container that they use to make their OS build? I can't work it out.

Basically some files are copied into our working tree from system directories, and they have permissions which are simply "wrong" in those system directories.

NWilson: I opened a ticket with GitHub, since it appears to be something strange on their end:
https://github.com/actions/runner-images/issues/11212

PR #632, opened 2024-12-18
Improve possessive repeat support

zherczeg:


NWilson (approved):
Again, I haven't precisely checked the code, but from a high-level it's all sane.

PR #633, opened 2024-12-18
Update references to maintainers in the README

NWilson:


README
```diff
@@ -964,7 +964,7 @@ The distribution should contain the files listed below.
   vms/pcre2.h_patch       )
   vms/stdint.h            )
 
-Philip Hazel
-Email local part: Philip.Hazel
-Email domain: gmail.com
-Last updated: 15 April 2024
+===========================
+Last Updated: 15 April 2024
+===========================
```
NWilson, line 969: I'm not trying to rewrite history to get rid of @PhilipHazel here. The motivation is just removing his email address from appearing in the PCRE2 tarball, since he's no longer the point of contact.

doc/pcre2.3
```diff
@@ -186,17 +186,20 @@ In the "man" and HTML formats, there is also a short page for each C library
...
-.SH AUTHOR
+.SH AUTHORS
 .rs
 .sp
-.nf
-Philip Hazel
-Retired from University Computing Service
-Cambridge, England.
-.fi
+The current maintainers of PCRE2 are Nicholas Wilson and Zoltan Herczeg.
```
NWilson, line 192: I would have used Zoltán's actual name (with á) but that causes difficulties inside manpages. It's totally possible on Linux, by using the `\[u00E1]` escape, but I don't know about how that would be handled on other systems. So, I'm keeping the manpages ASCII-only, out of caution.

PR #635, opened 2024-12-18
Add maint scripts for checking and updating library version & updated dates

NWilson:
The workflow shall be:
* When the release number is bumped, all references to that release number need to be bumped immediately. (For example, when the source code moves from 10.45 → 10.46, the man pages must do so as well.)
* When documentation is updated, there's no need to update the "last modified" dates by hand. We can sweep those all up during the release process. Or update them immediately - there's no harm in it; we simply aren't obliged to.


NWilson:
I'm probably the only person who'll ever run these scripts, so I'll go ahead and merge this soon.

PR #636, opened 2024-12-18
Update libtool version for 10.45

NWilson:
I believe this is the correct procedure, based on Philip's documentation.

The `libpcre2-posix` interface is completely unchanged, but the source code has been updated.

The `libpcre2-NN` interface has been extended in backwards-compatible ways, with new enum values and API functions, so callers compiled and linked against the old version may use the newer as a drop-in replacement, but callers compiled against the 10.45 headers will fail when used against the old version (if they require any newly-added functions).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -14,10 +14,10 @@ m4_define(pcre2_prerelease, [-DEV])
 m4_define(pcre2_date, [2024-06-09])
 
 # Libtool shared library interface versions (current:revision:age)
-m4_define(libpcre2_8_version,     [13:0:13])
-m4_define(libpcre2_16_version,    [13:0:13])
-m4_define(libpcre2_32_version,    [13:0:13])
-m4_define(libpcre2_posix_version, [3:5:0])
+m4_define(libpcre2_8_version,     [14:0:14])
+m4_define(libpcre2_16_version,    [14:0:14])
+m4_define(libpcre2_32_version,    [14:0:14])
+m4_define(libpcre2_posix_version, [3:6:0])
 
 # NOTE: The CMakeLists.txt file searches for the above variables in the first
 # 50 lines of this file. Please update that if the variables above are moved.
--- a/maint/manifest-cmakeinstall-linux
+++ b/maint/manifest-cmakeinstall-linux
@@ -13,20 +13,20 @@ drwxr-xr-x install-dir/lib/cmake/pcre2
 -rw-r--r-- install-dir/lib/cmake/pcre2/pcre2-config.cmake
 -rw-r--r-- install-dir/lib/libpcre2-16.a
 lrwxrwxrwx install-dir/lib/libpcre2-16.so -> libpcre2-16.so.0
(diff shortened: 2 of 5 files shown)


carenas (approved):
the commit should be talking about updating the soname version IMHO.

libtool version remines unhanged AFAIK.

NWilson:
@PhilipHazel I'd appreciate a confirmation that I'm following your rules for PCRE2 correctly.

PhilipHazel:
The are not my rules - I found them somewhere on the net ages ago, and just follow them blindly. Reading them through now I am wondering whether the "age" (3rd number) should become zero because "interfaces have been removed or changed", though exactly what "changed" means - in this just added options (unless I've missed something - entirely possible). This looks like note 2 after the rules, which does say to set the age to 0.

NWilson:
Thank you. I've found where they come from:
https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html

The libtool manual itself, so pretty authoritative.

I think the comments underneath are helpful to understand what the authors mean by an interface being "changed or removed".

> 1. Programs using the previous version may use the new version as drop-in replacement, and programs using the new version can also work with the previous one. In other words, no recompiling nor relinking is needed. In this case, bump revision only, don’t touch current nor age.
> 1. Programs using the previous version may use the new version as drop-in replacement, but programs using the new version may use APIs not present in the previous one. In other words, a program linking against the new version may fail with “unresolved symbols” if linking against the old version at runtime: set revision to 0, bump current and age.
> 1. Programs may need to be changed, recompiled, and relinked in order to use the new version. Bump current, set revision and age to 0.

It's to do with whether the new library can be use 100% safely by clients compiled and linked against the old library.

We can rule out situation №3, we're never intending to break the ABI in that way.

So we just have to decide whether each release is a "patch release" or "adds interfaces". Basically if pcre2.h has changed at all, then we're in situation №2; otherwise for a pure bugfix release it's №1.

Does that make sense?

carenas:
> Does that make sense?

there is an additional API change that doesn't necessarily change the ABI or adds functions to the API, and that we are actually having in this release.

errors (or their offsets) might change between versions and therefore an application that was compiled with a previous version and that would happily link with a newer version might behave differently when using that.

there is also a somehow related issue we have yet to address when the application might be linked to two different PCRE2 versions (ex: an application that uses a library that links itself to PCRE2) and that is related to [Debian's BUG 902060](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=902060)

NWilson:
Thanks Carlo for mentioning the issue of errors changing between releases. I think I wouldn't regard that as an ABI break, although it is certainly something to be aware of as we develop, to ensure that we don't change any return codes that callers might be depending on.

I have opened #640 for your comment about symbol versioning. Thank you for bringing that to my attention. I feel it would be much too risky to attempt a change like that for 10.45 however.

NWilson:
This PR updates the SOVERSION in exactly the same way as Philip did for the 10.44 release, so I'm reassured by that.

None of use exactly knows what the libtool version number does, so I'm happy to repeat whatever Philip has done previously.

PR #637, opened 2024-12-19
Small improvement to combination of substitution callout + overflow

NWilson:
I reckon that callers are assuming that when you use the PCRE2_SUBSTITUTE_OVERFLOW_LENGTH option, it will calculate the entire memory requirement in one go. Just two calls should be sufficient (rather than needing to loop with a gradually-increasing buffer size).

However, with a substitution callout this is not true. If you call once with PCRE2_SUBSTITUTE_OVERFLOW_LENGTH, the buffer length returned might still not be sufficient for the second call to succeed.

This is because the callout might not be called the first time, but the second time it will be called and can affect control flow, by requiring even more buffer to be used. This occurs even if the callout is completely stateless, idempotent and well-behaved.

This fix ensures that when we skip a callout (due to overflow), we still request enough buffer size for either option that the callout might return.

### DIFF
--- a/src/pcre2_substitute.c
+++ b/src/pcre2_substitute.c
@@ -1148,60 +1148,62 @@ do
     }   /* End of loop for scanning the replacement. */
 
   /* The replacement has been copied to the output, or its size has been
-  remembered. Do the callout if there is one and we have done an actual
-  replacement. */
+  remembered. Handle the callout if there is one. */
 
-  if (mcontext == NULL || mcontext->substitute_callout == NULL)
-    {}
-
-  else if (!overflowed)
+  if (mcontext != NULL && mcontext->substitute_callout != NULL)
     {
-    scb.subscount = subs;
-    scb.output_offsets[1] = buff_offset;
-    rc = mcontext->substitute_callout(&scb, mcontext->substitute_callout_data);
+    /* If we an actual (non-simulated) replacement, do the callout. */
 
-    /* A non-zero return means cancel this substitution. Instead, copy the
-    matched string fragment. */
-
-    if (rc != 0)
+    if (!overflowed)
       {
-      PCRE2_SIZE newlength = scb.output_offsets[1] - scb.output_offsets[0];
+      scb.subscount = subs;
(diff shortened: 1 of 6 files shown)


src/pcre2_substitute.c
```diff
@@ -1148,7 +1151,10 @@ do
   remembered. Do the callout if there is one and we have done an actual
   replacement. */
 
-  if (!overflowed && mcontext != NULL && mcontext->substitute_callout != NULL)
+  if (mcontext == NULL || mcontext->substitute_callout == NULL)
+    {}
```
zherczeg, line 1155: I don't really like these

`if ((mcontext != NULL || mcontext->substitute_callout != NULL) && !overflowed)`

zherczeg (approved):
LGTM

src/pcre2_substitute.c
zherczeg: I prefer nesting. Indentation makes easier to see the logical blocks.

NWilson: Great, done.

ltrzesniewski:
Beware, the "usual" callouts (`pcre2_set_callout`) may still return different results when called a second time.

I had to [store a queue of results](https://github.com/ltrzesniewski/pcre-net/blob/bacd34a02a0dad05df38bcdf5a59f42e517d4ddf/src/PCRE.NET.Native/pcrenet_substitute.c#L279-L280) for both callout types in PCRE.NET in case of buffer overflow.

NWilson:
Thanks @ltrzesniewski Lucas. That's very interesting! You have exactly the "while loop" I was thinking of, that some callers have probably not written.

If the callouts are stateful or not deterministic, then... all bets are off. You'll need a while loop. If the callouts are idempotent, then it should be possible to estimate the final buffer size in just one single pass.

ltrzesniewski:
Yes, in my case I have a binding library for C#, so all bets are off. I don't know what users will do so I need to make it look like each callout is invoked exactly once, and buffer the results.

PR #641, opened 2024-12-21
Improve non-greedy repeat support

zherczeg:
This should be the last patch of this work. All `[class]{a,b}+/?/nothing` now compiles `[class]` once. In the past a `[class]+` may compiled three times, which consumed a large space for some eclasses.


zherczeg:
I did some perf measurements and the new code has a very similar speed as the old. Code size is lower, that is the advantage.

NWilson (approved):
Cool, it's great to have them all matching, and all avoiding code bloat.

Issue #589, opened 2024-12-02
Simplify compile_class_not_nested (followup to ECLASS)


NWilson:
The function compile_class_not_nested is very big and complicated. I reckon we can help it out:

* Always have cranges. For the no-wide-char cases, we can just use a cranges object on the stack (to avoid malloc) and write into that. The wide-char cases would still need to malloc a cranges. But the advantage would be: all the code which follows can avoid having two separate paths.
* We should put the bitmap into the struct cranges. Then, the struct cranges would be a complete description of the characters within the class (that is - all the non-property characters in the class), which would make it easier to reason about. "The characters matched by the class are identified by _this_ bitmap, _these_ cranges/clists, plus the Unicode properties."
* We have code which is split out into separate places. For example, POSIX classes and escapes are handled twice - once when building the cranges, to add in the non-ASCII characters, and again after building the cranges, to add the ASCII characters to the bitmap.

I think we should do build the cranges from the METAs (ideally in one pass, although two passes would be acceptable); then one single pass over the cranges to build the output (without needing to back over the METAs - except possibly just to pick up Unicode property escapes which were completely ignored in the cranges).

Currently, the 16-bit library always, unconditionally, mallocs a cranges object even for super-simple character classes that just match ASCII characters! Using a small stack-alloced cranges as the starting point would be very nice, and we would only grow it to a heap-alloced one if the character class is too big to fit.

Items in the character class should handled in exactly one place: if the item is a character literal or POSIX class then it goes in the cranges, and shouldn't need to be handled elsewhere; if it's a Unicode escape then it's ignored in the cranges and is handled just once later on.

zherczeg:
The function simplified a lot compared to its original form. It could cache the entire compiled class though. Then no processing during the lengthptr != case, just a memory copy.

PR #644, opened 2024-12-23
Fix UCP-only handling of caseless character classes

### LINKED ISSUE #526: Bug in char class handling, in caseless mode, with UCP (caused by PR 474)

NWilson:
Fixes #526


NWilson:
The regression found by Alex was worse than he even reported.

Right now, on master, if you do `/[a]/i,ucp` then it will match `a` but not `A`!! The current code appears to be not doing any case-insensitivity whatsoever if you have UCP but not UTF. Ouch.

PhilipHazel:
`/[a\x{c1}]/i,ucp ` works on 10.44 (I guess you probably know that).

zherczeg:
https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_compile_class.c#L1065

Would it be better to run the range generator in ucp mode? Then ignore ranges > 256.

NWilson:
> Would it be better to run the range generator in ucp mode? Then ignore ranges > 256.

Hmm. I honestly don't know. The purpose of the cranges code feels it's all to do with supporting wide characters, as opposed to small/bitmapped ones. What you're suggesting basically means that the non-crange branches don't need to support Unicode at all.

I don't have a strong feeling either way? In any case, after 10.45, I was planning to just eliminate the non-crange code, and use a stack-allocated cranges always.

Alex's original PR (still open) did what you suggested.

NWilson:
Note that in Alex's PR, he found it was hard to prevent generating an XCLASS in certain situations, if you naively generate the cranges for UCP-without-UTF.

I didn't actually pursue it myself, to see how tricky it would be.

My biggest feeling was it was just a less risky change to add in the other-cased characters to the bitmap, than to change the baked-in assumptions in the code about when a cranges is/is-not generated.

The other-cased characters really do belong in the bitmap, so there can't be any harm or unexpected consequences to the change I made (I hope).

src/pcre2_compile_class.c
```diff
@@ -911,14 +913,45 @@ restriction is in force). Sometimes we can just extend the original range. */
...
   {
 #ifdef SUPPORT_UNICODE
-  if ((options & (PCRE2_UTF|PCRE2_UCP)) == 0)
+  /* UTF mode. This branch is taken if we don't support wide characters (e.g.
+  8-bit library, without UTF), but we do treat those characters as Unicode
+  (if UCP flag is set). In this case, we only need to expand the character class
+  set to include the case pairs which are in the 0-255 codepoint range. */
+  if ((options & (PCRE2_UTF|PCRE2_UCP)) != 0)
+    {
+      BOOL turkish_i = (xoptions & (PCRE2_EXTRA_TURKISH_CASING|PCRE2_EXTRA_CASELESS_RESTRICT)) ==
```
zherczeg, line 922: Don't this code duplicate logic here? This should happen only when 8 bit mode, without uftf, with ucp.

zherczeg:
The cranges can be quite big, I am not sure allocating it on the stack always works. Maybe for small cranges.

src/pcre2_compile_class.c
NWilson: Oh ouch. This is seriously-complicated code, with tight coupling.

You're right, with my change, we are doing caselessness twice, in some conditions, like 16-bit-without-UTF.

I've added one small extra change in a commit I've just pushed. I've also experimented with some additional asserts, to check that casing isn't being done twice.

---

I think that if we put a 256-bit map at the start of the cranges, then we can stack-allocate it for the no-wide-chars condition. It would also make it a little more efficient for UTF patterns which are only using ASCII characters (no need to heap-allocate those either).

zherczeg: I may misunderstand the code, but `(options & (PCRE2_UTF|PCRE2_UCP)) != 0` always redo the caseless computing, not just in non ucp mode.
What about checking the 8 bit with `#if` then use `if ((options & (PCRE2_UTF|PCRE2_UCP)) == PCRE2_UCP)` to limit

src/pcre2_compile_class.c
```diff
@@ -911,14 +913,45 @@ restriction is in force). Sometimes we can just extend the original range. */
...
+        PCRE2_EXTRA_TURKISH_CASING;
+      if (start < 128)
+        {
+        uint32_t lo_end = (classbits_end < 127 ? classbits_end : 127);
+        for (c = start; c <= lo_end; c++)
+          {
+          if (turkish_i && UCD_ANY_I(c)) continue;
```
zherczeg, line 929: As far as I remember, start / end should be limited to 0..255 range. How can turkish_i can affect this? You should never add char > 256 to the bitset, nor process a character > 256.

zherczeg: I realized this is not allowing i -> I. Ok, then we need this extra specialization. I am sorry.

src/pcre2_compile_class.c
NWilson: The code in the PR passes all the tests with this additional assertion, below.

However, I'm not sure I want to actually add that assertion in the code, since then everything becomes even more tightly coupled.

In any case, your invariant is currently true that `(options & (PCRE2_UTF|PCRE2_UCP)) == PCRE2_UCP`.

```diff
diff --git a/src/pcre2_compile_class.c b/src/pcre2_compile_class.c
index 329c9e89..cb73eefb 100644
--- a/src/pcre2_compile_class.c
+++ b/src/pcre2_compile_class.c
@@ -909,30 +909,31 @@ uint32_t classbits_end = (end <= 0xff ? end : 0xff);
 cases. In Unicode, there are 8-bit characters that have alternate cases that
 are greater than 255 and vice-versa (though these may be ignored if caseless
 restriction is in force). Sometimes we can just extend the original range. */

 if ((options & PCRE2_CASELESS) != 0)
   {
 #ifdef SUPPORT_UNICODE
   /* UTF mode. This branch is taken if we don't support wide characters (e.g.
   8-bit library, without UTF), but we do treat those characters as Unicode
   (if UCP flag is set). In this case, we only need to expand the character class
   set to include the case pairs which are in the 0-255 codepoint range. */
   if ((options & (PCRE2_UTF|PCRE2_UCP)) != 0)
     {
+      PCRE2_ASSERT((options & (PCRE2_UTF|PCRE2_UCP)) == PCRE2_UCP);
       BOOL turkish_i = (xoptions & (PCRE2_EXTRA_TURKISH_CASING|PCRE2_EXTRA_CASELESS_RESTRICT)) ==
         PCRE2_EXTRA_TURKISH_CASING;
       if (start < 128)
         {
         uint32_t lo_end = (classbits_end < 127 ? classbits_end : 127);
         for (c = start; c <= lo_end; c++)
           {
           if (turkish_i && UCD_ANY_I(c)) continue;
           SETBIT(classbits, cb->fcc[c]);
           }
         }
       if (classbits_end >= 128)
         {
         uint32_t hi_start = (start > 128 ? start : 128);
         for (c = hi_start; c <= classbits_end; c++)
           {
           uint32_t co = UCD_OTHERCASE(c);
```

src/pcre2_compile_class.c
NWilson: It is pretty weird, I have to admit. The existing tests and behaviour for Turkish I in 8-bit-with-UCP-no-UTF mode are very strange, because the user is explicitly saying "bytes 0x00..0xff are codepoints U+00...U+FF", but at the same time saying that `i/İ` is a case pair, even though İ can't be provided as input! For consistency though, I think it's right as it is.

src/pcre2_compile_class.c
```diff
@@ -958,21 +991,23 @@ case-independence.
...
+  xoptions      the extra options bits
   cb            contains pointers to tables etc.
   p             points to row of 32-bit values, terminated by NOTACHAR
 
 Returns:        cb->classbits is updated
 */
 
 static void
-add_list_to_class(uint32_t options, compile_block *cb, const uint32_t *p)
+add_list_to_class(uint32_t options, uint32_t xoptions, compile_block *cb,
```
zherczeg, line 1002: As far as I remember this code is only used for pre-defined bitsets, such as:
https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_compile_class.c#L1319

All of them removes caseless. Maybe we could just pass 0 as xoptions, and don't change the interface.
Minor optimization: maybe the caseless removal could be done in this function as well: `options &= ~PCRE2_CASELESS`.

zherczeg (approved):
LGTM. I really hope we can find a way to make this nicer in the future. Perhaps adding a separate "add_to_class_caseless", which is only available in 8 bit mode.

NWilson:
> LGTM. I really hope we can find a way to make this nicer in the future.

Agreed! I'll look into it in #589, in the new year

PR #645, opened 2024-12-23
Move L& conversion

zherczeg:
Minor code movement. The code runs once per pattern in its new place.


NWilson (approved):

zherczeg:
Adding a test case revealed a bug in jit.

NWilson:
Even better 😄

PR #646, opened 2024-12-26
Fix xclass compilation

zherczeg:
Another bugfix :(

### DIFF
--- a/src/pcre2_jit_test.c
+++ b/src/pcre2_jit_test.c
@@ -428,6 +428,7 @@ static struct regression_test_case regression_test_cases[] = {
 	{ MUP, 0, 0, 0 | F_NOMATCH, "[^[:print:]\\x{f6f6}]", "\xef\x9b\xb6" },
 	{ MUP, 0, 0, 0, "[[:xdigit:]\\x{6500}]#", "\xe6\x94\x80#" },
 	{ MUP, 0, 0, 0 | F_PROPERTY, "[\\pC\\PC]#", "A#" },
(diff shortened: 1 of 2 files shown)


src/pcre2_jit_test.c
```diff
@@ -428,6 +428,7 @@ static struct regression_test_case regression_test_cases[] = {
 	{ MUP, 0, 0, 0 | F_NOMATCH, "[^[:print:]\\x{f6f6}]", "\xef\x9b\xb6" },
 	{ MUP, 0, 0, 0, "[[:xdigit:]\\x{6500}]#", "\xe6\x94\x80#" },
 	{ MUP, 0, 0, 0 | F_PROPERTY, "[\\pC\\PC]#", "A#" },
+	{ MUP, 0, 0, 0 | F_PROPERTY, "[\\x80-\\xff\\x{800}\\x{802}\\x{804}\\p{Cc}]", "\xdf\xbf\xe0\xa0\x80" },
```
NWilson, line 431: It's great to have a test added in _jit_test.c. But is it not covered in any of the testdata tests, which are also run with the JIT?

I just wonder what we (or I) could do when I'm writing tests in future, so that these bugs are caught.

NWilson (approved):

src/pcre2_jit_test.c
zherczeg: There is no rule for this. I usually add something to jit tests, when the test is more of a jit correctness test (and I know the interpreter is correct) rather than a feature test.

For example, if I match a P pattern to `aaaaa` in pcre2test, and the result is `aaa`, we don't know which `a` are matching. In jit test, we compare offsets, although if both the interpreter and jit are wrong, we cannot detect that. So none of the methods are perfect.

PR #647, opened 2024-12-26
Pre-compute unicode category list for xclasses

zherczeg:
This patch moves a jit optimization to compiler optimization. The unicode categories (there are 30 of them) are combined into a 32 bit bitset, and stored in xclass, instead of a list of properties. This 32 bits, if present, follows the bitset for the first 256 characters. If the value of categories is 0, it is not stored in the xclass.

The patch is working, but the debug (`/B`) output is changed for 8-bit ucp case, because the bitset is stored in cranges, and the 8 bit ucp has no cranges.

When all 30 category bits are present, the xclass is converted to allany or nothing (negated case).

What do you think about this optimization?

### DIFF
--- a/src/pcre2_compile_class.c
+++ b/src/pcre2_compile_class.c
@@ -116,10 +116,11 @@ while (TRUE)
 
 #ifdef SUPPORT_UNICODE
 
-#define PARSE_CLASS_UTF               0x1
-#define PARSE_CLASS_CASELESS_UTF      0x2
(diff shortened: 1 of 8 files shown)


zherczeg:
Note: this patch can wait after the next release.

NWilson:
I like the idea. I'll have to think about it a bit, but it seems OK in principle.

Nice idea.

PR #649, opened 2024-12-27
Fix warning about signed/unsigned mismatch

ltrzesniewski:
Congrats on the RC1 🙂 

I've got the following warning when trying to compile it with MSVC though, which is a problem when you turn warning as errors on:

> Warning C4018 : '>': signed/unsigned mismatch

(I supposed the PR should target the master branch)

### DIFF
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -1767,7 +1767,7 @@ switch(*cc)
   if (max == 0)
     return (*cc == OP_CRRANGE) ? 2 : 1;
   max -= min;
-  if (max > (*cc == OP_CRRANGE ? 0 : 1))
+  if (max > (sljit_u32)(*cc == OP_CRRANGE ? 0 : 1))
     max = 2;
   return max;
 


NWilson:
Thank you very much! Our windows CI builds don't have all the warnings enabled, unfortunately, so that's how this got past testing. There's an existing issue in my personal task list to tighten up on them.

I'll let Zoltan review and merge, but the fix is fine to me. Adding "u" after the number constants would also work, if you wanted to avoid committing to a particular bit width. 

Merging to master is correct. 

Master is the new playground, open to all changes big and small, and I'll do the admin to cherry pick any fixes like this onto the final release branch when the time comes. (@zherczeg)

ltrzesniewski:
Oh, I should have mentioned I already had an issue like this in the past, see #111.

I used a cast because I did so in that issue, there already were a number of those in the file, and because `max` is a `sljit_u32` (so there's no bit width issue).

zherczeg (approved):
LGTM

Issue #651, opened 2024-12-28
Tighten up some compiler coverage in CI


NWilson:
After a recent regression in 10.45-RC1 (build fails with error-on-warn in MSVC), he's a list of some little improvements we could make: 

* Use much stricter windows warnings
* Use CMake as well on FreeBSD and Solaris
* Add 64 bit Solaris build and fix existing warnings
* Consider adding an intel ICC and/or Edison EDC build for even more testing diversity?
* Add a EBCDIC build & tests
* Add a big-endian platform? (e.g. Debian Bookworm S390 via qemu)

PR #655, opened 2025-01-07
Enable and fix additional build warnings

NWilson:
Part of #651

* Use much stricter windows warnings (`/W3` rather than `/W1`). This requires quite a few fixes for all the sloppy places where we do implicit assignment of 64-bit values to 32-bit storage.
* Use and test CMake build & install on FreeBSD and Solaris
* Add 64 bit Solaris build (`cc -m64`) and fix existing Solaris warnings
* Make compile flags used in CI consistent across platforms. Previously Mac & Linux were building with different warning flags.
* Add `--enable-Werror` to `configure.ac`. This means that you can build with `-Werror` in a clean way. Previously, you had to hackily override the CPPFLAGS when calling `make` since you can't pass `-Werror` as a CFLAG into `./configure` (it messes with compiler feature detection).


src/pcre2_jit_char_inc.h
```diff
@@ -1104,7 +1104,7 @@ if (ranges.range_count >= 6)
 
 depth = 0;
 first_item = 0;
-last_item = ranges.range_count - 2;
+last_item = (sljit_u32)(ranges.range_count - 2);
```
NWilson, line 1107: @zherczeg This PR is totally not urgent. But I won't merge any changes to your JIT code unless you confirm it's all OK.

src/pcre2_jit_neon_inc.h
```diff
@@ -82,6 +82,11 @@ POSSIBILITY OF SUCH DAMAGE.
 #  endif
 # endif
 
+#if defined(__GNUC__) || defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeclaration-after-statement"
```
NWilson, line 87: Fixing this file to comply with the policy of "declarations at the top" seemed too annoying, so I just disabled the warnings in this file.

We didn't notice about it previously, because this one is only used on Mac (or ARM Linux) and we previously didn't include `-Wdeclaration-after-statement` on those platforms in our CI.

src/pcre2grep.c
```diff
@@ -1263,8 +1263,8 @@ for (op = optionlist; op->one_char != 0; op++)
     n = 31 - printf("  -%c", op->one_char);
   else
     {
-    if (op->one_char > 0) sprintf(s, "-%c,", op->one_char);
-      else strcpy(s, "   ");
+    if (op->one_char > 0) snprintf(s, sizeof(s), "-%c,", op->one_char);
+      else snprintf(s, sizeof(s), "   ");
```
NWilson, line 1267: I have a fear of sprintf, strcpy and strcat. Perhaps we should ban them. All the current uses within the code seem harmless though - but they can go so badly wrong.

src/pcre2_jit_neon_inc.h
carenas: clang ALSO defines `__GNUC__`, so there is no need for the second check.

this file is expected to be removed once slits SIMD is also enabled for ARM, and indeed that would solve several known problems with the current implementation.

zherczeg: Still a lot of work ahead.

src/pcre2_jit_compile.c
```diff
@@ -2438,7 +2439,7 @@ SLJIT_ASSERT(stackpos == STACK(stacktop));
 typedef struct delayed_mem_copy_status {
   struct sljit_compiler *compiler;
   int store_bases[RECURSE_TMP_REG_COUNT];
-  int store_offsets[RECURSE_TMP_REG_COUNT];
+  sljit_sw store_offsets[RECURSE_TMP_REG_COUNT];
```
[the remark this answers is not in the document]
NWilson: OK. If I make it an s32, then I'll need to add casts when assigning (to tell the compiler we're sure the user isn't putting big values in there).

src/pcre2_jit_compile.c
```diff
@@ -3380,7 +3381,7 @@ OP2(SLJIT_SUB | SLJIT_SET_Z, COUNT_MATCH, 0, COUNT_MATCH, 0, SLJIT_IMM, 1);
 add_jump(compiler, &common->calllimit, JUMP(SLJIT_ZERO));
 }
 
-static SLJIT_INLINE void allocate_stack(compiler_common *common, int size)
+static SLJIT_INLINE void allocate_stack(compiler_common *common, sljit_sw size)
```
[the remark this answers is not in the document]
NWilson: Things like strlen,  or subtraction of pointers, return a size_t, which is then passed to `allocate_stack()`. So I assumed that it could indeed be passed a size_t input, and that the various operations below would be able to handle it. But if it's a stack allocation, I guess that's not correct anyway, and the caller does need to sure it's only passing in "small" values.

I'll change it back, and update the callers not to cast away any large values.

src/pcre2_jit_compile.c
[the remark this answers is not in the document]
NWilson: PS. I'm assuming that "sljit_sw" means "Signed Word", where "Word" = register width (rather than "word" = 32 bits).

maint/manifest-cmakeinstall-freebsd
```diff
@@ -0,0 +1,253 @@
+drwxr-xr-x install-dir
```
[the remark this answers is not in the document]
NWilson: They are manifests of the contents of the tarball, and all the files installed by `cmake --install` and `make install`. If files are added, removed, or installed with different permissions bits, then the tests will fail.

When we switched to a new Autoconf version in 2024 (between versions 10.44 and 10.45), or update CMake, I get really paranoid about files being lost or added accidentally, so I like having a fixed place where we assert that we are shipping the files we intend.

src/pcre2_jit_compile.c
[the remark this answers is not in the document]
zherczeg: Exactly.

It is confusing that there are two stacks in pcre2 jit. One is a machine stack, where globals (e.g. the current offset vector) is stored, and there is another stack, which is used for recursion (e.g. stores the previous values of globals). The machine stack is limited to 64Kbyte. The sljit supports 1Mbyte machine stack, but in practice 64K is more than enough.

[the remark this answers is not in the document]
carenas: > OK. If I make it an s32

wouldn't keep it as an int (and do the castings) be better in that case?, if someone has somewhere a DOS port of PCRE2?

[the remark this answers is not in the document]
zherczeg: JIT does not support 16 bit. sljit_s32 is good for this purpose

src/pcre2_jit_compile.c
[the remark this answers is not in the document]
NWilson: @zherczeg What's the maximum amount you can pass to allocate_stack? There's one caller which passes in the `callout_arg_size`, which is `2 * sizeof(void*) * top_bracket`. Because you can have 65536 capture groups, that's an entire 1MiB of data plus some extra few bytes.

Is the stack able to support enough for that?

src/pcre2_jit_compile.c
```diff
@@ -1838,7 +1838,8 @@ if (*next == OP_BRAZERO || *next == OP_BRAMINZERO)
 
       if (i == max)
         {
-        common->private_data_ptrs[max_end - common->start - LINK_SIZE] = next_end - max_end;
+        /* Patterns must fit into an int32 even for link-size=4. */
```
NWilson, line 1841: I think this comment I have just added is correct.

I was worried about whether there's any case where a pattern can by 3GiB big (ie, bigger than an S32 can hold, but smaller than the U32 maximum for a pattern with link-size=4).

I can see there's some nice code elsewhere:

> #define MAX_PATTERN_SIZE (1 << 30)   /* Keep it positive */

("Keep it positive" clearly means, "don't use `1<<32`" as the max size, but make sure it's comfortably smaller than that.)

This apparently ensures that you can't get in this situation, where you have a pattern whose compiled size is too large to fit in an int.

src/pcre2_jit_compile.c
```diff
@@ -2460,7 +2461,7 @@ status->compiler = common->compiler;
 }
 
 static void delayed_mem_copy_move(delayed_mem_copy_status *status, int load_base, sljit_sw load_offset,
-  int store_base, sljit_sw store_offset)
+  int store_base, int store_offset)
```
NWilson, line 2464: The `store_offset` arg is placed in the `status->store_offsets` array, which is already an `int` in the current code. So I think it's safe to make the argument an int (and hence avoid truncation when assigning to `status->store_offsets`).

That just moves the problem elsewhere though: now we need to worry about truncating to a 32-bit int in all the code which _calls_ `delayed_mem_copy_move`.

src/pcre2_jit_compile.c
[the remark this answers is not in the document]
zherczeg: It looks like I mixed this function up with `sljit_alloc_memory`. That is designed for small allocations.

This function should not be limited, although with very large numbers we could do a carry flag check.

src/pcre2_jit_compile.c
```diff
@@ -3184,7 +3185,7 @@ while (cc < ccend)
       SLJIT_ASSERT(private_srcw[i] != 0);
 
       if (!from_sp)
-        delayed_mem_copy_move(&status, base_reg, stackptr, SLJIT_SP, private_srcw[i]);
+        delayed_mem_copy_move(&status, base_reg, stackptr, SLJIT_SP, (int)private_srcw[i]); // XXX @zherczeg: cast OK?
```
NWilson, line 3188: @zherczeg I have added three casts here. I cannot verify whether these are OK. They are all truncation of an `sljit_sw` to an int.

It appears that all of the data placed inside kept_shared_srcw is within the range of an int; but private_srcw and shared_srcw are harder to check.

I'm hoping you "just know" already that these assignments are safe.

zherczeg: `sljit_s32` could be better than int. The casts are ok, this should copy between (machine) stack and (backtracking) stack :)

NWilson: Thank you very much for helping me out!

I have changed some uses of `int` to `sljit_s32`.

I think the changes to `pcre2_jit_compile` in this PR are now simply formatting - all the variables have the same type as before (since `int` and `sljit_s32` are actually the same), and we've just added explicit casts for things that were previously implicit casts. So the runtime code should be identical.

carenas: > I have changed some uses of `int` to `sljit_s32`.

historically (in the long obsolete PCRE1 codebase) all the interfaces used `int` and some might had not been updated since, so is usually safe to make this "conversion", but should also probably look for variables that were really meant to become PCRE2_SIZE, or in the case of JIT one of sljit_sw/sljit_uw.

NWilson: Yes Carlo, I _did_ look for cases where it should be PCRE2_SIZE or sljit_sw. That's what the discussion was about. I only changed values to sljit_s32 because Zoltan suggested it was better.

PhilipHazel: PCRE1 used int because, being started in 1997, it was written to the C90 standard, in which int32_t etc did not exist. (Just for the record.)

NWilson:
I've tagged Zoltan to re-confirm the changes I made in pcre2_jit_compile.c, before I merge.

zherczeg (approved):
LGTM

ucko:
I'm happy to confirm that this patch goes a long way towards addressing the warnings I tried to fix in #525.

However, I'd greatly appreciate it if you could please take another look at `pcre2_jit_neon_inc.h`; builds using Apple Clang with `-Wshorten-64-to-32` still warn heavily about two lines, particularly 104:
```
.../src/pcre2_jit_neon_inc.h:104:8: warning: implicit conversion loses integer precision: 'sljit_uw' (aka 'unsigned long') to 'unsigned int' [-Wshorten-64-to-32]
  104 | ic.x = chars;
      |      ~ ^~~~~
.../src/pcre2_jit_neon_inc.h:127:24: warning: implicit conversion loses integer precision: 'sljit_uw' (aka 'unsigned long') to 'sljit_u32' (aka 'unsigned int') [-Wshorten-64-to-32]
  127 | const sljit_u32 diff = IN_UCHARS(offs1 - offs2);
      |                 ~~~~   ^~~~~~~~~~~~~~~~~~~~~~~~
.../src/pcre2-GIT/src/pcre2_jit_compile.c:635:27: note: expanded from macro 'IN_UCHARS'
  635 | #define IN_UCHARS(x) ((x) * 4)
      |                       ~~~~^~~
```

Meanwhile, I also see four warnings from `pcre2grep.c` (one each about `length` truncation when calling `gzread` and `BZ2_bzread` and two about handling arguments that take 32-bit values) and one from `pcre2test.c` (about populating `errcode` via `strtol`), but each naturally shows up only once, and none should be a concern in practice.  I'd welcome cleanups here too (and can propose some if you're open to the possibility), but am willing to let these warnings slide. FTR, as of revision 81dced9, they cite lines 2598, 2604, 4278, and 4280 of `pcre2grep.c` and line 9823 of `pcre2test.c`.

Thanks!

NWilson:
@ucko I have added some additional warning suppressions and fixes for Clang in #689. Thank you very much for the detailed report of the error locations. I have tested and confirmed the warnings are all gone on GitHub's `macos-latest` environment, which uses Clang on an ARM macOS. I have no idea about iOS (which perhaps you use?) but I expect the fixes to work there too.

ucko:
Looks good; thanks again for all your work on this front!  At this point, the only warning I see in `master` (c421b939) is
```
src/pcre2test.c:4647:28: warning: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'int' [-Wshorten-64-to-32]
 4647 | int next_item_length = cb->next_item_length;
```
which is no big deal.  (FTR, I can't speak to iOS either, though.)

Sorry for the delayed reply -- long story short, I work for the US government (specifically, NIH) and was under (nominally) very broad communications restrictions when you tried to get back in touch.

NWilson:
Oh dear! I fully understand the situation at NIH, you have our deep sympathies.

Thank for reporting that one last compile warning. I don't know why it isn't appearing in the builds I'm running in CI, but I already spotted it in a local development build this week and I'll make a fix for it. Whew. Warnings are a pain, especially with so many platforms to support.

shaporanmp (approved):

shaporanmp (approved):

shaporanmp (approved):

PR #660, opened 2025-01-09
Add CI for MSYS2

### LINKED ISSUE #127: tests fail on MSYS2
sshot-17

No real idea what's happening.

NWilson:
Also fix a few minor portability issues.

- The unix "MSYS" environment has an usual (but sane) French locale, requiring a test update
- pcre2grep needs a fix to use the WIN32 when both WIN32 and sys/stat.h are available

Fixes #127

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -291,6 +291,49 @@ jobs:
       - name: Test
         run: cd build && ctest -j3 --output-on-failure
 
+  msys2:
+    # Tests with: MSYS2 unix-on-Windows environment
+    name: MSYS2
+    runs-on: windows-latest
+    if: github.event_name != 'pull_request'
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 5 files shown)


NWilson:
The issue is to do with the way Windows handles wildcard expansion. When you run `command *.txt` in CMD or Batch scripts, the glob is not expanded by the shell as it is on Unix: applications are expected to do this themselves (using the same FindNextFile Win32 function that is used for directory traversal).

`pcre2grep` does already have code to do this on Windows - however, it uses `#ifdef` to check first for the `sys/stat.h` header, and _then_ checks for whether `WIN32` is defined. On MSYS, this resulted in the Unix codepath being taken, since Unix headers are available in addition to Win32 headers.

The fix is a one-liner to ensure that the Win32 implementation is used (for the UCRT64 and other MinGW-x64 environments). The "MSYS" environment of MSYS2 (which is cygwin-like, and produces applications that are linked to the cygwin runtime) still uses the Cygwin-emulated Unix environment.

carenas (approved):

PR #663, opened 2025-01-11
Add multiarch build jobs

NWilson:
* S390x is important, because it's basically the only supported big-endian
  architecture I can find anywhere. I used to work on SPARC and PPC-be systems
  a long time ago, but even Debian has dropped those architectures now, so
  it's nice that there's *least one* arch remaining to shake out endian
  assumptions.
* PPC64LE is Big-iron POWER only (this is not the PowerPC arch used in old Apple Macs)
* ARMv7 is a 32-bit Linux build. i386 is mostly gone now, so ARMv7 is all that's left.
* AARCH64 is the only really widely-deployed non-x86 archicture, at least that's likely
  to be running PCRE2.


zherczeg:
Nice work! JIT changes are ok. Could RISCV be added as well?

NWilson:
OK, I added RiscV.

@zherczeg There's a build warning on S390x in the sljit code:

> https://github.com/PCRE2Project/pcre2/actions/runs/12724901045/job/35471575046#step:4:759

So, I've disabled `-Werror` on the multiarch build for now.

Adding the S390x build _was_ a good idea, it caught one recent small regression on big-endian architectures in pcre2test.c.

Everything else seems to be passing, which is great.

glaubitz:
FWIW, while Debian does not officially support big-endian PowerPC and SPARC anymore, it's still supported in Debian Ports.

There are developer machines running big-endian PowerPC and SPARC ports of Debian available in the [gcc.gnu.org/wiki/CompileFarm}(https://gcc.gnu.org/wiki/CompileFarm).

So, please don't block building pcre2 on these targets.

NWilson:
We haven't blocked or dropped support for anything, as far as I intended.

We've _added_ architectures here to our test matrix. However, because Debian no longer supports PowerPC (old arch, pre PPC64el) and SPARC, I have not added nightly testing for those. We never used to have nightly testing though.

I'm not sure how easily we'll be able to fix any regressions. I think for these "long tail" architectures I'm basically relying on users building and running the test suite, and reporting back if there are any problems. That's similar to my approach for OSes that I don't have easy access to (like AIX or HP-UX).

glaubitz:
Releases are regularly built and tested (by running the testsuite) in Debian unstable on large number of architectures, see: https://buildd.debian.org/status/package.php?p=pcre2&suite=sid

NWilson:
Terrific. Let's hope we don't break any of them. I aim not to, but it's hard when you don't have access to a test system.

PR #667, opened 2025-01-12
Update release number to 10.46-DEV #noupdate

NWilson:


zherczeg:
Do you have any plans / ideas for 10.46? I am thinking about improving the simd support, and there is one feature in my mind: recursion with escapable capture brackets. Normally recursion keep the capture brackets are private, which is great (kind of local variables), but it would be good if they could return with some of them (user selected). Syntax would look like this:

`(?name_or_id:coma_separated_list_of_capture_indicies)`. 

However, it looks like it is a bit complicated for supporting in the interpreter.

NWilson:
I don't have any firm plans.

I intend to tackle some of the backlog a bit - including some of Philip's ideas in `maint/README`, plus existing GitHub tickets.

I've tentatively selected about half of the GitHub tickets that look more interesting (to me) than the others:

https://github.com/PCRE2Project/pcre2/milestone/3

Many of those items are just to do with testing; test coverage; CI improvements; documentation...

However, there are a few that users would notice and benefit from - I moved them down to the bottom of the list, just to indicate that I'll do them last, if I have time. They're pretty minor though.

* Bug for patterns like `(x|(y)){N}\2`, where the backref should be unset if the last repetition didn't capture it (PCRE2 leaves the backref set to the last capture, even if the last repetition was an 'x').
* Support for `\p{Block=...}` should be easy, it's just a missing feature compared to Perl and other engines
* Support for `(x|y){5000}` which currently blows up the bytecode
* Sort out the CMake issues reported by the maintainers of the vcpkg port

NWilson:
I do have to pause my work for the rest of January though. I have things to do around the house (so I'm told!!) and I think it would be appreciated by my family if I take a little PCRE2 break.

PR #668, opened 2025-01-12
macOS related patches for 10.45

carenas:
A couple of "fixes" that came handy while building the snapshot in macOS (using an ARM CPU)

First one does just a aesthetic cleanup but has no side effects and is therefore likely safe.

Second one provides a mechanism for using gcc instead of clang and therefore also provides some minimal "fixes" to the codebase for building with an old version of it with all the extra warnings enabled.


zherczeg (approved):
LGTM

NWilson (approved):
It appears Philip added several AM_CONDITIONALs without actually making use of them. They are harmless if not used, but I'm happy for them to be used up.

As far as I'm aware, they don't affect the commandline options accepted by `./configure`; instead they define which variables are substituted inside `Makefile.in`. So if an AM_CONDITIONAL isn't referenced in Makefile.in, then it has no effect.

NWilson:
I apologize for the build failure in "Check autogenerated file freshness". You may ignore that. It seems I broke something by mistake over the weekend. I will fix it.

maint/ManyConfigTests
```diff
@@ -110,11 +110,12 @@ tmp=/tmp/pcre2testing
...
 CFLAGS="-g"
 OFLAGS="-O0"
+CC="${CC:=cc}"
 ISGCC=0
 
 # If the compiler is gcc, add a lot of warning switches.
 
-cc --version >/tmp/pcre2ccversion 2>/dev/null
+$CC --version >/tmp/pcre2ccversion 2>/dev/null
 if [ $? -eq 0 ] && grep GCC /tmp/pcre2ccversion >/dev/null; then
```
NWilson, line 119: Very curiously, on an Ubuntu machine, `cc --version` does not match "GCC", nor does `gcc --version` (it says "gcc" lowercase), nor does `clang --version`.

Hmm. I should check it's all OK in the CI!

NWilson: I found that the CI is not using all the GCC-specific warnings in the ManyConfigTests job! Oops. As you'd expect, the are some warnings that have crept in. I've made a separate PR for it: #670

PhilipHazel:
> t appears Philip added several AM_CONDITIONALs without actually making use of them. They are harmless if not used, but I'm happy for them to be used up.

I regret that I cannot remember the details of this, only that the building apparatus has been heavily hacked about over the decades. I am not surprised to learn that there is detritus in there.

PR #671, opened 2025-01-13
maint: avoid clang masquerading as gcc in ManyConfigTests

carenas:
__GNUC__ is defined b any compiler that clims compliance to GNU C but that doesn't include the cmdline interface, so avoid passing GCC specific warning flags to clang.


maint/ManyConfigTests
```diff
@@ -115,7 +115,7 @@ ISGCC=0
 
 # If the compiler is GCC or Clang, add a lot of warning switches.
 
-CC_VER_OUTPUT=`printf '#ifdef __GNUC__\nGCC=yes\n#endif\n' | $CC -E -`
+CC_VER_OUTPUT=`printf '#if defined(__GNUC__) && !defined(__clang__)\nGCC=yes\n#endif\n' | $CC -E -`
```
NWilson, line 118: It doesn't work? That's annoying, sorry.

We should fix the comment above to remove "or Clang".

carenas: It might be broken also with Intel's (at least the legacy) compiler as well then.

PR #672, opened 2025-01-21
Update README.md

NWilson:


README.md
```diff
@@ -1,56 +1,261 @@
...
+
+The default matching engine uses a depth-first tree search with backtracking, which is highly feature-rich but has worst-case exponential time (PCRE2 allows aborting the match if a time limit is exceeded, expressed as a maximum number of steps in the tree search). The second matching engine uses a JIT for greatly improved performance, compiling the regular expression to a block of equivalent native machine code.
+
+PCRE2 has a third matching engine, using a DFA engine which is generally slower, but has worst-case polynomial matching time and is able to find the POSIX-style "leftmost-longest" match.
+
+There are accompanying utility functions for converting glob patterns and POSIX BRE/ERE patterns to PCRE2 regular expressions; and also for performing high-level regular expression operations such as search-and-replace with a powerful replacement string syntax.
+
+As well as the PCRE2 API, the library also offers a POSIX-compatible `<regex.h>` header and `regexec()` function. However, this does not provide the ability to pass PCRE2 flags, so we recommend users consume the PCRE2 API if possible.
```
carenas, line 214: pcre2posix.h is actually no longer "POSIX" compatible, because it only supports "int" sized offsets

NWilson: Well yes, but even glibc has the same problem in their regex.h (_REGEX_LARGE_OFFSETS is not POSIX compliant). The POSIX guys just randomly changed it...

Whenever they fix it so that snprintf returns a size_t, we can worry about regoff_t. There's a vast number of standard C and POSIX C interfaces that incorrectly specify int, and the regex functions are the least of everyone's worries.

We should just ignore the problem, since we didn't start it.

PR #673, opened 2025-01-24
pcre2grep.1 - fix warning about undefined macro 0

MatthewVernon:
Debian's "lintian" picked this up - line 950 in the man page starts with a `'` which is how you start a roff request. You can reproduce the warning thus:

```
LC_ALL=C.UTF-8 MANROFFSEQ='' MANWIDTH=80 \
man --warnings -E UTF-8 -l -Tutf8 -Z doc/pcre2grep.1 >/dev/null
```

The fix is to add a zero-width space (`\&`) to the start of the relevant line (indeed `groff_man(7)` suggests exactly this use for `\&`).

This is obviously not a critical issue, but it'd be nice if it made it into 10.45 please :)

### DIFF
--- a/doc/pcre2grep.1
+++ b/doc/pcre2grep.1
@@ -947,7 +947,7 @@ Any substring (including the executable name) may contain escape sequences
 started by a dollar character. These are the same as for the \fB--output\fP
 (\fB-O\fP) option documented above, except that $0 or $& cannot insert the
 matched string because the match is still in progress. Instead, the character
-'0' is inserted. If you need a literal dollar or pipe character in any
+\&'0' is inserted. If you need a literal dollar or pipe character in any
 substring, use $$ or $| respectively. Here is an example:
 .sp
   echo -e "abcde\en12345" | pcre2grep \e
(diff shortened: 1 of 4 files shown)


NWilson:
Arrgh, roff and groff madness. Thank you Matthew.

Unfortunately, we have some Perl scripts which we use to process the man pages (which are the "source of truth") and the Perl scripts are strict at enforcing a known subset of the nroff dialect.

* I could add support for `\&` to our scripts.
* We could just change the line-wrapping so that `'0'` isn't at the start of a line!
* I'll also add something to our CI or validation scripts to run `man --warnings` over our man pages, to verify that this doesn't happen again.

We can definitely get this into 10.45.

MatthewVernon:
I've pushed a commit which causes CheckMan to be happy with `\&`, which seemed the least-painful approach?

We could change the line-wrapping, but I worry that future-us will come along and change it back again in the future...

NWilson:
Hmm, I expect the `do_line` function in `maint/132html` will need updating too. Otherwise, the `\&` won't be stripped out in the HTML conversion.

I'll do that later this morning.

MatthewVernon:
I think I've done that one too (I ran it locally, and the output _looked_ plausible)

NWilson (approved):

NWilson:
Great, thank you again Matthew!

The check I added will prevent this recurring in future.

PR #689, opened 2025-02-06
Fix some Clang warnings with `-Wshorten-64-to-32`

NWilson:
This addresses Aaron's (ucko's) complaint from #655.

### DIFF
--- a/src/pcre2_jit_neon_inc.h
+++ b/src/pcre2_jit_neon_inc.h
@@ -101,7 +101,7 @@ int_char ic;
 SLJIT_UNUSED_ARG(offs1);
 SLJIT_UNUSED_ARG(offs2);
 
-ic.x = chars;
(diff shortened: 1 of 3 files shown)


src/pcre2_jit_neon_inc.h
```diff
@@ -124,7 +124,7 @@ vect_t vmask = VDUPQ(mask);
 compare_type compare1_type = compare_match1;
 compare_type compare2_type = compare_match1;
 vect_t cmp1a, cmp1b, cmp2a, cmp2b;
-const sljit_u32 diff = IN_UCHARS(offs1 - offs2);
+const sljit_uw diff = IN_UCHARS(offs1 - offs2);
```
NWilson, line 127: @zherczeg I just wanted to point out this small change in the NEON code. I think it's correct.

zherczeg: Likely. diff is a value between 0 and 31

PR #690, opened 2025-02-06
Fix typo

ltrzesniewski:
I noticed this while reading the docs.


NWilson:
Thank you Lucas! (The .txt and .html files will be regenerated automatically, but it's fine to do that yourself too.)

PR #693, opened 2025-02-11
Add tests to check shared object contents

NWilson:
We want to ensure that the symbols included in shared object files (.so, .dylib, .dll) are as expected.

This verifies that our Autoconf and CMake builds are using the `-fvisibility` flags and attributes correctly.

As it turns out, the CMake build on Solaris is not working, and was exposing internal symbols due to a CMake issue.


CMakeLists.txt
```diff
@@ -119,6 +119,17 @@ set(CMAKE_C_STANDARD_REQUIRED TRUE)
...
 
+# Solaris-specific fix for "CMAKE_C_VISIBILITY_PRESET": this feature was only
+# recently added to CMake for the `cc` compiler (Oracle Developer Studio). The
+# CMake version from OpenCSW and Oracle's package repository is too old and
+# requires this fix.
+if (CMAKE_SYSTEM_NAME STREQUAL "SunOS" AND
+    CMAKE_VERSION VERSION_LESS 3.31 AND
+    CMAKE_C_COMPILER_ID STREQUAL "SunPro" AND
+    CMAKE_C_COMPILER_VERSION VERSION_GREATER_EQUAL 5.15)
+  set(CMAKE_C_COMPILE_OPTIONS_VISIBILITY "-fvisibility=")
```
NWilson, line 130: @carenas I think you know about this fix! We need to ship it in our CMakeLists, _if_ we want to support the version of CMake that's shipped by Oracle and their latest Oracle Developer Studio.

PR #695, opened 2025-02-11
Build fixes for z/OS

NWilson:


NWilson:
I received some help from z/OS users, who were trying out the build & tests on z/OS for me. IBM said they'd grant me access to a test system, but that was a while ago, so I'll see how that works out.

I can't really merge this until I hear back that it works, or am able to verify it myself.

The build fixes are good - but extra small changes might be needed in RunTest in order to get the tests all passing.

zherczeg (approved):
Nice! LGTM

NWilson:
I was granted access to a test system. I didn't spend that long on this, in the end, and it turned out to be not much extra work to actually support z/OS.

I don't feel strongly about it. I am glad though that I now know for certain that the EBCDIC code works. I didn't feel comfortable shipping it as untested and bitrotted code.

If we do ever drop z/OS support, then I would purge the EBCDIC code from the source tree, rather than leave it as dubious abandoned ifdefs.

zherczeg:
If you are willing to maintain the port, I have no objection to it.

PR #698, opened 2025-02-14
Optimize capture name compilation

zherczeg:
The new code only do the searches once.

Also rename META_SCS_NAME and META_SCS_NUMBER to META_CAPTURE_NAME and META_CAPTURE_NUMBER respectively, since these are not depend on scan substring.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -219,8 +219,8 @@ static unsigned char meta_extra_lengths[] = {
   3,             /* META_COND_VERSION */
   SIZEOFFSET,    /* META_OFFSET */
   0,             /* META_SCS */
-  1,             /* META_SCS_NAME */
-  1,             /* META_SCS_NUMBER */
+  1,             /* META_CAPTURE_NAME */
+  1,             /* META_CAPTURE_NUMBER */
   0,             /* META_DOLLAR */
   0,             /* META_DOT */
   0,             /* META_ESCAPE - one more for ESC_P and ESC_p */
@@ -1091,12 +1091,12 @@ for (;;)
     fprintf(stderr, "META (*scan_substring:");
     break;
 
-    case META_SCS_NAME:
-    fprintf(stderr, "META_SCS_NAME length=%d relative_offset=%d", *pptr++, (int)meta_arg);
+    case META_CAPTURE_NAME:
+    fprintf(stderr, "META_CAPTURE_NAME length=%d relative_offset=%d", *pptr++, (int)meta_arg);
     break;
 
-    case META_SCS_NUMBER:
-    fprintf(stderr, "META_SCS_NUMBER %d relative_offset=%d", *pptr++, (int)meta_arg);
+    case META_CAPTURE_NUMBER:
+    fprintf(stderr, "META_CAPTURE_NUMBER %d relative_offset=%d", *pptr++, (int)meta_arg);
     break;
 
     case META_MARK:
@@ -4693,7 +4693,7 @@ while (ptr < ptrend)
                 errorcode = ERR15;
                 goto FAILED;
                 }
-              meta = META_SCS_NUMBER;
+              meta = META_CAPTURE_NUMBER;
               namelen = (uint32_t)i;
               }
             else if (errorcode != 0) goto FAILED;   /* Number too big */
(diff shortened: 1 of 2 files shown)


src/pcre2_compile.c
```diff
@@ -6520,16 +6520,18 @@ for (;; pptr++)
     case META_COND_RNUMBER:   /* (?(Rdigits) */
     case META_COND_NAME:      /* (?(name) or (?'name') or ?(<name>) */
     case META_COND_RNAME:     /* (?(R&name) - test for recursion */
-    case META_SCS_NAME:       /* Name of scan substring */
+    case META_CAPTURE_NAME:   /* Generic capture name */
     bravalue = OP_COND;
+
+    if (lengthptr != NULL)
```
zherczeg, line 6526: We have two code paths now, one for length computation and one for compilation.

NWilson: Nice! That seems good.

src/pcre2_compile.c
```diff
@@ -6520,16 +6520,18 @@ for (;; pptr++)
...
-      int count, index;
-      unsigned int i;
+      uint32_t i;
       PCRE2_SPTR name;
       named_group *ng = cb->named_groups;
+      uint32_t *start_pptr = pptr;
       uint32_t length = *(++pptr);
 
-      if (meta == META_SCS_NAME)
+      if (meta == META_CAPTURE_NAME)
```
NWilson, line 6534: You renamed META_SCS_NAME to META_CAPTURE_NAME, which is fine. But as far as I can tell, it's still being used only by scan_substring? Just checking I haven't missed anything. I can't see any new places where it's generated. In future of course we could re-use now that it has a more generic name.

NWilson (approved):
I think it looks good. The code is complex, but I think I follow it all.

src/pcre2_compile.c
```diff
@@ -6578,61 +6580,118 @@ for (;; pptr++)
...
-        PUT2(code, 2+LINK_SIZE, groupnumber);
+        PCRE2_ASSERT(start_pptr[0] == META_COND_RNUMBER);
+        start_pptr[1] = groupnumber;
         skipunits = 1+IMM2_SIZE;
         goto GROUP_PROCESS_NOTE_EMPTY;
         }
-      else if (!ng->isdup)
+
+      /* For compatibility reasons, Rdigits is converted to a name. */
+      if (meta == META_COND_RNUMBER) meta = META_COND_NAME;
```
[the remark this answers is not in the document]
zherczeg: There is a comment before the whole block which describes this. But your suggestion is better.

src/pcre2_compile.c
```diff
@@ -6578,61 +6580,118 @@ for (;; pptr++)
...
+      if (!ng->isdup)
         {
-        /* Otherwise found a duplicated name */
+        /* Otherwise found a duplicated name. Since it is a global,
+        it is enough to update it in the pre-processing phase. */
         if (ng->number > cb->top_backref) cb->top_backref = ng->number;
 
-        if (meta == META_SCS_NAME)
+        start_pptr[0] = meta;
+        start_pptr[1] = ng->number;
```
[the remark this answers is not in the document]
zherczeg: Yes.

src/pcre2_compile.c
zherczeg: Currently only scs uses it. I have plans for using it for other cases in the future.

src/pcre2_compile.c
```diff
@@ -6578,61 +6580,118 @@ for (;; pptr++)
...
-        cb->assert_depth += 1;
-        goto GROUP_PROCESS;
+        skipunits = 1 + IMM2_SIZE;
+        goto GROUP_PROCESS_NOTE_EMPTY;
         }
 
       /* We have a duplicated name. In the compile pass we have to search the
       main table in order to get the index and count values. */
 
+      PCRE2_ASSERT(i == (uint32_t)(ng - cb->named_groups));
```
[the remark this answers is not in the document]
zherczeg: It is for extra safety, if we might accidentally reuse i for something (this is a very generic tmp variable). Asserts have no cost in release code.

PR #700, opened 2025-02-16
Improve named group handling

zherczeg:
Add a simple hash code for group names to improve search speed.
Ignore duplicates when group names are searched.
Improve finding of duplicates (they have the same name pointer).
Improve creating name table (duplicates are handled in one step).
Create a new file for name management.


zherczeg:
I don't plan to add more to this patch

NWilson:
Nice! I haven't inspected the details yet, but it looks useful.

The hash code is just used to speed up the search, by avoiding strcmp calls, is that correct? That seems a bit of a marginal optimisation, but I guess it could help. More useful might be to add a binary search helper function, since the table is sorted.

I'll look at it in more detail in a bit.

zherczeg:
The named_group table is not sorted, and cannot be because it follows declaration order. The final table is alphabetical, binary search could be used there.

src/pcre2_intmodedep.h
```diff
@@ -726,13 +726,15 @@ typedef struct branch_chain {
...
+compiling. When a duplicate name is stored in the list, its name is set to
+the name of the first entry with the same name, and its length is set to 0. */
 
 typedef struct named_group {
   PCRE2_SPTR   name;          /* Points to the name in the pattern */
   uint32_t     number;        /* Group number */
   uint16_t     length;        /* Length of the name */
-  uint16_t     isdup;         /* TRUE if a duplicate */
+  uint16_t     hash_dup;      /* A concatenation of a 15 bit hash code and
+                                 a singe bit which represents duplication */
```
NWilson, line 737: Using a bitfield would be a bit nicer, although I'm not sure we use that convention within PCRE2 yet:

```c
unsigned is_dup: 1;
unsigned hash: 15;
```

Does it really need to be compact anyway? It could be a 32-bit hash and a uint8_t boolean.

src/pcre2_compile_cgroup.c
```diff
@@ -0,0 +1,262 @@
...
+PRIV(compile_get_hash_from_name)(PCRE2_SPTR name, uint32_t length)
+{
+uint16_t hash;
+
+PCRE2_ASSERT(length > 0);
+
+hash = (uint16_t)((name[0] & 0x7f) | ((name[length - 1] & 0xff) << 7));
+PCRE2_ASSERT(hash <= NAMED_GROUP_HASH_MASK);
+return hash;
+}
```
NWilson, line 68: Ah, I see the hash is extremely simple. It's just a way of checking the first & last characters match, before calling into strncmp?

I... would be slightly reluctant to add complexity for a minor optimisation, unless you've actually benchmarked it. Do we have any evidence that adding an extra function call here is even going to save any time at all?

Sorry to be negative, I just want to check is the benefit is real, or just a guess.

src/pcre2_intmodedep.h
zherczeg: I don't like bitfields. I also don't want to increase the structure size.

src/pcre2_compile_cgroup.c
```diff
@@ -0,0 +1,262 @@
...
+  duplicate or later name, carry on. We do this for duplicates so that in the
+  simple case (when ?(| is not used) they are in order of their numbers. In all
+  cases they are in the order in which they appear in the pattern. */
+
+  if (crc < 0)
+    {
+    (void)memmove(slot + cb->name_entry_size * duplicate_count, slot,
+      CU2BYTES((tablecount - i) * cb->name_entry_size));
+    break;
+    }
```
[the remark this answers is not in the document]
zherczeg: Not all sorts are good. The second argument (group index) needs to maintain declaration order. Sort algorithms often moves things around and breaks this rule.

src/pcre2_compile_cgroup.c
zherczeg: Pattern generated with python:
```
n = 1000

print("/", end='');
for i in range(n):
  print("(?<a%d>)" % (i), end='')
for i in range(n):
  print("\\k<a%d>" % (i), end='')
print("/");
```

`pcre2test` compiled with -O3 and with  `-t 1000`

Old: Compile time 10116.0110 microseconds
New: Compile time 3931.3390 microseconds

zherczeg: Btw I think the function is inlined. It could be a macro as well (except the assert part).

NWilson: You're a legend Zoltan, thank you for checking. Ok, let's do this optimisation!

src/pcre2_intmodedep.h
NWilson: Ok, cool. A manual bitfield is fine then

src/pcre2_compile_cgroup.c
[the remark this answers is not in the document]
NWilson: True, it would need to be a stable sort. Someone can fix this later, it's pre-existing

zherczeg:
Thank you. There are many things to do, but they can be done later. No need to hurry.

PR #701, opened 2025-02-18
Fix doc. -47 corresponds to PCRE2_ERROR_MATCHLIMIT not PCRE2_ERROR_NOMEMORY

MegaManSec:
Title should be self-explanatory.


NWilson:
Many thanks! How on earth did you spot that?!?

I have updated the PR to make the same change to the other copies of the documentation. The "primary file" is `pcre2test.1`, and the others are updated automatically from there.

MegaManSec:
Haha, I spotted it because I googled error number -47 [during some research](https://joshua.hu/nginx-directives-regex-redos-denial-of-service-vulnerable) and that page was the first one I could find. It made no sense to me that it was a no-memory error so had to go digging.

PR #705, opened 2025-02-20
Implement recurse which keeps user defined captures after a match

zherczeg:
This patch proposes a new feature: a recurse which keeps the user selected captures after a match. Recurse is a great feature, which keeps its local changes internally, but this feature is also its biggest limitation.

Syntax: `(?id:arglist)`, which is an extension of the normal `(?id)` recursion. I have no plans to extend `\g<id>` recursion with argument list. The captures defined by arglist keeps their values after recurse matches.

This is a rather complex feature, probably will require several patches. However, PCRE2 is just released, so there is enough time before the next release to do everything. I believe this is a great extension to recursions.


zherczeg:
I think this patch is large enough for one patch. The core parts of the feature is there, and there are tests for deep recursions, repeats, reusing captures, control verbs. I am sure there are missing parts (bugs), but hopefully there are no major issues with the concept.

The syntax uses a colon for starting the arguments. There are other options though:
`(?id:arg1,arg2,...)` - current
`(?id,arg1,arg2,...)` - looks like a single list for me
`(?id|arg1,arg2,...)` - the | is usually used for alternations
`(?id?arg1,arg2,...)` - maybe, but looks somewhat strange for me
`(?id(arg1,arg2,...))` - feels like a function call, but the double `))` at the end is not the best. I like this though.

There is no documentation. Could you help me with the documentation?

NWilson:
Yes! I can certainly write some documentation. I won't get to this PR until next week, since it's a holiday this week and I'm only doing minor little things in some spare minutes.

The colon is a good choice, but I like the function call too. I wonder if there's any benefit to coordinating new syntax with other engines... perhaps.

zherczeg:
Patch has been updated. I have changed the syntax to `(?id:(id,...))` so all capture list has the same syntax (recommended by perl contributors). There is a lot of things we could do, but this code is enough for one patch.

NWilson:
It's looking great! I haven't written the documentation yet (my apologies). I was working on the z/OS patch in the one or two spare hours I had, because the z/OS guys were messaging me with help and I wanted to take advantage of that while I could.

I really will help out on this PR as my next task.

zherczeg:
Thank you. As a new feature, this patch only contains the core parts (proves that the concept is sound). I have also plans to do future updates (adding more tests for example).

zherczeg:
(Btw we might even change the syntax before release based on the discussions with perl developers)

zherczeg:
Another note: dfa does not support capturing brackets. This feature depends on captures, and currently we return an error. This is the same as for scan substring. We can also ignore the arguments instead of error.

```
./pcre2test -dfa
PCRE2 version 10.46-DEV 2024-06-09 (8-bit)
  re> /(a)(b)/
data> ab
 0: ab
```

NWilson:
Great! I'm happy to continue in future PRs.

An error for DFA sounds good. I'd rather be strict, than silently ignore explicit syntax from the user.

PR #710, opened 2025-02-27
Remove duplicated scan substring captures

zherczeg:
This patch optimizes the argument list of scan substring by removing duplications. Multinames are only removed if all corresponding captures are removed as well.

Not a simple patch unfortunately.

It would not be bad, if equality here could be checked somehow:
https://github.com/PCRE2Project/pcre2/blob/master/src/pcre2_compile.c#L10854

The greater case is only needed for a few patterns. Don't know how to do this. Maybe tests should contain a `#` option to enforce equality, and that could be disabled for specific patterns.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -6495,8 +6495,63 @@ for (;; pptr++)
     req_caseopt = ((options & PCRE2_CASELESS) != 0)? REQ_CASELESS : 0;
     break;
 
+    /* ===================================================================*/
+    /* Handle scan substring. Scan substring assertion starts with META_SCS,
+    which recursively calls compile_branch. The first opcode processed by
+    this recursive call is always META_OFFSET. */
+
     case META_OFFSET:
-    GETPLUSOFFSET(offset, pptr);
+    if (lengthptr != NULL)
+      {
+      pptr = PRIV(compile_parse_scan_substr_args)(pptr, errorcodeptr, cb, lengthptr);
+      if (pptr == NULL)
+        return 0;
+      break;
+      }
+
+    while (TRUE)
+      {
+      int count, index;
+      named_group *ng;
+
+      switch (META_CODE(*pptr))
+        {
+        case META_OFFSET:
+        pptr++;
+        SKIPOFFSET(pptr);
+        continue;
+
+        case META_CAPTURE_NAME:
+        ng = cb->named_groups + pptr[1];
+        pptr += 2;
+        count = 0;
+        index = 0;
+
+        if (!PRIV(compile_find_dupname_details)(ng->name, ng->length, &index,
+          &count, errorcodeptr, cb)) return 0;
+
+        code[0] = OP_DNCREF;
+        PUT2(code, 1, index);
+        PUT2(code, 1 + IMM2_SIZE, count);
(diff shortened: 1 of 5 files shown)


NWilson:
Do we really need to remove duplicates? If a user (strangely) writes a scan_substring or `(?N:(...))` list with duplicated entries, can't we just ignore that, and process them left-to-right without caring?

We don't need to "optimise" things that are unnatural. Only handle the case if the code somehow breaks when there are duplicates.

zherczeg:
Not necessary, but useful in a few cases (e.g. machine generated patterns). The code is not that big either.

zherczeg:
I can remove the duplication check if you insist on it. I think the patch still improves the code without it.

src/pcre2_compile_cgroup.c
```diff
@@ -378,9 +333,197 @@ while (TRUE)
...
+    pptr[-1] = 0;
+    continue;
+
+    case META_CAPTURE_NUMBER:
+    pptr += 2;
+    if (PRIV(compile_is_capture_checked)(pptr[-1], captures, captures_end))
+      {
+      pptr[-1] = 0;
+      continue;
+      }
```
NWilson, line 446: I see what you mean. It really isn't much code to check for duplicates.

At the moment, regex compiling is O(n^2) because we do several linear searches in various places. I'm planning to hunt down and kill them, to get rid of denial-of-service attacks.

(There was a post about this on the pcre2-dev mailing list a while back. Some Erlang users wanted to be able to bound the amount of time that PCRE2 blocks the main loop. Making the parser worst-case O(n logn) would be a nice promise to make to users about how bad the pathological cases can get.)

Would it be really pedantic and annoying to drop the duplicate search, or else make a temporary sorted copy and do it that way instead?

No users will ever enter duplicates anyway.

NWilson (commented):
You're right, the code is better overall. I've left some comments on a couple of things though.

I don't want to block you if you disagree. They're not big things, and I don't feel too strongly.

src/pcre2_compile.c
```diff
@@ -6495,8 +6495,63 @@ for (;; pptr++)
...
+      named_group *ng;
+
+      switch (META_CODE(*pptr))
+        {
+        case META_OFFSET:
+        pptr++;
+        SKIPOFFSET(pptr);
+        continue;
+
+        case META_CAPTURE_NAME:
```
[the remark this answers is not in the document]
zherczeg: I don't think they can be merged. META_SCS starts a recursive call to process its block, and another opcode is needed to show that we are processing scan substring. We could change the code generator to support optional arguments (it currently expects a fixed size byte code), but that is complex enough.

Option: the 16 bit arg of META_OFFSET is not used, it could represent the different types of blocks if needed.

src/pcre2_compile_cgroup.c
zherczeg: I can use a bitset (each capture is one bit) as an alternative. The memory needed is (max_capture+7/8) bytes.

zherczeg:
The algorithm is changed to linear.

NWilson:
> I plan to land this patch soon

Terrific, thank you for the reminder to review it Zoltan and for making the changes.

NWilson (approved):

PR #723, opened 2025-03-07
JIT compiler update

zherczeg:


zherczeg:
@NWilson how can I fix the verify check fail?

NWilson:
You renamed README to README.md (which is fine).

The Makefile.am lists that file for inclusion in our tarball, so the reference there will need to updated. Then secondly, the manifest in `maint/manifest-tarball` will need to be updated too.

NWilson (approved):

PR #724, opened 2025-03-07
update build.zig for release of 0.14.0 breaking langugage changes

### LINKED ISSUE #722: Update build.zig against breaking changes in 0.14.0 release
I'm only making this Issue as a 'TODO' for whomever maintains the build.zig for this project. There is no insistence to complete this by any timeframe, I was simply in the strange position of both beginning to use zig and PCRE2 almost simultaneously as the new 0.14 spec was released.

I plan on using PCRE2 for completing advent of code challenges in zig. As well as for empowering the regex portion of what will eventually be a game engine (as I have been greatly impressed by it's use already in Godot).

for now of course I am happy to stay on 0.13 release

qaptoR:
fixes #722

notable changes are that `.root_module` fields needs dedicated `*Module` structs created (cannot be anonymous.

additionally, those modules are now how we add the C macros to the files being compiled.

also, for the pcre2test executable that we build, instead of having `linkLibrary` steps, I've directly added the `pcre2` and `posix` modules as imports into the .root_module of the executable.

lastly, two new default values were necessary to give values: `PCRE2GREP_BUFSIZE` and `_MAX_BUFSIZE` for the `config_header`.

None of this was explicitly outlined in the zig documentation. It was inferred from the std.Build, std.Build.Compile, and std.Build.Module listings in the zig std Library reference.


build.zig
```diff
@@ -17,7 +17,7 @@ pub fn build(b: *std.Build) !void {
 
     const config_header = b.addConfigHeader(
         .{
-            .style = .{ .cmake = b.path("src/config-cmake.h.in") },
+            .style = .{ .cmake = b.path("config-cmake.h.in") },
```
NWilson, line 20: There are a couple of regressions, here and below "pcre2_compile_cgroup.c".

build.zig
```diff
@@ -41,28 +41,33 @@ pub fn build(b: *std.Build) !void {
             .PCRE2_MAX_VARLOOKBEHIND = 255,
             .NEWLINE_DEFAULT = 2,
             .PCRE2_PARENS_NEST_LIMIT = 250,
+
+            .PCRE2GREP_BUFSIZE = 20480,
+            .PCRE2GREP_MAX_BUFSIZE = 1048576,
```
NWilson, line 46: I'm curious why these would be needed, given that the Zig build doesn't do pcre2grep currently.

build.zig
```diff
@@ -108,36 +112,47 @@ pub fn build(b: *std.Build) !void {
...
+        .link_libc = true,
+        .imports = &.{std.Build.Module.Import{ .name = b.fmt("pcre2-{s}", .{@tagName(codeUnitWidth)}), .module = lib_mod }},
     });
 
+    pcre2test_mod.addCMacro("HAVE_CONFIG_H", "");
+
+    const pcre2test = b.addExecutable(.{
+        .name = "pcre2test",
+        .root_module = pcre2test_mod,
+        .linkage = linkage,
```
NWilson, line 129: What does "linkage" mean for an executable? Altering the linkage to choose between `libpcre2-8.{so,a}` shouldn't force a property change on the pcre2test binary.

NWilson (commented):
You or I will also need to update the CI build to validate this against 0.14.

I don't know Zig at all, so I'll just leave this PR open for a few days (or a week maybe?) in case anyone else from the Zig community comes by to comment.

Then I'll merge it, if it all builds and the tests pass.

build.zig
qaptoR: Actually, the error messages I got I sort of 'tackled' progressively. It's possible that something I later had to implement actually over-rode the necessity. I'll test that right now.

build.zig
qaptoR: another great question. This was likely added in an effort to fix some error messages when I  hadn't yet removed the previous `linkLibrary()` steps.

I inferred the necessity by the executeable simply having the field at all, and the fact that we were 'linking libraries'. But once I switched to Importing the libraries into the pcre2test module, that likely was no longer a meaningful inference. I'll also test this now.

qaptoR:
@NWilson Definitely pause on merging.

I naively made this pull request assuming the build worked on master, but I forgot that the repo I was using to rewrite the build.zig file was a `release/10.45` branch (that I got following the steps in the Sheran test repo I had mentioned earlier).

It looks like the build.zig is failing now on master because of a missing `config-cmake.h.in`.
Looking at the directory listing of both branches, it looks like there's more changes to the build system.

so. this PR should be for the 10.45 branch. or just be halted until I can figure out how to do without those files.

build.zig
qaptoR: on the 10.45 branch (that I now realize I have always been testing on) it does seem like these are necessary to define. To what end, I can't tell, but it may be for uniformity with the rest of the `#defines` that are in `config-cmake.h.in`.

in fact, this might be the very part of the build.zig file that it turns out is unnecessary on master. I will test that.

build.zig
qaptoR: yes it does seem like this is a redundant line, likely because we are importing now, instead of linking. likely whatever error it initially suppressed was a red herring.

build.zig
qaptoR: oh! i see what you mean now. took me a while to understand this (namely figuring out that I wasn't on master!)

build.zig
qaptoR: for the record they are also necessary on master.

build.zig
NWilson: Note that we won't target PRs at old release branches like 10.45. Aiming for master is correct, since the next release will come from there.

build.zig
NWilson: Thank you, I see that they are needed due to Zig becoming stricter. Previously it didn't complain if some substitutions were not defined - it would just substitute null/empty-string.

PR #731, opened 2025-03-17
A few further tweaks to Zig 0.14 support

NWilson:
Followup to #722

* Fix Windows build warning by adding PCRE2_STATIC to pcre2test
* Remove unnecessary `.linkLibc()` command, now that we have `link_libc = true` in the Module declarations
* Update the old `std.Build.Step.Compile.create(...)` calls to the brand-new `b.addLibrary(...)` function. They added this exactly for our use-case, to allow dynamically selecting between `addStaticLibrary/addDynamicLibrary` variants.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -432,7 +432,7 @@ jobs:
         run: |
           # Zig does something weird with the stack - it uses more space than the
           # equivalent plain C program.
-          ulimit -S -s 16384
+          ulimit -S -s 65536
           srcdir=`pwd` pcre2test=`pwd`/zig-out/bin/pcre2test ./RunTest
 
   bazel:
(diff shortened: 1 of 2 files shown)


NWilson:
@qaptoR Here are a few small changes. I hope these work for you too.

NWilson:
Those bindings look interesting. I hope your project goes well.

Your attribution looks perfectly fine. You say you're using PCRE2, and you reproduce the license text. There's no magic formula or legally required wording for attribution. Permissive licenses like PCRE2's are designed to make this sort of reuse super-easy. Go for it!

qaptoR:
Thanks! I appreciate the feedback. So far my bindings are reproductions of the ones implemented by the Godot Engine (why I also have attribution for them). But as I learn more about regex, and especially PCRE, I think I will try to add support for more of the library.

PR #733, opened 2025-03-18
New pcre2_next_match() API to simplify pcre2demo, test, and substitute

### LINKED ISSUE #685: Add a match-iterator API
The largest chunk of detailed code inside `pcre2demo.c` involves explaining how to iterate over matches. This is non-trivial. The caller needs to try again at the end of the previous match location (in case `\G` does the bumpalong), and also try again (with "not-empty-at-start"), and if that doesn't match, then do a manual bump-along to search starting from the next character (but also skip `\r\n` as one sequence, if appropriate).

This logic is complex, and it's repeated in `pcre2_substitute`.

It may even be wrong, because the `\G` handling differs from Perl. In Perl, `\G` always means "end of previous match", but when there's a manual bump-along with repeated matches, PCRE2 gives it the behaviour of "start of current match process". We should at least offer an API to set a field on the match context, so that callers can specify a location for `\G` which is distinct from the start-offset used to begin the search for a match.

Let's offer a new API, "pcre2_match_iterator_create" which returns an object that handles this state. Then users can just call "pcre2_match_iterator_next" to go through all the matches. We'd use it ourselves in `pcre2_substitute`, but we'd also be able to make the demo code simpler and shorter.

NWilson:
Fixes #685 

* The primary purpose of pcre2_next_match() is to make it much easier for
  PCRE2 clients to iterate over matches, without needing an advanced knowledge
  of regular expressions. The current amount of code, and comments, in
  pcre2demo makes it look like we are placing a heavy burden on clients, when
  we could take that burden away by providing a function to do the work for them.
* Secondly, we can simplify our own code by merging the three duplicate
  implementations of the /g global match behaviour: pcre2demo, pcre2_substitute,
  and pcre2test.
* Thirdly, as I look closely at the issue, I can improve the documentation.
* Fourthly, I would like to actually simplify the logic, removing a complex loop
  which makes several match attempts, swallows duplicate matches, and more.
  We can have identical behaviour with a simple retry using
  PCRE2_NOTEMPTY_ATSTART.
* Implementation:
  - Add a new field start_offset to pcre2_match_data, so that pcre2_next_match()
    can inspect where the match attempt began, and ensure that the next attempt
    advances through the subject.


zherczeg:
This is a good idea

NWilson:
PR description updated.

I have polished the documentation, which I think is now clearer and more helpful.

The improvement in readability in pcre2demo is very nice: clients can actually use this API without pain!

The improved handling in pcre2_substitute is an extra bonus.

I am finished with this now.

NWilson:
Zoltan, are you happy for me to merge this new API function? I want to make sure you have a chance to check/veto it.

I'm happy to self-review the implementation, if you're OK with the API design. (I have also added a field to `match_data`, and I _think_ I have correctly updated the three locations that need to set it, whenever one of the three matching functions returns.)

zherczeg:
I haven't checked everything in detail, but I think this is a useful user function, and the patch looks good. The overhead of setting the `start_offset` probably negligible.

NWilson:
Thank you very much Zoltan!

PR #737, opened 2025-03-28
Remove colon from recursion with kept groups

zherczeg:
Since the list is enclosed in parenthesis, the colon is unnecessary.

@NWilson could you help me writing the documentation for this feature sometimes?


NWilson:
> @NWilson could you help me writing the documentation for this feature sometimes?

Yes, I'm sorry! I'll do it shortly...

NWilson:
> Since the list is enclosed in parenthesis, the colon is unnecessary.

By that logic, the colon is also unnecessary in scan_substring, which could be: `(?scs(grouplist)...)`

I guess you feel the colon in scan_substring is due to the fact that it's an "alpha assertion" (like napla), and those always end in colon?

Hmm. I can see it either way - we want the colon to be consistent with scan_substring; or alternatively, we don't want it because recusion isn't an alpha assertion, so the grouplist should follow immediately.

What made you change it?

zherczeg:
it is not `(?scs...)` it is `(*scs:...)`. All constructs with `(*` uses semicolon to terminate the name for consistency. As for `(?id)` the semicolon is not part of the syntax, and since the arguments are using the capture list format, the semicolon is unnecessary.

NWilson (approved):

PR #738, opened 2025-03-28
Add documentation for subroutine return values

NWilson:
It also appears that some of the syntax variants, such as `(?P>name(grouplist))` were not covered by tests, so I have added some extra examples to make sure.

To be honest, this is my first time using the feature, and I must say it's very nice.


zherczeg (approved):
LGTM! Thank you for writing the documentation.

PR #746, opened 2025-05-02
Add License variable to pkg-config file

iwamatsu:
The pkg-config file has License variable that that sets the software license.
This sets 'BSD-3-Clause WITH PCRE2-exception' as defined in SPDX to License.

Ref: https://github.com/pkgconf/pkgconf/blob/master/man/pc.5#L116


NWilson (approved):
This looks good. Thank you for the contribution! I will merge it.

PR #748, opened 2025-05-12
Fix heapsort in capture list construction

zherczeg:
It looks like I am bad at copying code.


NWilson (approved):
Nice! This is a satisfying fix. I haven't checked against a textbook, but from memory, this "looks right" for a heapify.

I'm very, very glad the fuzzer found this. Luckily it wasn't hard-to-hit.

PR #750, opened 2025-05-13
Add test output to make erroroffsets more readable

NWilson:
Currently, pcre2test prints the `erroroffset` as a simple number, which makes it impossible in the test output to see where it's actually pointing.

I have added printing of the relevant portion of the pattern, to show visually where the offset points.

There is also a new function `error_direction()` where I have started to classify which PCRE2 errors point to the left or right of the erroroffset. When I have fixed this all to be consistent to my satisfaction, I will document this so that clients have a better way to indicate which character is wrong in a regex.

This PR is almost test-only, with just a handful of tiny changes in pcre2_compile.c.


src/pcre2test.c
```diff
@@ -3217,6 +3244,84 @@ return i+1;
...
+Returns:      > 0  => the number of 16-bit units consumed
+              -1   => malformed UTF-16
+*/
+
+static int
+utf16_to_ord(PCRE2_SPTR16 utf16units, PCRE2_SPTR16 end, uint32_t *vptr)
+{
+uint32_t c = *utf16units++;
+
+if (c >= 0xDC00 && c <= 0xDFFF) return -1;
```
zherczeg, line 3269: Do we use uppercase for hex?

testdata/testoutput10
```diff
@@ -1603,14 +1618,17 @@ No match
 
 /(?'ABáC6789012345678901234567890123012345678901234567890123456789AB012345678901234567890123456789AB012345678901234567890123456789AB'...)/utf
 Failed: error 148 at offset 132: subpattern name is too long (maximum 128 code units)
+        here: ...23456789AB |<--| '...)
 
 /(?'ABZC6789012345678901234567890123012345678901234567890123456789AB012345678901234567890123456789AB012345678901234567890123456789AB'...)/utf
 
 /(?(n/utf
 Failed: error 142 at offset 4: syntax error in subpattern name (missing terminator?)
+        here: (?(n |<--|
```
zherczeg, line 1627: I often see using the caret (^)
```
/(?(n/utf
     ^
Failed: error 142 at offset 4: syntax error in subpattern name (missing terminator?)
```

zherczeg (commented):
Nice improvement!

src/pcre2test.c
NWilson: Apparently we do. I just extracted this code into a function, from elsewhere in the file.

testdata/testoutput10
NWilson: That would be a better user interface. I did consider it, and even attempted it. However, editors and terminals differ in how they render Unicode combining characters with multiple codepoints, for example characters with decomposing accents. Rendering the caret on the line below would be less precise for the purpose of a testing tool.

src/pcre2test.c
NWilson: Lowercase is far more common throughout PCRE2. I've reformatted these few uppercase ones here for consistency.

testdata/testoutput10
zherczeg: Ok

PR #753, opened 2025-05-21
Remove PCRE2_STATIC_RUNTIME

### LINKED ISSUE #752: `PCRE2_STATIC_RUNTIME` no longer works
`CMakeLists.txt` currently replaces the `CMAKE_<LANG>_FLAGS_<CONFIG>` variables directly in order to select a suitable MSVC runtime library:

https://github.com/PCRE2Project/pcre2/blob/098655996cfb1d5d8a527d5fd62fca1646cc5cd8/CMakeLists.txt#L840-L855

As of #542, however, the minimum CMake version is now 3.15, which means the `/MD` or `MT` flags are now provided by CMake's [`CMAKE_MSVC_RUNTIME_LIBRARY`](https://cmake.org/cmake/help/latest/variable/CMAKE_MSVC_RUNTIME_LIBRARY.html) variable and absent in these `FLAGS` variables, unless `CMAKE_POLICY_DEFAULT_CMP0091` is set to `OLD`.

Overriding `CMAKE_MSVC_RUNTIME_LIBRARY` at the command line now suffices, so it is best to remove `PCRE2_STATIC_RUNTIME` entirely, or make it modify `CMAKE_MSVC_RUNTIME_LIBRARY` instead. (Setting `CMAKE_POLICY_DEFAULT_CMP0091` to `NEW` also allows CMake 3.15 or above to apply `CMAKE_MSVC_RUNTIME_LIBRARY` to previous PCRE2 releases that had a lower minimum version in `CMakeLists.txt`.)

NWilson:
Since we have raised our minimum CMake version to 3.15, this updates the default value of the "CMP0091" backwards- compatibility policy, such that PCRE2_STATIC_RUNTIME is now broken (since 10.45 when cmake_minimum_required was increased).

We can remove this flag, since the new built-in CMAKE_MSVC_RUNTIME_LIBRARY provides the same behaviour.

Fixes #752


carenas (changes requested):
Need to also update the `bat` CI job in `dev.yml` so it will continue testing this configuration.

Bonus points for making sure that the test actually verifies the static runtime was actually linked with by the produced artifacts so it is not silently broken, and maybe tell cmake in the `Configure` step to fail when an "unknown variable" warning is triggered to make its failure more visible.

NWilson:
I have updated the dev.yml file. I have done a manual test that CMake's new built-in option works, using `dumpbin.exe`.

I have opened another PR with improvements to our CMake warnings.

carenas (approved):

Issue #549, opened 2024-11-06
Add erroroffset support to perltest.sh and fix discrepancies


NWilson:
* perltest.sh would need to support outputting error offsets from the Perl messages
* RunPerlTest would need to perform a "loose" diff on certain lines, to handle the fact that the Perl error messages won't have matching error codes
* Migrate as many error cases as possible from testinput{2,5} → testinput{1,4}
* Fix as many discrepancies as possible between the error offsets reported by Perl and PCRE2

There's no particular urgency to this. It would be nice though, perhaps when I have spare time and all the other more urgent tasks on my list are done.

NWilson:
I have made quite a few fixes. The three remaining cases are:

* PCRE2_ERROR_BAD_SUBPATTERN_REFERENCE
* PCRE2_ERROR_DEFINE_TOO_MANY_BRANCHES
* PCRE2_ERROR_BACKSLASH_K_IN_LOOKAROUND

There's a code TODO for each. These cases basically return a bad or unhelpful error-offset.

When they are fixed, I would consider extracting the pcre2test function `error_direction()` into a new public helper. It would indicate, for a given error value, whether the syntax issue is to the left or to the right of the error-offset.

The new function can be of the form `pcre2_get_error_offset_direction(compile_context* ctx)`, taking the compile context and returning additional error information.

For the use-cases we have, I have considered and basically ruled out converting the single error offset into a start/end span. That would be laborious, and we don't even know if anyone would ever use it. However, we would have the option in future of adding `pcre2_get_error_offset_length(compile_context* ctx)` to indicate the length of the span.

It basically comes down to "how many customers care about offering really nice syntax highlighting to pinpoint where the error is". For now, I think offering a good, high-quality single offset (with left/right direction) is good enough.

PR #756, opened 2025-06-02
Changes to error offsets, fixing inconsistencies

NWilson:
Part of #549

### DIFF
--- a/src/pcre2.h.generic
+++ b/src/pcre2.h.generic
@@ -345,6 +345,7 @@ pcre2_pattern_convert(). */
 #define PCRE2_ERROR_PERL_ECLASS_UNEXPECTED_CHAR    216
 #define PCRE2_ERROR_EXPECTED_CAPTURE_GROUP         217
 #define PCRE2_ERROR_MISSING_OPENING_PARENTHESIS    218
+#define PCRE2_ERROR_MISSING_NUMBER_TERMINATOR      219
 
 /* "Expected" matching error codes: no match and partial match. */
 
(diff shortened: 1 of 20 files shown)


jirislaby:
If I pass `".*.*$*X"` to `pcre2_compile()`, I get `erroroffset` at 6 (`X`) now. It used to be 5 (`*`). Is this expected?

NWilson:
In Perl, if I try the regex:

```perl
my $regex = qr/.*.*$*X/;
```

The error message is `$* is no longer supported as of Perl 5.30` which contains no character offset.

However if I try the (similar) `my $regex = qr/.*.*\K*X/` the error message is:

```
\K* matches null string many times in regex; marked by <-- HERE in m/.*.*\K* <-- HERE X/
```

Perl's error offset is consistently pointing to the caret position in between the `*` and `X`.

In `pcre2test` the same test gives:

```
$ ./pcre2test
PCRE2 version 10.48-DEV 2025-10-21 (8-bit)
  re> /.*.*$*X/
Failed: error 109 at offset 6: quantifier does not follow a repeatable item
        here: .*.*$* |<--| X
```

I am satisfied that this is consistent with the way Perl reports error messages, with the offset just to the right of the error.

The change was intentional. I apologise if this has caused problems for your application.

In a future version of PCRE2, I plan to add a function to retrieve the "error direction" to improve on the current "error offset". The error direction would indicate to text-based applications whether the error is `<-- HERE` to the left of the offset, or `HERE -->` to the right. I have considered adding a start/end position to the offset (for highlighting), but I have not heard any demand for that yet.

jirislaby:
> The change was intentional. I apologise if this has caused problems for your application.

Fine by me. Except the reporting has to be different for 10.47+ and earlier versions...

PR #758, opened 2025-06-09
Update INSTALL by adding cross-compile guide for OpenHarmony

kuniqi:
This is cross-compile guide of building the pcre2 in OpenHarmony, so we'd like to contribute it. Hope it will help the developers.


NWilson:
Thank you for the contribution.

Are you expecting people to use CMake or Autoconf on OpenHarmony? I'm surprised that you wouldn't direct users to the CMake build for a modern platform.

The text you've contributed is a link to a document which doesn't contain any details specific to PCRE2. Is that what you expected?

Secondly, the text interrupts the explanation about building "fat" binaries on MacOS. The paragraph afterwards explains how to use "lipo", which connects with the paragraph above.

kuniqi:
Yes, we using the cmake on OpenHarmony. And the real project we use is in this link (https://gitcode.com/openharmony-sig/tpc_c_cplusplus/tree/master/thirdparty/pcre2). And because we have our own build framework named lycium, so we just put the most common ways into the install file. 

For the wrong position of the install file, i'v move the lines after the 'lipo' tools line. Thanks again for your reply.

NWilson:
I'm sorry I never followed up on this PR.

I think at this point I'm going to close it. The reason is: you've just added a link to our docs, but not added any information or instructions to our repo. And the destination of the link doesn't have any information specific to PCRE2 either - it's very short and just says "run CMake and make to build 3rd-party software".

Basically, it sounds like PCRE2 just works out of the box, which is great.

PR #768, opened 2025-08-06
pcre2_match_data: allow pcre2_match_data_create_from_pattern() with a NULL as the first parameter

carenas:
would allow something like:

```c
re = pcre2_compile(...);
can_jit = !pcre2_jit_compile(re, PCRE2_JIT_TEST_ALLOC);
md = pcre2_match_data_create_from_pattern(re, ...);
if (re != NULL && md != NULL) do {
  /* do some work here, including matches and maybe even some on demand JIT compilation */
} while (!done_matching_or_replacing)
/* do some cleanup here, no need to check for NULL when calling the corresponding `_free()` functions */
```

which matches a similar construct that uses pcre2_match_data_create() instead.

### DIFF
--- a/doc/pcre2_match_data_create_from_pattern.3
+++ b/doc/pcre2_match_data_create_from_pattern.3
@@ -15,19 +15,19 @@ PCRE2 - Perl-compatible regular expressions (revised API)
 .rs
 .sp
 This function creates a new match data block for holding the result of a match.
-The first argument points to a compiled pattern. The number of capturing
-parentheses within the pattern is used to compute the number of pairs of
-offsets that are required in the match data block. These form the "output
-vector" (ovector) within the match data block, and are used to identify the
-matched string and any captured substrings when matching with
-\fBpcre2_match()\fP. If you are using \fBpcre2_dfa_match()\fP, which uses the
-output vector in a different way, you should use \fBpcre2_match_data_create()\fP
-instead of this function.
+If the first argument is NULL, this function returns NULL, otherwise the first
+argument points to a compiled pattern. The number of capturing parentheses
+within the pattern is used to compute the number of pairs of offsets that are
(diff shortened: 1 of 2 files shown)


zherczeg (approved):
LGTM

doc/pcre2_match_data_create_from_pattern.3
```diff
@@ -15,19 +15,18 @@ PCRE2 - Perl-compatible regular expressions (revised API)
...
 This function creates a new match data block for holding the result of a match.
-The first argument points to a compiled pattern. The number of capturing
-parentheses within the pattern is used to compute the number of pairs of
-offsets that are required in the match data block. These form the "output
-vector" (ovector) within the match data block, and are used to identify the
-matched string and any captured substrings when matching with
-\fBpcre2_match()\fP. If you are using \fBpcre2_dfa_match()\fP, which uses the
-output vector in a different way, you should use \fBpcre2_match_data_create()\fP
-instead of this function.
+If the first argument is NULL, this function returns NULL, otherwise The first
```
NWilson: typo: `The` → `the`

NWilson (approved):
I suppose I can see why someone may want to do this, and it never hurts to add a null-check to make the behaviour defined. OK!

doc/pcre2_match_data_create_from_pattern.3
carenas: Agree it doesn't hurt, BUT it also makes the API use less defined, which is why my first version of the PR comment had a "maybe" and ended in a question mark.

For example, `pcre2_jit_compile()` wouldn't accept a NULL pointer normally and will crash deservedly, because. it clearly says it expects a compiled expression, and NULL is not it, and I wouldn't expect us to change that behavior like in this case.

PR #773, opened 2025-08-15
pcre2test: improve `expand` and its use of buffers

carenas:
Mainly as a way to clean the code and make it more resilient.

It changes the behavior of the pattern `/\[]{-1}/expand` which will now be treated as a literal instead of triggering a bogus syntax error and clarifies that no pattern or buffer could be larger than SIZE_MAX bytes.

Improves error checking for 32-bit platforms and even 64-bit ones that are not LP64.

### DIFF
--- a/src/pcre2test.c
+++ b/src/pcre2test.c
@@ -343,10 +343,22 @@ ints. They are defined not to be shorter. */
 #define U32OVERFLOW(x) (x == UINT32_MAX)
 #endif
 
+#if ULONG_MAX > UINT32_MAX
+#define U32OVERFLOWE(x) U32OVERFLOW(x)
+#else
+#define U32OVERFLOWE(x) (errno != 0 && x == UINT32_MAX)
+#endif
+
 #if LONG_MAX > INT32_MAX
-#define S32OVERFLOW(x) (x > INT32_MAX || x < INT32_MIN)
+#define S32OVERFLOW(x) (x > INT32_MAX || INT32_MIN > x)
 #else
 #define S32OVERFLOW(x) (x == INT32_MAX || x == INT32_MIN)
+#endif
+
+#if LONG_MAX > INT32_MAX
+#define S32OVERFLOWE(x) S32OVERFLOW(x)
+#else
+#define S32OVERFLOWE(x) (errno != 0 && (x == INT32_MAX || x == INT32_MIN))
 #endif
 
 /* When PCRE2_CODE_UNIT_WIDTH is zero, pcre2_internal.h does not include
@@ -6007,7 +6019,6 @@ static int
 process_pattern(void)
 {
 BOOL utf;
-uint32_t k;
 uint8_t *p = buffer;
 unsigned int delimiter = *p++;
 int errorcode;
@@ -6100,7 +6111,7 @@ if (pat_patctl.convert_type != CONVERT_UNSET &&
 /* Check for mutually exclusive control modifiers. At present, these are all in
 the first control word. */
 
-for (k = 0; k < sizeof(exclusive_pat_controls)/sizeof(uint32_t); k++)
+for (uint32_t k = 0; k < sizeof(exclusive_pat_controls)/sizeof(uint32_t); k++)
   {
   uint32_t c = pat_patctl.control & exclusive_pat_controls[k];
   if (c != 0 && c != (c & (~c+1)))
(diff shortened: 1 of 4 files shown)


src/pcre2test.c
```diff
@@ -6166,59 +6165,68 @@ else if ((pat_patctl.control & CTL_EXPAND) != 0)
...
+      the pattern need to backtrack to avoid jumping over the NUL. */
+      if (*pe == 0) pp--;
       }
 
     /* Add to output. If the buffer is too small expand it. The function for
     expanding buffers always keeps buffer and pbuffer8 in step as far as their
     size goes. */
 
-    while (pt + count * length > pbuffer8 + pbuffer8_size)
+    while (m > pbuffer8_size)
```
NWilson: Could change while to `if` (if expand_input_buffers is improved to actually respect `minimum`)

src/pcre2test.c
```diff
@@ -3838,37 +3838,36 @@ return 0;
...
-uint8_t *new_pbuffer8 = (uint8_t *)malloc(new_pbuffer8_size);
+size_t buffer_size;
 
-if (new_buffer == NULL || new_pbuffer8 == NULL)
-  {
-  fprintf(stderr, "pcre2test: malloc(%" SIZ_FORM ") failed\n",
-          new_pbuffer8_size);
-  exit(1);
-  }
+buffer_size = (pbuffer8_size < SIZE_MAX / 2)? 2 * pbuffer8_size : minimum;
```
NWilson: Wouldn't it make sense to expand it to `minimum`, if the min size is larger than the doubled size?

src/pcre2test.c
```diff
@@ -6166,59 +6165,68 @@ else if ((pat_patctl.control & CTL_EXPAND) != 0)
...
     uint32_t count = 1;
     size_t length = 1;
+    size_t m = 1;
 
     /* Check for replication syntax; if not found, the defaults just set will
-    prevail and one character will be copied. */
+    prevail and `length` characters will be copied once. */
 
-    if (pp[0] == '\\' && pp[1] == '[')
+    if (memcmp(pp, "\\[", 2) == 0)
```
NWilson: The previous code was (implicitly) checking whether `pp[0]` is NUL before checking `pp[1]`. It's not terrible - I don't _think_ that it's possible to read one-byte over the end of the buffer with the memcmp, but you are relying here on knowing that the buffer is oversized.

As a small nit, I find it marginally safer and preferable to avoid using memcmp on null-terminated strings unless you know that the length fits.

src/pcre2test.c
```diff
@@ -6166,59 +6165,68 @@ else if ((pat_patctl.control & CTL_EXPAND) != 0)
...
     size goes. */
 
-    while (pt + count * length > pbuffer8 + pbuffer8_size)
+    while (m > pbuffer8_size)
       {
       size_t pc_offset = pc - buffer;
       size_t pp_offset = pp - buffer;
       size_t pt_offset = pt - pbuffer8;
-      expand_input_buffers();
+      expand_input_buffers(m);
```
NWilson: This thing with `m` can't be correct, can it? It looks like `m` is being set to the size of the current expansion, not the size of what's in the buffer plus the new content.

What happens for this: `/[a]{1000}[b]{1000}/expand`

src/pcre2test.c
carenas: The point of doing exponential expansion in `expand_input_buffers()`, specially now that we are actually using `realloc()` is to help the allocator so it is unlikely to need to move the data (just assign more virtual pages) hence keeping it as a multiple of 2 (which will also likely be a multiple of the pagesize) is preferred IMHO.

src/pcre2test.c
carenas: Good point; thanks for the insight about `memcmp()`

src/pcre2test.c
NWilson: Yes - it should allocate a power of two if possible. However, the code appears to simply double the buffer, even if the doubled size is less than minimum!

NWilson:
I apologize, I think I have created a ton of conflicts in this branch after my splitting of `pcre2test.c`.

I will rebase your changes if you would like, just ask.

NWilson:
@carenas I don't know if any of these changes are still relevant? I will close if there's no desire to rebase and update this.

PR #776, opened 2025-08-18
maint: introduce LintMan to aid on tracking/updating values and fix pcre2limits.3 for the correct name length

carenas:
Correct a minor issue with the documentation/configuration and introduce a tool to help prevent similar issues in the future.

The tracking is not intrusive (meaning that the documentation is still valid nroff) and independent of the build system, but I am curious on comments and alternatives or where it fits in the development process.

### DIFF
--- a/maint/LintMan
+++ b/maint/LintMan
@@ -6,7 +6,7 @@ use Getopt::Long;
 use vars qw /$opt_verbose/;
 
 # A script to scan PCRE2's man pages to check for values that might need to
-# be updatd to match the code.
+# be updated to match the code.
 #
 # It updates numerical values after \" DEFINE <name> or errors if name is
 # not found.
@@ -14,7 +14,7 @@ use vars qw /$opt_verbose/;
 my $file;
 my %defs;
 
-foreach $file ("../src/config.h")
+foreach $file ("../src/config.h.generic")
   {
   open (INCLUDE, $file) or die "Failed to open include $file\n";
 
--- a/maint/UpdateAlways
+++ b/maint/UpdateAlways
@@ -56,7 +56,7 @@ echo Processing documentation
 perl ../maint/CheckMan *.1 *.3
 if [ $? != 0 ] ; then exit 1; fi
 
-if [ -f ../src/config.h ] ; then
+if [ -f ../src/config.h.generic ] ; then
   perl ../maint/LintMan -v *.3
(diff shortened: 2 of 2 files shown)


NWilson:
I like it! Thanks!

Is it possible to call LintMan from the UpdateAlways script. Otherwise, we'll never remember to check it.

carenas:
Since it modifies the repository and requires a configured tree it is more suitable for a manual run in a developer tree (maybe as a pre-push hook if you want to automate the checks).

Goes without saying that changes to the tree it does, should be committed or discarded (if not relevant) and that is better done by a human IMHO, but had added it to the script in a way that is safe for the CI use of it as requested.

maint/LintMan
```diff
@@ -0,0 +1,66 @@
...
+# A script to scan PCRE2's man pages to check for values that might need to
+# be updated to match the code.
+#
+# It updates numerical values after \" DEFINE <name> or errors if name is
+# not found.
+
+my $file;
+my %defs;
+
+foreach $file ("../src/config.h.generic")
```
NWilson, line 17: I made one change here. If the user configures their PCRE2 differently, they should still (probably) get the same documentation is everyone else?

I've made the docs match the generic config. Users don't receive the maintainer scripts in `maint/` as part of a tarball release, so they won't be customising the values in the docs for their installation anyway.

Finally, I've checked that it's running nicely in CI - and it seems to work well. Great!

carenas: but isn't config.h.generic created by the Makefile?

in that scenario, if a new setting is added to the documentation then this will break CI, right?

NWilson: Well, if you want to reference a new constant using your `.\" DEFINE` syntax then it has to be defined in config.h. That's not really changed from what you did. `config.h.generic` is just an exact copy of `config.h` when you run `./configure` with no arguments, so the two config files always have the same constants in.

carenas: but I only change config.h.in, and `configure` creates config.h, and after `make` runs then `config.h.generic` is created from it.

but the CI checks for that last file, which hasn't been updated yet because `configure` and `make` didn't run yet, right?, and when `LintMan` runs finds a `DEFINE` that doesn't exist in that file and dies.

NWilson: CI ensures that `config.h.generic` is kept up to date.

PR #777, opened 2025-08-21
build.zig: update to zig 0.15

carenas:
Syntax is compatible with 0.14 and now can generate cross compiled dynamic builds that run on Windows

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -473,14 +473,11 @@ jobs:
           submodules: true
 
       - name: Build
-        run: zig build
+        run: zig build -Dsupport_jit
 
       - name: Test
         run: |
-          # Zig does something weird with the stack - it uses more space than the
-          # equivalent plain C program.
-          ulimit -S -s 65536
-          srcdir=`pwd` pcre2test=`pwd`/zig-out/bin/pcre2test ./RunTest
+          srcdir=`pwd` pcre2test=`pwd`/zig-out/bin/pcre2test ./RunTest -bigstack
 
   bazel:
     # Tests with: Bazel build system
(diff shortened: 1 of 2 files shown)


NWilson:
The Ubuntu snap version of zig doesn't seem to have updated to Zig 0.15 yet, so I'll wait until the CI is running 0.15 before merging this. (Let me know if that doesn't make sense.)

NWilson:
I have cherry-picked your `LC_ALL=C` fix, while I wait for the version of Zig in CI to update to 0.15

build.zig
```diff
@@ -149,6 +152,8 @@ pub fn build(b: *std.Build) !void {
         posixLib_mod.addCMacro("PCRE2_CODE_UNIT_WIDTH", @tagName(codeUnitWidth));
         if (linkage == .static) {
             posixLib_mod.addCMacro("PCRE2_STATIC", "");
+        } else {
+            posixLib_mod.addCMacro("PCRE2POSIX_DYNAMIC", "");
```
NWilson:
```suggestion
            posixLib_mod.addCMacro("PCRE2POSIX_SHARED", "");
            pcre2test_mod.addCMacro("PCRE2POSIX_SHARED", "");
```

PCRE2POSIX_DYNAMIC doesn't exist. I think you meant SHARED.

Also, it should be added to pcre2test_mod as well.

build.zig
```diff
@@ -167,6 +172,8 @@ pub fn build(b: *std.Build) !void {
             },
         });
 
+        posixLib.linkLibrary(lib);
```
NWilson, line 175: You replaced all but one of the `addImport` -> `linkLibrary`.

The remaining last one to replace is the dependency of pcre2test on pcre2_posix.

.github/workflows/dev.yml
```diff
@@ -473,14 +473,11 @@ jobs:
           submodules: true
 
       - name: Build
-        run: zig build
+        run: zig build -DJIT
```
NWilson: Actually, could we rename `-DJIT` to `-DSUPPORT_JIT`, so that the zig build options match CMake? (Suggestion)

carenas: zig options are lowercase, and it looked like a sore thumb when I tried originally (note cmake uses PCRE2_SUPPORT_JIT), so I actually used "JIT" for "compatibility", but "support_jit" looks better.

build.zig
carenas: Ironically, this fixes the last issue with zip 0.14 that I thought were bugs in that version.

NWilson: I love it, that's lucky!

PR #778, opened 2025-08-22
pcre2test: improve `binmode`

carenas:
Simplify the code and print all parameters

### DIFF
--- a/doc/pcre2pattern.3
+++ b/doc/pcre2pattern.3
@@ -1,4 +1,4 @@
-.TH PCRE2PATTERN 3 "29 August 2025" "PCRE2 10.47-DEV"
+.TH PCRE2PATTERN 3 "2 September 2025" "PCRE2 10.47-DEV"
 .SH NAME
 PCRE2 - Perl-compatible regular expressions (revised API)
 .SH "PCRE2 REGULAR EXPRESSION DETAILS"
@@ -3186,8 +3186,7 @@ For example:
   (?(VERSION>=10.4)yes|no)
 .sp
 This pattern matches "yes" if the PCRE2 version is greater or equal to 10.4, or
-"no" otherwise. The fractional part of the version number may not contain more
-than two digits.
+"no" otherwise. The fractional part of the version number could be ommited.
 .
 .
 .SS "Assertion conditions"
@@ -4231,6 +4230,6 @@ Cambridge, England.
 .rs
 .sp
 .nf
-Last updated: 29 August 2025
+Last updated: 2 September 2025
 Copyright (c) 1997-2024 University of Cambridge.
 .fi
--- a/doc/pcre2syntax.3
+++ b/doc/pcre2syntax.3
@@ -1,4 +1,4 @@
-.TH PCRE2SYNTAX 3 "28 March 2025" "PCRE2 10.47-DEV"
+.TH PCRE2SYNTAX 3 "2 September 2025" "PCRE2 10.47-DEV"
 .SH NAME
 PCRE2 - Perl-compatible regular expressions (revised API)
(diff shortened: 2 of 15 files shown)


zherczeg (approved):
LGTM! Nice!

NWilson:
Thank you Carlo for all these PRs!

I've been extremely busy this week at home and hardly had a minute to look at anything PCRE2. I'm sorry - I'll deal with all of them next week.

src/pcre2_internal.h
```diff
@@ -1988,7 +1988,8 @@ some cases doesn't actually use these names at all). */
   "Cond",                                                         \
   "SBra", "SBraPos", "SCBra", "SCBraPos",                         \
   "SCond",                                                        \
-  "Capture ref", "Capture dnref", "Cond rec", "Cond dnrec",       \
+  "Capture ref", "Capture ref <",                                 \
+  "Cond recurse", "Cond recurse <",                               \
```
NWilson: Why are you using "<" for dnref and dnrec? I think "dn" is "dynamic", for cases where there's a capture group specified by name (not by number), and where the names are not unique, so the name-to-number resolution is done dynamically at match time.

NWilson: Ah I see. You moved the "<" from printinc.h.

I'd keep these names here unchanged, and later use `fprintf(f, " %s %s<", flag, OP_names[*code]);`

doc/pcre2syntax.3
```diff
@@ -627,22 +627,26 @@ following ways:
...
+  (?(VERSION[>]=n[.m]) test PCRE2 version
+  (?(assert)           assertion condition
 .sp
 Note the ambiguity of (?(R) and (?(Rn) which might be named reference
 conditions or recursion tests. Such a condition is interpreted as a reference
 condition if the relevant named group exists.
+.sp
+The parts within brackets for the VERSION conditional syntax could be ommited.
+The fractional part of the version number defaults to 0, and can be up to two
+digits long.
```
NWilson: Where do you get "can be up to two digits long"? It appears to support up to `999.999` (big enough for any version we'll ever use - and so probably no need to specify what the limit is).

NWilson: But good spot that the fractional part is optional! Nice doc improvement.

carenas: it came from `pcre2pattern.3`, will remove from both, but I think the limit was documented to allow 10.04 to work like 10.4 but not 10.0004, even if 10.004 might.

doc/pcre2pattern.3
```diff
@@ -1,4 +1,4 @@
-.TH PCRE2PATTERN 3 "29 August 2025" "PCRE2 10.47-DEV"
+.TH PCRE2PATTERN 3 "2 September 2025" "PCRE2 10.47-DEV"
```
NWilson, line 1: In future, you don't worry about updating this manually in the PR. I'll just do it once automatically using `maint/UpdateDates` before a release. It's completely harmless to do manually, just not necessary.

NWilson (approved):
Thanks! That looks tidy

doc/pcre2pattern.3
carenas: Isn't this file used to generate the documentation in github though or that also gets its dates updated automatically?, Philip's argument for updating the date in this case was that API changes that might be relevant since the previous release are not easily identifiable.

NWilson: The `*.1` and `*.3` files are canonical, you are correct. The dates however can be updated automatically, to simply match the last commit date in the history (excluding commits which only bump the version number).

carenas: ok, I am not clear though, if by that you mean that they are being updated already whenever the documentation gets updated or not.

PR #781, opened 2025-08-25
pcre2_match: correctly reject overlong matches for backreferences

carenas:
Address a bug in `pcre2_match()` that was made visible by the test case of another bug for scan substring.

Filed as a draft for feedback and to get a better test and impact.

Requires: #771


src/pcre2test.c
```diff
@@ -8478,11 +8481,15 @@ the unused start of the buffer unaddressable. If we are using the POSIX
...
-  VALGRIND_MAKE_MEM_NOACCESS(dbuffer, dbuffer_size - (len + c));
+c = code_unit_size * ((((pat_patctl.control & CTL_POSIX) != 0) +
+                       ((dat_datctl.control & CTL_ZERO_TERMINATE) != 0))? 1 : 0);
+
+#ifndef SUPPORT_VALGRIND
+pp = memmove(dbuffer + dbuffer_size - (len + c), dbuffer, len + c);
+#else
+pp = memmove(dbuffer + dbuffer_size - (len + c + x), dbuffer, len + c);
+VALGRIND_MAKE_MEM_NOACCESS(dbuffer, dbuffer_size - (len + c + x));
+VALGRIND_MAKE_MEM_NOACCESS(dbuffer + dbuffer_size - x, code_unit_size);
```
NWilson: You're adding an extra unit at the end of the buffer. But is this actually necessary? Can you give a test case where it makes a difference?

Valgrind will already warn (loudly) if you read past the end of the buffer, so extending the buffer and then telling it not to read the final bytes seems like extra complexity.

And - it makes it potentially _less_ likely for bugs to be caught. Before, a standard build without `SUPPORT_VALGRIND` would catch read-past-the-end errors (if run under valgrind), whereas now you actually have to specially build the library to make the VALGRIND_MAKE_MEM_NOACCESS calls in order for this to work.

carenas: The extra unit was only added when building with `SUPPORT_VALGRIND` so the normal usage (including running with valgrind a build that didn't have that) was not affected, but agree that the extra complexity was unwarranted

carenas:
this should be ready to merge (better rebase or cherry-pick though) to master once the release branch is merged back.

src/pcre2_match.c
```diff
@@ -485,8 +486,8 @@ else
 
   else
     {
-    if ((PCRE2_SIZE)(mb->end_subject - eptr) < length) return 1; /* Partial */
-    if (memcmp(p, eptr, CU2BYTES(length)) != 0) return -1;  /* No match */
+    if ((PCRE2_SIZE)(mb->end_subject - eptr) < length ||
+        memcmp(p, eptr, CU2BYTES(length)) != 0) return -1;  /* No match */
```
NWilson, line 490: I can see that the original code here is wrong: it's returning 1 (Partial) even in cases where it's definitely not a partial match.

Can you confirm to me what the impact of this is? As far as I can see, when `partial == 0` (neither hard nor soft partial mode) the return value when `match_ref != 0` is ignored, and the distinction between `+/-1` doesn't matter.

On the other hand, if `partial != 0` then SCHECK_PARTIAL will distinguish between positive and negative returns from match_ref.

However, there's one detail I'm not 100% convinced of: does it make any difference to do `Feptr = mb->end_subject` just before `RRETURN(MATCH_NOMATCH)`? I think not - the end pointer is ignored in this case? But it's used in too many places for me to confirm quickly.

---

tldr: Carlo, are you actually fixing a bug here? Based on your understanding of this change, is there any observable effect on behaviour?

carenas: yes, saving some useless work triggered by the engine retrying in a new frame for a partial match that would never succeed instead of failing to match at the first try if partial match wasn't requested.

NWilson: I don't understand. Are you saying that there is a change to the behaviour here?

My understanding was, that when `partial == 0` it doesn't make any difference whether you return `1 / -1`. But you're saying that this change has an impact on execution paths?

carenas: Nevermind; you are correct and I was mistaken. There is no change of behaviour, except for incorrectly setting `Feptr = mb->end_subject`, which is not relevant.

NWilson: Thank you! I just like to be sure I understand the change (especially in core match logic) before I merge it.

As long as we're agreed, and there's no bug, I think the change makes sense.

PR #782, opened 2025-08-31
maint: build fix for ucptest and related code tyding up

carenas:
Mainly fix a simple build failure because of the removal of the "config.h" include, but also some "fixes" for a "-Weverything" build and other tidying up of the code that made sense.


src/pcre2_ord2utf.c
```diff
@@ -80,16 +80,17 @@ PRIV(ord2utf)(uint32_t cvalue, PCRE2_UCHAR *buffer)
 /* Convert to UTF-8 */
 
 #if PCRE2_CODE_UNIT_WIDTH == 8
-int i, j;
+unsigned int i;
+
 for (i = 0; i < PRIV(utf8_table1_size); i++)
   if ((int)cvalue <= PRIV(utf8_table1)[i]) break;
```
NWilson, line 86: Wow, this writes out a rather weird & ill-formed sequence if you pass in `cvalue = 0xffff_ffff`. However that value shouldn't be passed in, and the code doesn't do any bad loops or out-of-bounds access. So, I guess it's OK.

NWilson (approved):
Great, thanks!

NWilson:
Ah - the CI is failing. It looks like you need to find all the other uses of `utf8_table1_size` (now that you've made it unsigned) and convert them also to use it in an unsigned context.

src/pcre2_ord2utf.c
carenas: correct, I almost threw it a PCRE2_ASSERT there as that function shouldn't be getting anything over 0x10ffff anyway AFAIK, at least when called from the library.

src/pcre2test.c
```diff
@@ -4026,11 +4026,13 @@ Returns:    < 0, = 0, or > 0, according to the comparison
 static int
 strncmpic(const uint8_t *s, const uint8_t *t, size_t n)
 {
-while (n--)
+if (n > 0) do
```
carenas, line 4029: this triggers the integer sanitizer in Apple clang version 17.0.0 (clang-1700.0.13.5).

It is likely bogus though as `n--` (even when it wraps around at 0) in an unsigned is not UB, and because it is not used after the wraparound is not even relevant.

NWilson: Interesting. The wraparound for size_t is defined behaviour, but I guess the Apple devs wanted to give special behaviour for size_t, since it's a special type that basically never is intended to wrap around.

Personally I'd have maybe done a simple `for (; n > 0; --n)` loop as the clearest. But yours appears to be identical in behaviour, so let's keep it.

NWilson (approved):

src/pcre2test.c
carenas: It triggers with all (even char) unsigned integers though, so I think it is just a bug.

Remember I tracked it down before and it seemed to be intentional and there were suggestions from the developer (that I think was from Google) to add a suppression syntax though as a "solution", but this is definitely one of the reasons why gcc wouldn't enable the "integer" UB library, it I recall correctly.

NWilson: Wow, OK. Signed wraparound shouldn't warn! We should only fix the ones (like this on) where it's accidental wraparound that we don't need to do. If we have legitimate uses of wraparound then we shouldn't complicate the code to satisfy warning spam.

PR #786, opened 2025-09-01
maint: teach utf8 how to encode utf8_input

carenas:
Hopeful not controversial, but I have to admit a little more involved than needed, implementation of a feature for `utf8` (the maint tool).

It also fixes a bug in EBCDIC systems, and even does some text corrections that are hopeful right as a non native english speaker and clarify things.

### DIFF
--- a/src/pcre2test.c
+++ b/src/pcre2test.c
@@ -3746,6 +3746,7 @@ if (!utf && (pat_patctl.control & CTL_UTF8_INPUT) == 0)
   {
   for (; len > 0; len--) *pp++ = *p++;
   }
+
 else while (len > 0)
   {
   uint32_t c;
@@ -3830,7 +3831,6 @@ if (pbuffer32_size < 4*len + 4)
   }
 
 pp = pbuffer32;
-
 if (!utf && (pat_patctl.control & CTL_UTF8_INPUT) == 0)
   {
   for (; len > 0; len--) *pp++ = *p++;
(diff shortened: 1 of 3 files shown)


NWilson:
> It also fixes a bug in EBCDIC systems

What is the EBCDIC bug? The `utf8.c` tool certainly doesn't need to run on EBCDIC systems.

src/pcre2test.c
```diff
@@ -8191,15 +8191,22 @@ while ((c = *p++) != 0)
...
+#endif
     if ((utf || (pat_patctl.control & CTL_UTF8_INPUT) != 0) &&
-      HASUTF8EXTRALEN(c)) { GETUTF8INC(c, p); }
+        HASUTF8EXTRALEN(c))
+      {
+      GETUTF8INC(c, p);
+      }
+#ifdef SUPPORT_PCRE2_32
     c |= topbit;
+#endif
```
NWilson: This change appears to be harmless (no impact on execution that I can see).

You could save the second ifdef - just change to:

```c
    uint32_t topbit = 0;
#ifdef SUPPORT_PCRE2_32
    if (test_mode == PCRE32_MODE && c == 0xff && *p != 0)
      {
      topbit = 0x80000000;
      c = *p++;
      }
#endif
    if ((utf || (pat_patctl.control & CTL_UTF8_INPUT) != 0) &&
        HASUTF8EXTRALEN(c))
      {
      GETUTF8INC(c, p);
      }
    c |= topbit;
```

carenas:
> What is the EBCDIC bug?

`'a' - 10` was assumed to always be `W` in:

https://github.com/PCRE2Project/pcre2/blob/498bef0b77069b447a6b78362ff2eb011037c4af/maint/utf8.c#L274

NWilson:
OK. We don't need anything in `maint/` to work on EBCDIC systems, but the fix is OK.

PR #788, opened 2025-09-02
maint: portability improvements for UpdateAlways

carenas:
Mainly to allow running this script in macOS and fixing some of the issues that it found with GNU groff version 1.23.0 and the system provided mandoc.


maint/UpdateAlways
```diff
@@ -162,21 +154,22 @@ perl <<"END" >pcre2demo.3
...
-done
+if man --help 2>/dev/null | grep -q warnings ; then
+  for file in *.1 *.3 ; do
+    MAN_OUT=`MANROFFSEQ='' MANWIDTH=80 man --warnings=w,all -E UTF-8 -l -Tutf8 -Z "$file" 2>&1 >/dev/null`
+    if [ "$MAN_OUT" != "" ]; then
+      printf "Running man generated warnings:\n%s\n" "$MAN_OUT"
+      exit 1
+    fi
+  done
+fi
```
NWilson, line 176: I'd put an `else echo "Skipping man page check (non-GNU system)";` at the end, so that we don't accidentally silently skip this check if some bitrot occurs (eg if `man --help` ever changes in future). Just to increase the chance we notice.

NWilson (approved):
Nice, thanks. I hadn't expected to run `UpdateAlways` on Mac, but it's nice to see it's not too hard.

NWilson (approved):

PR #789, opened 2025-09-03
Silence clang compiler warning: unannotated fall-through between switch labels

Gsus42:
Hi there,

we are using clang with -Wimplicit-fallthrough and -Werror enabled which helped us identify a few odd places in our code base. In the PCRE2 code base those places have already been commented and I propose to use the corresponding compiler attribute to make the compiler aware the fact that fallthrough is allowed at these places.

### DIFF
--- a/src/pcre2_auto_possess.c
+++ b/src/pcre2_auto_possess.c
@@ -804,21 +804,21 @@ for(;;)
 
       case OP_NOT_DIGIT:
       invert_bits = TRUE;
-      /* Fall through */
+      __attribute__((fallthrough));
       case OP_DIGIT:
       set2 = (const uint8_t *)(cb->cbits + cbit_digit);
       break;
 
       case OP_NOT_WHITESPACE:
       invert_bits = TRUE;
-      /* Fall through */
+      __attribute__((fallthrough));
       case OP_WHITESPACE:
       set2 = (const uint8_t *)(cb->cbits + cbit_space);
       break;
 
       case OP_NOT_WORDCHAR:
       invert_bits = TRUE;
-      /* Fall through */
+      __attribute__((fallthrough));
       case OP_WORDCHAR:
       set2 = (const uint8_t *)(cb->cbits + cbit_word);
       break;
@@ -1101,7 +1101,7 @@ for(;;)
 
       case OP_NCLASS:
       if (chr > 255) return FALSE;
-      /* Fall through */
+      __attribute__((fallthrough));
 
       case OP_CLASS:
       if (chr > 255) break;
--- a/src/pcre2_study.c
+++ b/src/pcre2_study.c
@@ -175,7 +175,7 @@ for (;;)
       cc += 1 + LINK_SIZE;
       break;
       }
-    /* Fall through */
+    __attribute__((fallthrough));
 
     case OP_ONCE:
     case OP_SCRIPT_RUN:
@@ -252,7 +252,7 @@ for (;;)
     case OP_ASSERT_SCS:
     case OP_ASSERTBACK_NA:
     do cc += GET(cc, 1); while (*cc == OP_ALT);
-    /* Fall through */
+    __attribute__((fallthrough));
 
     /* Skip over things that don't match chars */
 
@@ -352,7 +352,7 @@ for (;;)
     case OP_PROP:
     case OP_NOTPROP:
     cc += 2;
-    /* Fall through */
+    __attribute__((fallthrough));
 
     case OP_NOT_DIGIT:
     case OP_DIGIT:
@@ -433,7 +433,7 @@ for (;;)
       case OP_CRMINPLUS:
       case OP_CRPOSPLUS:
       branchlength++;
-      /* Fall through */
+      __attribute__((fallthrough));
 
       case OP_CRSTAR:
       case OP_CRMINSTAR:
@@ -1303,7 +1303,7 @@ do
 
         case OP_PROP:
         if (ncode[1] != PT_CLIST) break;
-        /* Fall through */
+        __attribute__((fallthrough));
         case OP_ANYNL:
         case OP_CHAR:
         case OP_CHARI:
@@ -1327,7 +1327,7 @@ do
         tcode = ncode;
         continue;   /* With the following significant opcode */
         }
-      /* Fall through */
+      __attribute__((fallthrough));
 
       /* For a group bracket or a positive assertion without an immediately
       following mandatory setting, recurse to set bits from within the
@@ -1455,7 +1455,7 @@ do
 
       case OP_EXACT:
       tcode += IMM2_SIZE;
-      /* Fall through */
+      __attribute__((fallthrough));
       case OP_CHAR:
       case OP_PLUS:
       case OP_MINPLUS:
@@ -1466,7 +1466,7 @@ do
 
       case OP_EXACTI:
       tcode += IMM2_SIZE;
-      /* Fall through */
+      __attribute__((fallthrough));
       case OP_CHARI:
       case OP_PLUSI:
       case OP_MINPLUSI:
@@ -1602,7 +1602,7 @@ do
       case OP_TYPEUPTO:
       case OP_TYPEMINUPTO:
       case OP_TYPEPOSUPTO:
-      tcode += IMM2_SIZE;  /* Fall through */
+      tcode += IMM2_SIZE;  __attribute__((fallthrough));
 
       case OP_TYPESTAR:
       case OP_TYPEMINSTAR:
@@ -1783,7 +1783,7 @@ do
       /* It seems that the fall through comment must be outside the #ifdef if
       it is to avoid the gcc compiler warning. */
 
-      /* Fall through */
+      __attribute__((fallthrough));
 
       /* Enter here for a negative non-XCLASS. In the 8-bit library, if we are
       in UTF mode, any byte with a value >= 0xc4 is a potentially valid starter
@@ -1801,7 +1801,7 @@ do
 #elif PCRE2_CODE_UNIT_WIDTH != 8
       SET_BIT(0xFF);                             /* For characters >= 255 */
 #endif
-      /* Fall through */
+      __attribute__((fallthrough));
 
       /* Enter here for a positive non-XCLASS. If we have fallen through from
       an XCLASS, classmap will already be set; just advance the code pointer.
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -1936,7 +1936,7 @@ else
 
     if (c >= CHAR_8) break;
 
-    /* Fall through */
+    __attribute__((fallthrough));
 
     /* \0 always starts an octal number, but we may drop through to here with a
     larger first octal digit. The original code used just to take the least
@@ -2908,21 +2908,21 @@ switch(escape)
   {
   case ESC_D:
   prop = ESC_P;
-  /* Fall through */
+  __attribute__((fallthrough));
   case ESC_d:
   ascii_option = PCRE2_EXTRA_ASCII_BSD;
   break;
 
   case ESC_S:
   prop = ESC_P;
-  /* Fall through */
+  __attribute__((fallthrough));
   case ESC_s:
   ascii_option = PCRE2_EXTRA_ASCII_BSS;
   break;
 
   case ESC_W:
   prop = ESC_P;
-  /* Fall through */
+  __attribute__((fallthrough));
   case ESC_w:
   ascii_option = PCRE2_EXTRA_ASCII_BSW;
   break;
@@ -4560,7 +4560,7 @@ while (ptr < ptrend)
 
           default:
           PCRE2_DEBUG_UNREACHABLE();
-          /* Fall through */
+          __attribute__((fallthrough));
 
           case ESC_A:
           case ESC_Z:
@@ -5224,7 +5224,7 @@ while (ptr < ptrend)
         ++ptr;
         goto FAILED_FORWARD;
         }
-      /* Fall through */
+      __attribute__((fallthrough));
 
       case CHAR_0: case CHAR_1: case CHAR_2: case CHAR_3: case CHAR_4:
       case CHAR_5: case CHAR_6: case CHAR_7: case CHAR_8: case CHAR_9:
@@ -5974,7 +5974,7 @@ for (;;)
     case OP_UCP_WORD_BOUNDARY:
     case OP_NOT_UCP_WORD_BOUNDARY:
     if (!skipassert) return code;
-    /* Fall through */
+    __attribute__((fallthrough));
 
     case OP_CALLOUT:
     case OP_CREF:
@@ -6493,7 +6493,7 @@ for (;; pptr++)
     case META_PRUNE:
     case META_SKIP:
     cb->had_pruneorskip = TRUE;
-    /* Fall through */
+    __attribute__((fallthrough));
     case META_COMMIT:
     case META_FAIL:
     *code++ = verbops[(meta - META_MARK) >> 16];
@@ -6518,7 +6518,7 @@ for (;; pptr++)
     case META_PRUNE_ARG:
     case META_SKIP_ARG:
     cb->had_pruneorskip = TRUE;
-    /* Fall through */
+    __attribute__((fallthrough));
     case META_MARK:
     case META_COMMIT_ARG:
     VERB_ARG:
@@ -7498,7 +7498,7 @@ for (;; pptr++)
       group_return = -1;  /* Set "may match empty string" */
 
       /* Now treat as a repeated OP_BRA. */
-      /* Fall through */
+      __attribute__((fallthrough));
 
       /* If previous was a bracket group, we may have to replicate it in
       certain cases. Note that at this point we can encounter only the "basic"
@@ -8348,7 +8348,7 @@ for (;; pptr++)
       if ((options & PCRE2_UCP) != 0 && (xoptions & PCRE2_EXTRA_ASCII_BSW) == 0)
         meta_arg = (meta_arg == ESC_B)? OP_NOT_UCP_WORD_BOUNDARY :
           OP_UCP_WORD_BOUNDARY;
-      /* Fall through */
+      __attribute__((fallthrough));
 
       case ESC_A:
       if (cb->max_lookbehind == 0) cb->max_lookbehind = 1;
@@ -9282,7 +9282,7 @@ do {
 
      case OP_EXACT:
      scode += IMM2_SIZE;
-     /* Fall through */
+     __attribute__((fallthrough));
 
      case OP_CHAR:
      case OP_PLUS:
@@ -9295,7 +9295,7 @@ do {
 
      case OP_EXACTI:
      scode += IMM2_SIZE;
-     /* Fall through */
+     __attribute__((fallthrough));
 
      case OP_CHARI:
      case OP_PLUSI:
@@ -9727,7 +9727,7 @@ for (;; pptr++)
     case META_BACKREF_BYNAME:
     if ((cb->external_options & PCRE2_MATCH_UNSET_BACKREF) != 0)
       goto ISNOTFIXED;
-    /* Fall through */
+    __attribute__((fallthrough));
 
     case META_RECURSE_BYNAME:
       {
@@ -9776,7 +9776,7 @@ for (;; pptr++)
       goto RECURSE_OR_BACKREF_LENGTH;
       }
 
-    /* Fall through */
+    __attribute__((fallthrough));
     /* For groups >= 10 - picking up group twice does no harm. */
 
     /* A true recursion implies not fixed length, but a subroutine call may
@@ -9856,7 +9856,7 @@ for (;; pptr++)
 
     case META_CAPTURE:
     group = META_DATA(*pptr);
-    /* Fall through */
+    __attribute__((fallthrough));
 
     case META_ATOMIC:
     case META_NOCAPTURE:
@@ -9903,7 +9903,7 @@ for (;; pptr++)
         else itemlength = (max - 1) * lastitemlength;
       break;
       }
-    /* Fall through */
+    __attribute__((fallthrough));
 
     /* Any other item means this branch does not have a fixed length. */
 
--- a/src/pcre2_compile_class.c
+++ b/src/pcre2_compile_class.c
@@ -65,7 +65,7 @@ b) none of the cases here:
 #define CLASS_END_CASES(meta) \
   default: \
   PCRE2_ASSERT((meta) <= META_END); \
-  /* Fall through */ \
+  __attribute__((fallthrough)); \
   case META_CLASS: \
   case META_CLASS_NOT: \
   case META_CLASS_EMPTY: \
@@ -2169,7 +2169,7 @@ switch (meta)
     }
 
   ptr++;
-  /* Fall through */
+  __attribute__((fallthrough));
 
   default:
   /* Scan forward characters, ranges, and properties.
--- a/src/pcre2_convert.c
+++ b/src/pcre2_convert.c
@@ -226,7 +226,7 @@ while (plength > 0)
           posix++;
           continue;    /* With next character after :] */
           }
-        /* Fall through */
+        __attribute__((fallthrough));
 
         case POSIX_CLASS_NOT_STARTED:
         if (c == CHAR_LEFT_SQUARE_BRACKET)
@@ -321,7 +321,7 @@ while (plength > 0)
 
     case CHAR_LEFT_PARENTHESIS:
     bracount++;
-    /* Fall through */
+    __attribute__((fallthrough));
 
     case CHAR_QUESTION_MARK:
     case CHAR_PLUS:
@@ -329,7 +329,7 @@ while (plength > 0)
     case CHAR_RIGHT_CURLY_BRACKET:
     case CHAR_VERTICAL_LINE:
     if (!extended) goto ESCAPE_LITERAL;
-    /* Fall through */
+    __attribute__((fallthrough));
 
     case CHAR_DOT:
     case CHAR_DOLLAR_SIGN:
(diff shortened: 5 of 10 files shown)


NWilson:
We'll need to do something for portability here (since many major compilers don't have `__attribute__((fallthrough))`, eg MSVC).

A bit nasty, hmm.

But, I welcome the suggestion to allow this warning to be enabled on Clang. According to [this LLVM issue](https://github.com/llvm/llvm-project/issues/58693) and [this Flex issue](https://github.com/westes/flex/issues/427) (which Clang devs commented on), Clang simply isn't going to allow suppression of fallthrough warnings via a comment, which (in my opinion) is rather uncompromising and unhelpful, given decades of industry-wide use of `/* fallthrough */` comments.

Our CI builds _do_ enable `-Wimplicit-fallthough` on GCC, and the build does succeed due to GCC's support for suppression comments.

I _think_ what we want is:

* In `pcre2_util.h`, please add:

```c
/* We define this fallthrough macro for suppressing -Wimplicit-fallthrough warnings.
Clang only allows this via an attribute, whereas other compilers (eg. GCC) match attributes
and also specially-formatted comments.

For maximum portability, please use this macro as:

  PCRE2_FALLTHROUGH /* Fall through */
#if defined(__has_attribute)
#if __has_attribute(fallthrough)
#define PCRE2_FALLTHROUGH __attribute__((fallthrough));
#endif
#endif
#ifndef PCRE2_FALLTHROUGH
#define PCRE2_FALLTHROUGH
#endif
```

* Use that syntax throughout, to work with GCC and Clang and any other compilers (there are dozens that people use with PCRE2...)

* Two of the YAML files in `.github/` have a definition for `CFLAGS_GCC_STYLE`. We should update it to add `-Wimplicit-fallthrough`. It's included in `-Wextra` on GCC, but not on Clang.

Thank you!

Gsus42:
Ah sorry, I was too quick changing stuff. I've introduced a similar check as for HAVE_ATTRIBUTE_UNINITIALIZED. Would that be okay as well?

NWilson:
Well done! Sorry we were working simultaneously there.

Users often pull files from PCRE2 into their own build system, so I try and keep use of Autoconf and CMake minimal. We probably have too many configure-time checks already.

The changes you made were not wrong... But I think I slightly prefer the approach I suggested!

Using `__has_attribute` will work with more build systems. I think every compiler with `__attribute__((fallthrough))` will also have `__has_attribute` (it's only GCC and LLVM), so coverage should equally good.

(Interestingly, our GCC build on CI shows `checking for __attribute__((fallthrough))... no` and `Performing Test HAVE_ATTRIBUTE_FALLTHROUGH - Failed` with your change.)

Gsus42:
Okay, then I'd revert that last change and go with your approach.

NWilson:
Thank you for your understanding!

In my comment, I wrote `-Wno-implicit-fallthrough` in my suggested code snipped, but I've realised it's actually `-Wimplicit-fallthrough`. Apologies

Gsus42:
> Thank you for your understanding!
> 
> In my comment, I wrote `-Wno-implicit-fallthrough` in my suggested code snipped, but I've realised it's actually `-Wimplicit-fallthrough`. Apologies

I also had to change the `/* Fall through */` to `// Fall through //` to avoid:
src/pcre2_util.h:136:21: error: '/*' within block comment [-Werror,-Wcomment]

src/pcre2_study.c
```diff
@@ -1602,7 +1602,7 @@ do
       case OP_TYPEUPTO:
       case OP_TYPEMINUPTO:
       case OP_TYPEPOSUPTO:
-      tcode += IMM2_SIZE;  /* Fall through */
+      tcode += IMM2_SIZE;  PCRE2_FALLTHROUGH;
```
carenas: move the PCRE2_FALLTHROUGH down to the next line and don't forget to include a comment next to it.

src/pcre2_util.h
```diff
@@ -127,6 +127,23 @@ the reason and the actions that should be taken if it ever triggers. */
 #define PCRE2_ASSERT(x) do {} while(0)
 #endif
 
+/* We define this fallthrough macro for suppressing -Wimplicit-fallthrough warnings.
+Clang only allows this via an attribute, whereas other compilers (eg. GCC) match attributes
+and also specially-formatted comments.
+
+For maximum portability, please use this macro as:
```
carenas: "use this macro" TOGETHER with a "fall through" comment.

and make sure to follow your own advice.

src/pcre2_util.h
```diff
@@ -127,6 +127,23 @@ the reason and the actions that should be taken if it ever triggers. */
...
+
+  PCRE2_FALLTHROUGH // Fall through // */
+
+#if defined(__has_attribute)
+#if __has_attribute(fallthrough)
+#define PCRE2_FALLTHROUGH __attribute__((fallthrough));
```
carenas, line 159: no semicolon at the end.

It might be also better to make it a "function" like macro:

`PCRE2_FALLTHROUGH()`

src/pcre2_util.h
```diff
@@ -127,6 +127,23 @@ the reason and the actions that should be taken if it ever triggers. */
...
+#endif
+#endif
+#ifndef PCRE2_FALLTHROUGH
+#define PCRE2_FALLTHROUGH
```
carenas, line 166: IMHO we could do better here as C23 allows a standard attribute:

https://en.cppreference.com/w/c/language/attributes/fallthrough

obviously, you need to have a "nothing" fallback anyway, as our code is only C99 compatible.

src/pcre2_util.h
Gsus42: Should be fixed by last commit.

src/pcre2_util.h
Gsus42: I have left that in my last commit as a non-function like macro to let it look like `[[fallthrough]];` in effect. If that is not desired I'm happy to change that of course.

src/pcre2_util.h
Gsus42: At this point I'm not sure how to detect whether the compiler supports `[[fallthrough]]` or just `__attribute__((fallthrough))`. I haven't come up with some good check there yet...

src/pcre2_study.c
Gsus42: Should also be fixed by last commit.

src/pcre2_util.h
carenas: that is what the `__STDC_VERSION__` variable is for.

you need to check for C23 support, which means something like:

```
#if 202311L - __STDC_VERSION__ == 0
#define PCRE2_FALLTHROUGH [[ __fallthrough__ ]]
#endif
```

Gsus42: Of course there is `__has_cpp_attribute(fallthrough)` (see https://en.cppreference.com/w/cpp/feature_test.html) but that won't fit here I suppose.

NWilson: It think it would be:

```c
#if defined(__has_c_attribute)
#if __has_c_attribute(fallthrough)
#define PCRE2_FALLTHROUGH [[fallthrough]]
#endif
#endif
#if defined(__has_attribute) && !defined(PCRE2_FALLTHROUGH)
#if __has_attribute(fallthrough)
#define PCRE2_FALLTHROUGH __attribute__((fallthrough))
#endif
#endif
#ifndef PCRE2_FALLTHROUGH
#define PCRE2_FALLTHROUGH
#endif
```

However I don't see that this is really necessary. Maybe nice as a bit of future-proofing?

src/pcre2_auto_possess.c
```diff
@@ -804,21 +804,21 @@ for(;;)
 
       case OP_NOT_DIGIT:
       invert_bits = TRUE;
-      /* Fall through */
+      PCRE2_FALLTHROUGH; // Fall through
```
carenas: we use C comments in this codebase, just like the one that was replaced.

src/pcre2_util.h
Gsus42: Too me that makes it a bit overly complicated... but I'd change that if you'd prefer that.

src/pcre2_util.h
NWilson: I had kind-of made it deliberate in my request that it would be defined as `#define PCRE2_FALLTHROUGH __attribute__((fallthrough));` (with semicolon), and that it would be used without a semi-colon, so that we don't generate useless statements in builds not supporting this.

That seems reasonable to me. Although, inserting a useless do-while isn't exactly harmful either. It just seemed unnecessary to me.

src/pcre2_util.h
NWilson: We generally prefer `/* */` comments, although I think a few of the "new" (haha!) `//` ones have crept into the codebase.

It's a shame that it's awkward to put that in the doc string with that `-Wcomment` warning. We'll just have to format it like this or something:

```c
/* ... Use this macro as (deleting the underscore characters):

  PCRE2_FALLTHROUGH /__* Fall through *__/
*/
```

src/pcre2_util.h
```diff
@@ -127,6 +127,23 @@ the reason and the actions that should be taken if it ever triggers. */
 #define PCRE2_ASSERT(x) do {} while(0)
 #endif
 
+/* We define this fallthrough macro for suppressing -Wimplicit-fallthrough warnings.
+Clang only allows this via an attribute, whereas other compilers (eg. GCC) match attributes
+and also specially-formatted comments.
+
+For maximum portability, please use this macro together with a fall through comment:
+
+  PCRE2_FALLTHROUGH; // Fall through */
```
carenas: maybe something like:

```
... comment, for example: */

//  PCRE2_FALLTHROUGH; /* Fall through */
```

src/pcre2_auto_possess.c
Gsus42: Understood. I've changed that accordingly. The only thing that is a bit inconsistent is how it is written in the comment (as hinted in my last commit).

src/pcre2_util.h
Gsus42: Personally I'm happy with both. I adhere to whatever kind of style you'd like to see in PCRE2.

carenas (changes requested):
also, don't forget the CI changes so that the warning is actually enabled

src/pcre2_util.h
Gsus42: Ah good hint. Will adopt this change.

src/pcre2_util.h
carenas: Until recently, this codebase was C89 compatible, and // is not allowed in that standard.

C99 allows //, but we are meant to only use them as temporary placeholders, like "// TODO: fix this before releasing", which then are easy to spot.

src/pcre2_util.h
NWilson: Hmm. Just been thinking. You know what? Standards are good (C23 etc). Let's add the `[[fallthrough]]` version too, for future-proofing. Then we won't need to revisit this for a long time!

src/pcre2_util.h
NWilson: I don't mind - my formatting (in other comment) or Carlo's is OK. You get to choose!

src/pcre2_util.h
NWilson: You've done the semicolons at point-of-use. There's no functional impact. I won't ask you to make another change. It doesn't matter, we'll just roll with it.

I do apologise this turned into 8 revisions (!!) of bike-shedding for a small code change. (And two reviewers didn't help reduce that.) Many thanks!

carenas: You original code had BOTH a semicolon in the macro definition and when used, which means we had empty commands added everywhere.

My suggestion was to remove the `;` in the macro definition, but could have done the simpler thing and remove the semicolon in the places where it is used as well, but you can't have BOTH.

NWilson: My original comment said:

```c
/* ...Use this as:

  PCRE2_FALLTHROUGH /* Fall through */

...
#define PCRE2_FALLTHROUGH __attribute__((fallthrough));
```

But who cares! It's all fine!

Gsus42: Okay... is there anything else I've missed?

src/pcre2_util.h
carenas: > Maybe nice as a bit of future-proofing?

actually gcc-15  defaults to C23 already, so it isn't that far in the future.

I would also prefer using `__fallthrough__` instead so there is no possibility of an accidental `#define fallthrough` somewhere messing it up.

NWilson (approved):

Gsus42:
> Oh wait, actually the JIT build needs fixing :(

Mh, I wonder why we get:

```
src/pcre2_util.h:140:27: error: ISO C does not support '[[]]' attributes before C23 [-Werror=pedantic]
  140 | #define PCRE2_FALLTHROUGH [[fallthrough]]
```

as I thought `__has_c_attribute` would check for that (see https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fc_005fattribute.html). Mh?

NWilson:
> > Oh wait, actually the JIT build needs fixing :(
> 
> Mh, I wonder why we get:
> 
> ```
> src/pcre2_util.h:140:27: error: ISO C does not support '[[]]' attributes before C23 [-Werror=pedantic]
>   140 | #define PCRE2_FALLTHROUGH [[fallthrough]]
> ```
> 
> as I thought `__has_c_attribute` would check for that (see https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fc_005fattribute.html). Mh?

Apparently not. Perhaps:

```c
#if defined(__has_c_attribute)
#if __has_c_attribute(fallthrough) && __STDC_VERSION__ >= 202311L
#define PCRE2_FALLTHROUGH [[fallthrough]]
#endif
#endif
```

It's apparently an error to use the `[[fallthrough]]` syntax if the compiler isn't using `--std=c23`. Fussy, but fair enough.

Gsus42:
Ah, for the clang compile we get:

```
src/pcre2grep.c:2058:3: error: [[]] attributes are a C23 extension [-Werror,-Wc23-extensions]
 2058 |   PCRE2_FALLTHROUGH; /* Fall through */
``` 

I guess we cannot have `-Wc23-extensions` as `[[fallthrough]]` is a C23 extension. 
Can this be modified somehow, or should I rather stick with the original approach?

carenas:
> Ah, for the clang compile we get:
> 
> ```
> src/pcre2grep.c:2058:3: error: [[]] attributes are a C23 extension [-Werror,-Wc23-extensions]
>  2058 |   PCRE2_FALLTHROUGH; /* Fall through */
> ```
> 
> I guess we cannot have `-Wc23-extensions` as [[fallthrough]] is a C23 extension. Can this be modified somehow, or should I rather stick with the original approach?

yes you can, but need to check for the C version that is being used first, see my comment on __STDC_VERSION__

Gsus42:
Okay, I've extended the check accordingly. I also replaced `[[fallthrough]]` by `[[__fallthrough__]]` as hinted above.

src/pcre2_util.h
```diff
@@ -127,6 +127,28 @@ the reason and the actions that should be taken if it ever triggers. */
...
+//  PCRE2_FALLTHROUGH; /* Fall through */
+
+#if defined(__has_c_attribute)
+#if __has_c_attribute(fallthrough) && __STDC_VERSION__ >= 202311L
+#define PCRE2_FALLTHROUGH [[__fallthrough__]]
+#endif
+#endif
+#if defined(__has_attribute) && !defined(PCRE2_FALLTHROUGH)
+#if __has_attribute(fallthrough)
+#define PCRE2_FALLTHROUGH __attribute__((fallthrough))
```
carenas: you can also use the underscored version of fallthrough here AFAIK

Gsus42:
Ah right, I forgot about the JIT support... this requires a few additional changes. Working on it...

src/pcre2_util.h
```diff
@@ -127,6 +127,28 @@ the reason and the actions that should be taken if it ever triggers. */
...
 
+/* We define this fallthrough macro for suppressing -Wimplicit-fallthrough warnings.
+Clang only allows this via an attribute, whereas other compilers (eg. GCC) match attributes
+and also specially-formatted comments.
+
+For maximum portability, please use this macro together with a fall through C comment, for example: */
+
+//  PCRE2_FALLTHROUGH; /* Fall through */
+
+#if defined(__has_c_attribute)
```
carenas: the order here seems wrong, IMHO we should check for the C standard first, and while we required an ANSI C compiler with C99 support, some really old ones didn't define this macro, so it is usually better to check with a substraction (an undefined macro == 0)

NWilson:
But it won't be as easy as you hoped. Because one of the source files is in another repo! 😢 

I guess we're going to have to make a (very similar) PR into `zherczeg/sljit` to do the same whole business over there.

carenas (commented):
only a few nitpicks, assume this is being tested other than the CI?

Gsus42:
> But it won't be as easy as you hoped. Because one of the source files is in another repo! 😢
> 
> I guess we're going to have to make a (very similar) PR into `zherczeg/sljit` to do the same whole business over there.

Indeed... I hope this finds similar acceptance.

NWilson:
Zoltan will take it (I think). It's a pat of PCRE2 that's been split out into a re-usable project. You can define SLJIT_FALLTHROUGH next to SLJIT_ASSERT, same as you defined PCRE2_FALLTHROUGH in the same file as PCRE2_ASSERT.

src/pcre2_util.h
NWilson: Not-underscored is canonical. But the standard pretty-much guarantees that `__fallback__` must work too for back-compat. Don't care, they're effectively the same.

src/pcre2_util.h
NWilson: The order of checks doesn't matter.

And, if `__STDC_VERSION__` is not defined, the check evaluates to false, so this is a perfectly good way of doing the C23 test.

src/pcre2_util.h
carenas: again, it is just to protect against a `#define fallthrough` somewhere

src/pcre2_util.h
Gsus42: I guess, hereby you mean by checking whether `__STDC_VERSION__` is defined at all?
Would it be okay to implement this as follows:

```
#if defined(__STDC_VERSION__) && __STDC_VERSION__ >= 202311L
#if defined(__has_c_attribute)
#if __has_c_attribute(__fallthrough__)
#define PCRE2_FALLTHROUGH [[__fallthrough__]]
#endif
#endif
#endif
```

I guess this might work for future C-standards following C23, hence the >= check, wouldn't it?

Gsus42:
Are the recent changes okay for you? If so I'd then do a similar pull request for the SLJIT repository.

src/pcre2_util.h
carenas: practically, it doesn't make a difference as Nicholas pointed out, so I guess it is OK if the original is simpler to read, even if it feels backwards, but at least it is clear on the intent: "only use `[[__fallthrough__]]` if supported AND using C23"

NWilson:
> Are the recent changes okay for you? If so I'd then do a similar pull request for the SLJIT repository.

Yes. I won't randomise you on any cosmetic changes - it's perfectly fine as-is.

I'd be grateful if you did do the change in sljit.

Gsus42:
> > Are the recent changes okay for you? If so I'd then do a similar pull request for the SLJIT repository.
> 
> Yes. I won't randomise you on any cosmetic changes - it's perfectly fine as-is.
> 
> I'd be grateful if you did do the change in sljit.

The pull request has been placed:

https://github.com/zherczeg/sljit/pull/329

--

One thing I still need to investigate is the failure for the `manyconfig` run:
```
In file included from src/pcre2_internal.h:2393,
                 from src/pcre2_study.c:46:
src/pcre2_study.c: In function 'set_start_bits':
src/pcre2_util.h:147:27: warning: a label can only be part of a statement and a declaration is not a statement [-Wpedantic]
  147 | #define PCRE2_FALLTHROUGH __attribute__((__fallthrough__))
       |                           ^~~~~~~~~~~~~
src/pcre2_study.c:1805:7: note: in expansion of macro 'PCRE2_FALLTHROUGH'
 1805 |       PCRE2_FALLTHROUGH; /* Fall through */
      |       ^~~~~~~~~~~~~~~~~
```

NWilson:
> One thing I still need to investigate is the failure for the `manyconfig` run:
> 
> ```
> In file included from src/pcre2_internal.h:2393,
>                  from src/pcre2_study.c:46:
> src/pcre2_study.c: In function 'set_start_bits':
> src/pcre2_util.h:147:27: warning: a label can only be part of a statement and a declaration is not a statement [-Wpedantic]
>   147 | #define PCRE2_FALLTHROUGH __attribute__((__fallthrough__))
>        |                           ^~~~~~~~~~~~~
> src/pcre2_study.c:1805:7: note: in expansion of macro 'PCRE2_FALLTHROUGH'
>  1805 |       PCRE2_FALLTHROUGH; /* Fall through */
>       |       ^~~~~~~~~~~~~~~~~
> ```

That's nasty. That's a very vexing parse!

The compiler sees:

```
case 1:
__attribute__((fallthrough));
case 2:
```

The problem is that the attribute attached to the "null statement" is not actually parsed as a statement unless there's a statement before it. (Yeah, C...) It's ambiguous - because it could be a declaration of a variable (which isn't a legal target for a case label). To disambiguate, it's treated as a null statement only if it follows another statement ("statement context").

Here's the solution...

```c
      case OP_NCLASS:
#if defined SUPPORT_UNICODE && PCRE2_CODE_UNIT_WIDTH == 8
      if (utf)
        {
        re->start_bitmap[24] |= 0xf0;            /* Bits for 0xc4 - 0xc8 */
        memset(re->start_bitmap+25, 0xff, 7);    /* Bits for 0xc9 - 0xff */
        }
#elif PCRE2_CODE_UNIT_WIDTH != 8
      SET_BIT(0xFF);                             /* For characters >= 255 */
#else
      ; /* Ensure case label interacts correctly with PCRE2_FALLTHROUGH
#endif
      PCRE2_FALLTHROUGH; /* Fall through */
```

carenas:
> Here's the solution...
> 
> ```c
> #else
>       ; /* Ensure case label interacts correctly with PCRE2_FALLTHROUGH */
> #endif
>       PCRE2_FALLTHROUGH; /* Fall through */
> ```

note, that it was missing a closing `*/` in the added "null statement", so it should look instead like the one above.

PhilipHazel:
I have been entertained by watching this saga roll by ... it just goes to show how complicated the world has become, with multiple compilers and multiple standards. I could never have imagined this when I first wrote PCRE. One comment for the record: the // comment is not new, not even "new". BCPL has had it from its birth in the 1960s, pre-dating C. I always wondered why // got dropped when the BCPL => B => C evolution happened. (Another thing they dropped was "endcase", overloading "break" instead, which is sometimes a right nuisance.)

NWilson:
You can partly blame C++. There's a lot of "enterprise" tooling in that area, which gradually trickles down to C as well, although the C folks try very hard to keep alive the bygone era of simple stuff.

Gsus42:
> > One thing I still need to investigate is the failure for the `manyconfig` run:
> > ```
> > In file included from src/pcre2_internal.h:2393,
> >                  from src/pcre2_study.c:46:
> > src/pcre2_study.c: In function 'set_start_bits':
> > src/pcre2_util.h:147:27: warning: a label can only be part of a statement and a declaration is not a statement [-Wpedantic]
> >   147 | #define PCRE2_FALLTHROUGH __attribute__((__fallthrough__))
> >        |                           ^~~~~~~~~~~~~
> > src/pcre2_study.c:1805:7: note: in expansion of macro 'PCRE2_FALLTHROUGH'
> >  1805 |       PCRE2_FALLTHROUGH; /* Fall through */
> >       |       ^~~~~~~~~~~~~~~~~
> > ```
> 
> That's nasty. That's a very vexing parse!
> 
> The compiler sees:
> 
> ```
> case 1:
> __attribute__((fallthrough));
> case 2:
> ```
> 
> The problem is that the attribute attached to the "null statement" is not actually parsed as a statement unless there's a statement before it. (Yeah, C...) It's ambiguous - because it could be a declaration of a variable (which isn't a legal target for a case label). To disambiguate, it's treated as a null statement only if it follows another statement ("statement context").
> 
> Here's the solution...
> 
> ```c
>       case OP_NCLASS:
> #if defined SUPPORT_UNICODE && PCRE2_CODE_UNIT_WIDTH == 8
>       if (utf)
>         {
>         re->start_bitmap[24] |= 0xf0;            /* Bits for 0xc4 - 0xc8 */
>         memset(re->start_bitmap+25, 0xff, 7);    /* Bits for 0xc9 - 0xff */
>         }
> #elif PCRE2_CODE_UNIT_WIDTH != 8
>       SET_BIT(0xFF);                             /* For characters >= 255 */
> #else
>       ; /* Ensure case label interacts correctly with PCRE2_FALLTHROUGH
> #endif
>       PCRE2_FALLTHROUGH; /* Fall through */
> ```

Mh, I see. Yet, here I'd rather move the fallthrough statement into the `#if`/`#elif` parts. Just adding an empty statement looks odd to me.

Gsus42:
I've closed my original pull request https://github.com/zherczeg/sljit/pull/329. @carenas  has refined that stuff in https://github.com/zherczeg/sljit/pull/330 which is required before this one can finally be merged (unless further changes are required).

Gsus42:
Given that https://github.com/zherczeg/sljit/pull/330 has been merged can you give the recent changes another try? Or do you expect further changes to be done?

NWilson:
Thank you for the reminder. I was away for a few days and didn't check in on this.

I will do an update of the sljit component (#791) and then we can merge this.

If there any tweaks I wish to make, I'll just push them to this branch and not trouble you with them.

NWilson:
Many thanks again @Gsus42. I don't know how this turned into such an epic thread. That's not the normal process for contributions in PCRE2! Thank you for your patience.

Gsus42:
Many thanks for having this change! Indeed this was much larger than I imagined :).

PR #791, opened 2025-09-08
Update sljit to current master

NWilson:
This is a routine update, so that we can enforce the `-Wimplicit-fallthrough` warning to our CI after https://github.com/zherczeg/sljit/pull/330.


NWilson:
Unfortunately the build fails. Attempt at fix over here: https://github.com/zherczeg/sljit/pull/331

zherczeg:
@carenas it looks like there is still more combinations... :(

carenas:
> @carenas it looks like there is still more combinations... :(

fixing `-Wundef` should be straight forward; relying too much in `__has*attribute()` might not be worth doing though, as we know clang has a broken implementation of `__has_attribute(fallthrough)` in older versions.

zherczeg:
Note: a feature development is in progress, so the current sljit does not compile everywhere. Another update will be needed.

NWilson:
Thanks! I'm happy for PCRE2 master to track sljit master, and for them both to be work-in-progress.

We could even do a regular (automated?) weekly merge. Or just manual - I don't mind.

PR #796, opened 2025-09-17
Improvements to CI code coverage

NWilson:
- Add maint/RunCoverage script to allow gathering Clang coverage locally more easily
- Add LCOV exclusion comments to unreachable code
- Add a filtering script so that the CI job respects LCOV_EXCL_* comments


zherczeg:
I thought this patch adds more test cases :)

NWilson:
It does add a very few extra lines of coverage - up by 0.5% apparently.

But there are definitely more PRs coming.

In the next PR we'll get to 100% function coverage - we still have some totally unhit functions, which aren't especially important but it would still be better to avoid that.

PR #797, opened 2025-09-18
De-macro the pcre2test code

NWilson:
I have been wanting to do this for a while.

The current style in pcre2test.c is rather hard to work with. Every little call to PCRE2 functions has to be guarded behind many layers of complicated macros like "SET", "CASTVAR", and more.

At the cost of a modest increase in binary size, I would like to split out roughly half of the logic in pcre2test.c into a separate file, which can be multiply-included.

This means that instead of having a single callsite to a function like `pcre2_substitute`, but via a macro PCRE2_SUBSTITUTE that dispatches each and every call using `test_mode`, we can instead have three copies of all the pattern-handling functions, each one making a simple, direct call into PCRE2.

Overall, we lose about 10% of pcre2test.c (1500 lines), which is all the macros like this:

```
#define PCRE2_JIT_MATCH(a,b,c,d,e,f,g,h) \
  if (test_mode == PCRE8_MODE) \
    a = pcre2_jit_match_8(G(b,8),(PCRE2_SPTR8)c,d,e,f,G(g,8),h); \
  else if (test_mode == PCRE16_MODE) \
    a = pcre2_jit_match_16(G(b,16),(PCRE2_SPTR16)c,d,e,f,G(g,16),h); \
  else \
    a = pcre2_jit_match_32(G(b,32),(PCRE2_SPTR32)c,d,e,f,G(g,32),h)
```

Gone!

But more importantly than that, the resulting code seems to me to be simpler to work with.


NWilson:
GitHub appears to be using git's default diff algorithm, rather than something better (like `git config --global diff.algorithm histogram`). This causes the PR to show as thousands of lines of green and red that don't line up, whereas the change is actually rather smaller than than ("only" 1000 lines changed, or about 90% of the lines from pcre2test.c kept without modification).

Looking at `git log -p --diff-algorithm=histogram -n1` in the terminal gives a much better view.

Additionally, this PR is based off a commit I haven't merged yet, which simply copies `pcre2test.c` to `pcre2test_inc.h` (without adding or deleting any lines at all). That's the only way to get git to show changes when splitting a file.

NWilson:
OK! I have a few "XXX" comments left in the code to address (some coverage-related code I want to move elsewhere), but overall I'm happy with this now.

It _looks_ major, like I said, but it's not as many lines of change as it looks from the standard Git diff.

PhilipHazel:
Sorry about the macro mess. You probably realize how it happened: the original PCRE was just 8-bit; when we added 16-bit support it seemed easiest just to macro it - but if I remember it right there were also fewer functions to support. Then it grew. And of course memory is less of an issue than it was 30 years ago.

NWilson:
That's what I had assumed. These things happen, and I just took the opportunity to carry on the good work. 

I'm really feeling quite at home now in PCRE2 code, and feeling like I know most of it fairly well now. I've got some plans for some cleanup, features, and optimisations. 

I hope you're keeping well Philip.

PhilipHazel:
Excellent news, glad to hear you have settled in. I am keeping well, thank you, tussling with MusicXML and getting things to work on MacOS (when I don't actually own a Mac :-)

NWilson:
I have measured the resulting size increase in the pcre2test binary (JIT+Unicode supported, 8+16+32 bits).

Obviously, testing with shared libraries - the code size change is a drop in the bucket if pcre2test is statically linking in PCRE2.

Size of pcre2test:

| Configuration | Old | New |
|---|---|---|
| `-Os`, dynamic | 199K | 242K (+21%) |
| `-O3`, dynamic | 239K | 300K (+25%) |
| `-O3`, static | 2.27M | 2.33M (+2.5%) |

For the substantial maintenance improvement, I think this is justifiable. Since when has anyone actively tried to have a smaller test suite!

PR #798, opened 2025-09-19
Fix three minor issues in pcre2test

NWilson:
- A dangerous STRLEN macro was casting only part of its argument, which lead to unexpected behaviour in one instance (luckily harmless)
- A "compiler warning suppression" from a previous version of the code, which is definitely now obsolete.
- A belt-and-braces check which attempted to add a null terminator, but which was only writing a single byte. Thankfully no code actually reads the terminator anyway.

### DIFF
--- a/src/pcre2test.c
+++ b/src/pcre2test.c
@@ -1168,6 +1168,11 @@ are supported. */
   (test_mode == PCRE16_MODE)? (uint32_t)(((PCRE2_SPTR16)(a))[b]) : \
   (uint32_t)(((PCRE2_SPTR32)(a))[b]))
 
+#define INCSTR(x,offset) ( \
+  (test_mode == PCRE8_MODE)? (void *)((PCRE2_SPTR8)(x)+offset) : \
+  (test_mode == PCRE16_MODE)? (void *)((PCRE2_SPTR16)(x)+offset) : \
+  (void *)((PCRE2_SPTR32)(x)+offset))
+
 #define CONCTXCPY(a,b) \
   if (test_mode == PCRE8_MODE) \
     memcpy(G(a,8),G(b,8),sizeof(pcre2_convert_context_8)); \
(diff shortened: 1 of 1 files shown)


src/pcre2test.c
```diff
@@ -5337,7 +5350,7 @@ if ((pat_patctl.control & CTL_INFO) != 0)
     for (; namecount > 0; namecount--)
       {
       int imm2_size = test_mode == PCRE8_MODE ? 2 : 1;
-      size_t length = STRLEN(nametable + imm2_size);
```
NWilson, line 5340: This was really surprising to me.

`nametable + imm2_size` is really not valid, because `nametable` has type `void*`. And yet, due to the macro, it was fortuitously expanding out to `strlen32((PCRE2_SPTR32)nametable + imm2_size)`, which luckily applies the cast only to the first half of the expression.

The STRLEN macro is broken, but it just happens to luck out on this case.

The whole system macros here is really, really hard to get right. I'm glad I'm getting rid of it in my other PR.

src/pcre2test.c
```diff
@@ -9109,9 +9122,6 @@ for (gmatched = 0;; gmatched++)
 
   if ((dat_datctl.control & (CTL_FINDLIMITS|CTL_FINDLIMITS_NOHEAP)) != 0)
     {
-    capcount = 0;  /* This stops compiler warnings */
```
NWilson, line 9112: I've looked back to the code in 2017 when Philip added this.

In that version, `capcount` was assigned conditionally, so it would indeed have generated a warning. The suppression makes no sense anymore though, because we assign to `capcount` unconditionally just a line or two below. This is a "dead store", a value we put in the variable and never read back.

src/pcre2test.c
```diff
@@ -9250,20 +9260,20 @@ for (gmatched = 0;; gmatched++)
...
     if (pp == NULL)
       {
 #ifdef SUPPORT_VALGRIND
       /* Mark the start of dbuffer addressable again. */
-      VALGRIND_MAKE_MEM_UNDEFINED(dbuffer, 1);
+      VALGRIND_MAKE_MEM_UNDEFINED(dbuffer, code_unit_size);
 #endif
       pp = dbuffer;
-      pp[0] = 0;
+      memset(pp, 0, code_unit_size);
```
NWilson, line 9276: Another lucky bug. We are writing only a single byte here, not a full code unit (on 16-bit and 32-bit libraries). However, the whole block of code is just belt-and-braces stuff. Even if you remove it totally (or write `0xDEADBEEF` instead of null) the tests all pass absolutely fine, even under valgrind.

Basically we don't need the trailing null, so it's harmless that we didn't write it correctly.

PR #799, opened 2025-09-22
Chip away at more coverage gaps

NWilson:


src/pcre2test_inc.h
```diff
@@ -5474,21 +5648,104 @@ if (rc != PCRE2_ERROR_BADOPTION) failure = "pcre2_pattern_info(bad option)";
...
+if (test_con_context_copy != NULL) pcre2_convert_context_free(test_con_context_copy);
+if (test_dat_context_copy != NULL) pcre2_match_context_free(test_dat_context_copy);
+if (test_pat_context_copy != NULL) pcre2_compile_context_free(test_pat_context_copy);
+if (test_gen_context_copy != NULL) pcre2_general_context_free(test_gen_context_copy);
+if (test_con_context != NULL) pcre2_convert_context_free(test_con_context);
+if (test_dat_context != NULL) pcre2_match_context_free(test_dat_context);
+if (test_pat_context != NULL) pcre2_compile_context_free(test_pat_context);
+if (test_gen_context != NULL) pcre2_general_context_free(test_gen_context);
+
 if (failure != NULL)
```
zherczeg, line 6010: It looks like a single failure is written. Could we improve this to print all errors? Maybe a `prinf` instead of a local variable?

NWilson: These are tests which really, really shouldn't fail. I guess I could add a macro so that they all just assert & abort, rather than flow down to the end. That would at least mean that the tests all fail immediately, rather than cheerfully carrying on.

PR #801, opened 2025-09-23
Add fix and tests for callback enumerate

### LINKED ISSUE #800: pcre2_callout_enumerate fails for patterns with Unicode character classes
This is a bug in PCRE2 10.45 and 10.46, not present in 10.44 and earlier.

Applications which call pcre2_callout_enumerate should potentially downgrade to PCRE2 10.44 until we release an update of PCRE2 with the fix.

Since this function is (clearly!!) not used often by applications, I am not currently treating this as very high severity. There is no way for an attacker to make any application call this function, if it is not currently using it.

The root cause seems to be commit 24f9d8df0b8ddabc217ec4e7856a678e09f52773 (#540).

(I apologise for pointing to the specific commit - I am not blaming anyone. I just think we need to be sure we understand when the bug was introduced.)

The layout of the `pcre2_code` object was altered from `struct, nametable, code` to `struct, nametable, classlists, code`. Nearly all the relevant code was updated, but clearly, pcre2_callout_enumerate was missed out.

We also are guilty of having inadequate code coverage of this function. We should be confident to make changes like this and know that our test suite will catch any errors. Sadly, the test suite only covers a few happy paths through pcre2_callout_enumerate, on a handful of simple patterns, and none of them have complicated character classes.

---

Please comment if you want to express an opinion on how we handle this bug (whether to do a rapid release, or to hurry up our scheduled 10.47, or to just fix it whenever 10.47 comes around, with no extra hurry).

NWilson:
Fixes #800


src/pcre2test_inc.h
```diff
@@ -83,6 +83,7 @@ library. */
 #define print_error_message_file          PCRE2_SUFFIX(print_error_message_file_)
 #define print_error_message               PCRE2_SUFFIX(print_error_message_)
 #define callout_enumerate_function        PCRE2_SUFFIX(callout_enumerate_function_)
+#define callout_enumerate_function_void   PCRE2_SUFFIX(callout_enumerate_function_void_)
```
carenas, line 87: no need for this `#define`

src/pcre2test_inc.h
```diff
@@ -5526,6 +5529,7 @@ if (failure != NULL)
 #undef print_error_message_file
 #undef print_error_message
 #undef callout_enumerate_function
+#undef callout_enumerate_function_void
```
carenas, line 6053: ditto

src/pcre2test_inc.h
NWilson: The build would definitely fail without this define. All the functions in `pcre2test_inc.h` require the PCRE2_SUFFIX in order not to conflict with each other.

carenas: ... but there is no need to have up to 3 copies of this callback, is there?

NWilson: The callout function has a different signature on each bitwidth.

Technically I suppose we could just have one function type as `int (void*,void*)` but it would need a cast to be usable.

I have preferred to avoid casts and just give everything in `pcre2test_inc.h` its proper correct signature.

NWilson:
Thank you, those changes look helpful, I'll take them.

src/pcre2_pattern_info.c
```diff
@@ -289,8 +289,7 @@ if (re->magic_number != MAGIC_NUMBER) return PCRE2_ERROR_BADMAGIC;
 if ((re->flags & (PCRE2_CODE_UNIT_WIDTH/8)) == 0) return PCRE2_ERROR_BADMODE;
 
 cb.version = 0;
-cc = (PCRE2_SPTR)((const uint8_t *)re + sizeof(pcre2_real_code))
-     + re->name_count * re->name_entry_size;
+cc = (PCRE2_SPTR)((uint8_t *)re + re->code_start);
```
NWilson, line 292: Zoltan, I tagged you on this review just for this one line, really.

I'm sure the code itself is correct, but I guess I wanted to give you a chance to say whether you're happy or not with releasing the fix in this way, waiting until the next release (basically: it's a nasty crash, but we don't need to do a security advisory, because no-one uses this function).

zherczeg: I am happy with it. We don't need another security release.

NWilson: Thank you!

PR #802, opened 2025-09-26
JIT compiler update

zherczeg:


zherczeg:
Yes. I don't want to do bigger changes before the next release.

NWilson:
Is this the final sljit update? Just let me know when you are release-ready.

zherczeg:
This is a good point for release. I don't plan more updates unless a fix is needed.

carenas:
> don't plan more updates unless a fix is needed

I think there is still a missing update (to avoid clashing SIMD definitions) for LoongArch that is pending though.

PR #805, opened 2025-09-26
Remove memmove, bcopy, and VPCOMPAT macros

NWilson:
The `memmove` function dates back to C89, and has been available in all systems since 1993 (when SunOS 5 and BSD 4.4 were released).

The Visual Pascal compiler had its last release in 2003 and its original author posted in 2005 that "Virtual Pascal has died". Those macros can also be removed now.

The only reason I want to be proactive about this is because many users incorporate the source files from PCRE2 into their own build system. Some of these users are undoubtedly getting the memmove fallback code, if they are not providing the HAVE_MEMMOVE macro. Our minimum supported C version is now C99, so there is no need for pre-C89 fallbacks.

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -22,7 +22,6 @@ copy_file(
 
 LOCAL_DEFINES = [
     "HAVE_CONFIG_H",
-    "HAVE_MEMMOVE",
     "HAVE_STRERROR",
     "PCRE2_STATIC",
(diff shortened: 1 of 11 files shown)


carenas (approved):
build.zig also needs updating though

NWilson:
Oops! Thank you. I missed that because my editor is usually set to exclude `build*` from searches (to hide noise from build directories).

PR #810, opened 2025-10-03
Avoid zero-length allocations, which return NULL on z/OS

NWilson:
It's totally legal to return NULL for malloc(0), but every other platform that we target returns a non-NULL pointer. We don't need to do an allocation anyway (nor do we need to copy the terminating NUL byte).

This PR fixes a CI build failure on the z/OS runner.


NWilson:
I did do a quick check of every other call to malloc/calloc/realloc in our codebase. Every single other one appears to allocate a size which is definitely >0.

carenas (approved):

PR #811, opened 2025-10-04
Print `pcre2test` output in colour.

IsaacOscar:
This adds a `-c` option to `pcre2test` that will print various different output in colour.

For ease of reviewing I've split the changes into two commits:
1. contains the interesting changes, but it doesn't compile without the second commit.
2. simply changes all remaining `printf`, `pchars`, etc. calls to also include the colour to use.

Most output from `pcre2test` is given an explicit colour, however this is often just `clr_output` which is the terminals default colour (i.e. it likely won't be in "colour").

### DIFF
--- a/RunTest
+++ b/RunTest
@@ -198,7 +198,7 @@ checkresult()
 checkspecial()
   {
   expect=${2:-0}
-  $sim $valgrind $vjs $pcre2test $1 >>testSoutput
+  $sim $valgrind $vjs $pcre2test $1 >>testSoutput 2>&1
   if [ $? -ne "$expect" ] ; then
     echo "** pcre2test $1 failed - check testSoutput"
     yield=1


NWilson:
This is great! I have a few requests:
* I _think_ I prefer having the output `FILE*` be explicit, and only output to stdout if it's not specified. This matches the C convention - `printf` goes to stdout, and all the other methods (fprintf etc) have an explicit output destination. This basically means adding "outfile" as an argument to your cprintf throughout. Thoughts?
* I'd prefer `--color={always,auto,never}` instead of the confusingly-short `-c`. (I'm happy to add `auto` later myself. If it were to be the default, it would need to check isatty and `getenv("NO_COLOR")` per https://no-color.org/.)
* I'm not sure about the valgrind errors being yellow. Speculatively switching to yellow just in case anything else is printed seems a little eager. Why not just restore standard white for anything that doesn't need an explicit colour.

PR #812, opened 2025-10-06
Add runtime checks for invalid uses of \K in lookaround

### LINKED ISSUE #736: `\K` in lookbehind/lookahead should be always invalid
`\K` in lookbehind/lookahead should be always invalid as it affects "the global capture group" only, but lookbehind/lookahead does not participate in global capture group matching.

repro: https://3v4l.org/W9s8q

NWilson:
Fixes #736

### DIFF
--- a/src/pcre2_jit_match_inc.h
+++ b/src/pcre2_jit_match_inc.h
@@ -107,6 +107,7 @@ return match_data->rc = PCRE2_ERROR_JIT_BADOPTION;
 pcre2_real_code *re = (pcre2_real_code *)code;
 executable_functions *functions = (executable_functions *)re->executable_jit;
 pcre2_jit_stack *jit_stack;
+PCRE2_SIZE *ovector = match_data->ovector;
 uint32_t oveccount = match_data->oveccount;
 uint32_t max_oveccount;
 union {
@@ -171,6 +172,17 @@ if (jit_stack != NULL)
   }
 else
   rc = jit_machine_stack_exec(&arguments, convert_executable_func.call_executable_func);
+
+/* XXXXXXXXXX HACK
+Surely there's a better place to put this? I've just stuffed it in here, badly
+I guess it should go into the epilogue of the JIT-generated code somehow,
+conditionally on PCRE2_EXTRA_ALLOW_LOOKAROUND_BSK? */
+if (rc >= 0 &&
+    (ovector[0] < start_offset || ovector[0] > ovector[1]) &&
+    (re->extra_options & PCRE2_EXTRA_ALLOW_LOOKAROUND_BSK) == 0)
+  {
+  rc = PCRE2_ERROR_BAD_BACKSLASH_K;
+  }
 
 if (rc > (int)oveccount)
   rc = 0;
--- a/src/pcre2_match.c
+++ b/src/pcre2_match.c
@@ -1010,9 +1010,26 @@ fprintf(stderr, "++ %2ld op=%3d %s\n", Fecode - mb->start_code, *Fecode,
         }
 
 #ifdef DEBUG_SHOW_OPS
-      fprintf(stderr, "++ Failed ACCEPT not at end (endanchnored set)\n");
+      fprintf(stderr, "++ Failed ACCEPT not at end (endanchored set)\n");
 #endif
       return MATCH_NOMATCH;   /* (*ACCEPT) */
+      }
+
+    /* Fail if we detect that the start position was moved to be either after
+    the end position (\K in lookahead) or before the start offset (\K in
+    lookbehind). If this occurs, the pattern must have used \K in a somewhat
+    sneaky way (e.g. by pattern recursion), because if the \K is actually
+    syntactically inside the lookaround, it's blocked at compile-time. */
+
+    if (Fstart_match < mb->start_subject + mb->start_offset ||
+        Fstart_match > Feptr)
+      {
+      /* The \K expression is fairly rare. We assert it was used so that we
+      catch any unexpected invalid data in start_match. */
+      PCRE2_ASSERT(mb->hasbsk);
+
+      if (!mb->allowlookaroundbsk)
+        return PCRE2_ERROR_BAD_BACKSLASH_K;
       }
 
     /* We have a successful match of the whole pattern. Record the result and
@@ -7393,8 +7410,11 @@ mb->start_offset = start_offset;
 mb->end_subject = end_subject;
 mb->true_end_subject = true_end_subject;
 mb->hasthen = (re->flags & PCRE2_HASTHEN) != 0;
+mb->hasbsk = (re->flags & PCRE2_HASBSK) != 0;
 mb->allowemptypartial = (re->max_lookbehind > 0) ||
     (re->flags & PCRE2_MATCH_EMPTY) != 0;
+mb->allowlookaroundbsk =
+  (re->extra_options & PCRE2_EXTRA_ALLOW_LOOKAROUND_BSK) != 0;
 mb->poptions = re->overall_options;          /* Pattern options */
 mb->ignore_skip_arg = 0;
 mb->mark = mb->nomatch_mark = NULL;          /* In case never set */
--- a/src/pcre2.h.generic
+++ b/src/pcre2.h.generic
@@ -440,6 +440,7 @@ released, the numbers must not be changed. */
 #define PCRE2_ERROR_DIFFSUBSSUBJECT   (-72)
 #define PCRE2_ERROR_DIFFSUBSOFFSET    (-73)
 #define PCRE2_ERROR_DIFFSUBSOPTIONS   (-74)
+#define PCRE2_ERROR_BAD_BACKSLASH_K   (-75)
 
 
 /* Request types for pcre2_pattern_info() */
--- a/src/pcre2.h.in
+++ b/src/pcre2.h.in
@@ -440,6 +440,7 @@ released, the numbers must not be changed. */
 #define PCRE2_ERROR_DIFFSUBSSUBJECT   (-72)
 #define PCRE2_ERROR_DIFFSUBSOFFSET    (-73)
 #define PCRE2_ERROR_DIFFSUBSOPTIONS   (-74)
+#define PCRE2_ERROR_BAD_BACKSLASH_K   (-75)
 
 
 /* Request types for pcre2_pattern_info() */
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -8370,6 +8370,10 @@ for (;; pptr++)
       case ESC_A:
       if (cb->max_lookbehind == 0) cb->max_lookbehind = 1;
       break;
+
+      case ESC_K:
+      cb->external_flags |= PCRE2_HASBSK;  /* Record */
+      break;
       }
 
     *code++ = meta_arg;
--- a/src/pcre2_error.c
+++ b/src/pcre2_error.c
@@ -305,6 +305,7 @@ static const unsigned char match_error_texts[] =
   "substitute subject differs from prior match call\0"
   "substitute start offset differs from prior match call\0"
   "substitute options differ from prior match call\0"
+  "disallowed use of \\K in lookaround\0"
   ;
 
 
--- a/src/pcre2_internal.h
+++ b/src/pcre2_internal.h
@@ -538,6 +538,7 @@ bytes in a code unit in that mode. */
 #define PCRE2_DUPCAPUSED    0x00200000u /* contains (?| */
 #define PCRE2_HASBKC        0x00400000u /* contains \C */
 #define PCRE2_HASACCEPT     0x00800000u /* contains (*ACCEPT) */
+#define PCRE2_HASBSK        0x01000000u /* contains \K */
 
 #define PCRE2_MODE_MASK     (PCRE2_MODE8 | PCRE2_MODE16 | PCRE2_MODE32)
 
--- a/src/pcre2_intmodedep.h
+++ b/src/pcre2_intmodedep.h
@@ -969,7 +969,9 @@ typedef struct match_block {
   uint32_t match_call_count;      /* Number of times a new frame is created */
   BOOL hitend;                    /* Hit the end of the subject at some point */
   BOOL hasthen;                   /* Pattern contains (*THEN) */
+  BOOL hasbsk;                    /* Pattern contains \K */
   BOOL allowemptypartial;         /* Allow empty hard partial */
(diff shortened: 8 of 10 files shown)


src/pcre2_jit_match_inc.h
```diff
@@ -172,6 +173,17 @@ if (jit_stack != NULL)
...
+/* XXXXXXXXXX HACK
+Surely there's a better place to put this? I've just stuffed it in here, badly
+I guess it should go into the epilogue of the JIT-generated code somehow,
+conditionally on PCRE2_EXTRA_ALLOW_LOOKAROUND_BSK? */
+if (rc >= 0 &&
+    (ovector[0] < start_offset || ovector[0] > ovector[1]) &&
+    (re->extra_options & PCRE2_EXTRA_ALLOW_LOOKAROUND_BSK) == 0)
+  {
+  rc = PCRE2_ERROR_BAD_BACKSLASH_K;
+  }
```
NWilson: @zherczeg I have done something bad here I think. The code in this file is a tiny wrapper around the JIT, so it feels wrong to implement this check as a post-processing step here.

zherczeg: True. This should be done before the pattern returns with a valid match (and PCRE2_HASBSK is set, and PCRE2_EXTRA_ALLOW_LOOKAROUND_BSK is not set).

src/pcre2_jit_compile.c
```diff
@@ -13772,6 +13772,49 @@ common->accept_label = LABEL();
...
+  bad_eom = CMP(SLJIT_GREATER, TMP1, 0, TMP2, 0);
+
+  // If either comparison is true, return error and jump to abort
+  OP1(SLJIT_MOV, SLJIT_RETURN_REG, 0, SLJIT_IMM, PCRE2_ERROR_BAD_BACKSLASH_K);
+  JUMPTO(SLJIT_JUMP, common->abort_label);
+
+  // Patch the jumps to skip the error if the checks pass
+  JUMPHERE(bad_som);
+  JUMPHERE(bad_eom);
+  }
```
NWilson, line 13801: The JIT is still very new to me; I don't really understand how it all works yet.

I've made an attempt at some code here. It crashes, but I'm trying.

NWilson: OK, I have fixed it so that it doesn't crash.

However, the results of the comparisons don't seem correct.

How should I debug this code? Is there any way at all to set breakpoints, step through...? I can obviously step through the entire massive blob of code in assembler, but I have no idea how I'd find the few instructions that I added here.

src/pcre2_match.c
```diff
@@ -1010,11 +1010,28 @@ fprintf(stderr, "++ %2ld op=%3d %s\n", Fecode - mb->start_code, *Fecode,
...
 
 #ifdef DEBUG_SHOW_OPS
-      fprintf(stderr, "++ Failed ACCEPT not at end (endanchnored set)\n");
+      fprintf(stderr, "++ Failed ACCEPT not at end (endanchored set)\n");
 #endif
       return MATCH_NOMATCH;   /* (*ACCEPT) */
       }
 
+    /* Fail if we detect that the start position was moved to be either after
+    the end position (\K in lookahead) or before the start offset (\K in
```
zherczeg, line 1019: I am not sure I understand the operation.

Why start offset is a problem? I think the only issue is when ovector(0) > ovector(1), which confuses some simple implementations. Nobody complained about startoffset before.

Do this happens when \K is executed, or as a post check?

NWilson: * Start offset is a problem, because when you are doing "search all matches" you don't want matches to overlap or go "backwards". The canonical list of matches should be ordered, non-overlapping, and without duplicates. I believe that many clients will expect this. For example, pcre2_substitute itself fails if the list of matches is overlapping.
* The checks are done as a post check, after a match is accepted (much, much later than when \K is encountered). These checks do not cause backtracking: it simply turns an accepted match into an error.

src/pcre2_jit_compile.c
zherczeg: `sljit_emit_op0(compiler, SLJIT_BREAKPOINT);` stops gdb where you put it

src/pcre2_match.c
zherczeg: This rule is strange to me. As long as the end of the matches are <= ordered, the beginning should not matter when you do a global match. This is true now regardless of \K. Substitute (string insert) should not care where the match starts as long as it is <= than the end.

NWilson: It does matter for substitute.

Here is the check that Philip added, long before my involvement: 

https://github.com/PCRE2Project/pcre2/blob/d29e729000a3724e2aebaa64318dfd7530a55370/src/pcre2_substitute.c#L454

Here's one way to think about it: a match that ends before it starts is weird. But also, if the unmatched text (text in between matches) that is backwards is equally weird. 

Any application that does something with the text in between matches, rather than just throwing it away, wants the matches to be non overlapping. 

Example: 
Subject "abcde", and matches "a", "cd", "de".
If you wanted to split the string on matches (very common), you'd get:

Empty unmatched text at start
a, match
b, unmatched
cd, match
!!! Unmatched d backwards
de, match
Empty unmatched text at end

Find and replace is based around the concept of string split. If you wanted to replace all those three matches with "zzz", what would the result even be? After replacing "cd" with "zzz", how do you even begin to then replace "de"?

The invariant must be that for both matches and the unmatched text in between, the substring start is before the end (or "not after").

Finally, even if applications only want a single match, it is surprising (and possibly broken) to get a match starting at offset 3, if you ask for searching to start from offset 4.

src/pcre2_jit_compile.c
NWilson: Thank you! I've fixed my JIT bug now, and I'm pretty confident the code does what I intend (although please, do check - I'm really not confident with sljit).

src/pcre2_jit_compile.c
```diff
@@ -13772,6 +13772,41 @@ common->accept_label = LABEL();
...
+  bad_som_behind = CMP(SLJIT_LESS, TMP1, 0, TMP3, 0);  /* (ovector[0] < jit_arguments->str)? */
+  bad_som_ahead = CMP(SLJIT_GREATER, TMP1, 0, TMP2, 0);  /* (ovector[0] > ovector[1])? */
+  fallthrough = JUMP(SLJIT_JUMP);
+
+  JUMPHERE(bad_som_behind);
+  JUMPHERE(bad_som_ahead);
+  OP1(SLJIT_MOV, SLJIT_RETURN_REG, 0, SLJIT_IMM, PCRE2_ERROR_BAD_BACKSLASH_K);
+  add_jump(compiler, &common->abort, JUMP(SLJIT_JUMP));
+
+  JUMPHERE(fallthrough);
```
zherczeg: Too many jumps. Perhaps we could preload SLJIT_RETURN_REG (TMP1) with an error code.

src/pcre2_jit_compile.c
```diff
@@ -13772,6 +13772,41 @@ common->accept_label = LABEL();
...
+    {
+    OP1(SLJIT_MOV, TMP1, 0, ARGUMENTS, 0);
+    OP1(SLJIT_MOV, TMP3, 0, SLJIT_MEM1(TMP1), SLJIT_OFFSETOF(jit_arguments, str));
+    }
+  else
+    {
+    OP1(SLJIT_MOV, TMP3, 0, SLJIT_MEM1(ARGUMENTS), SLJIT_OFFSETOF(jit_arguments, str));
+    }
+  OP1(SLJIT_MOV, TMP1, 0, SLJIT_MEM1(SLJIT_SP), OVECTOR(0));
+  OP1(SLJIT_MOV, TMP2, 0, STR_PTR, 0);  /* STR_PTR, since OVECTOR(1) hasn't been populated yet */
```
zherczeg: No need to copy. CMP can use this directly. Also, it can use memory arguments as well.

src/pcre2_match.c
zherczeg: Using `max(start_offset, ovector[0])` is ok for replacing. Throwing errors is not necessary. Especially that the pattern cannot detect this case.

src/pcre2_jit_compile.c
NWilson: I've done the other change, but this one seems a little harder.

We have

```
if (A || B) return ERROR;
// ...fallthrough
```

I've implemented it as:

```
if (A) goto do_return;
if (B) goto do_return;
goto fallthrough;
do_return: return ERROR;
fallthrough:
```

I'm struggling to see other ways of rearranging it.

```
if (A) goto do_return;
if (NOT_B) goto fallthrough;
do_return: return ERROR;
fallthrough:
```

Is that better? There would be two CMPs, each one with its own JUMPHERE, and without the extra goto.

NWilson: I've pushed this change. Let me know if you're happy.

src/pcre2_jit_compile.c
```diff
@@ -13772,6 +13772,41 @@ common->accept_label = LABEL();
...
+  else
+    {
+    OP1(SLJIT_MOV, TMP1, 0, SLJIT_MEM1(ARGUMENTS), SLJIT_OFFSETOF(jit_arguments, str));
+    }
+
+  /* (ovector[0] < jit_arguments->str)? */
+  bad_som_behind = CMP(SLJIT_LESS, SLJIT_MEM1(SLJIT_SP), OVECTOR(0), TMP1, 0);
+  /* (ovector[0] > STR_PTR)?  NB. ovector[1] hasn't yet been set to STR_PTR. */
+  bad_som_ahead = CMP(SLJIT_GREATER, SLJIT_MEM1(SLJIT_SP), OVECTOR(0), STR_PTR, 0);
+  fallthrough = JUMP(SLJIT_JUMP);
```
zherczeg: This jump is still there. In general, it is better to avoid unnecessary jumps. Just use TMP2 instead of TMP1, and preload the error. Double loading is also bad, use TMP2 to store OVECTOR(0). Compare it with arguments->str (single load), and OVECTOR(1) (single load).

NWilson: I see what you mean. Thanks for helping me learn.

I think I've done it better now!

src/pcre2_jit_compile.c
```diff
@@ -13772,6 +13772,31 @@ common->accept_label = LABEL();
...
+    {
+    OP1(SLJIT_MOV, TMP2, 0, SLJIT_MEM1(ARGUMENTS), SLJIT_OFFSETOF(jit_arguments, str));
+    }
+  OP1(SLJIT_MOV, TMP3, 0, SLJIT_MEM1(SLJIT_SP), OVECTOR(0));
+
+  OP1(SLJIT_MOV, SLJIT_RETURN_REG, 0, SLJIT_IMM, PCRE2_ERROR_BAD_BACKSLASH_K);
+  /* (ovector[0] < jit_arguments->str)? */
+  add_jump(compiler, &common->abort, CMP(SLJIT_LESS, TMP3, 0, TMP2, 0));
+  /* (ovector[0] > STR_PTR)?  NB. ovector[1] hasn't yet been set to STR_PTR. */
+  add_jump(compiler, &common->abort, CMP(SLJIT_GREATER, TMP3, 0, STR_PTR, 0));
```
zherczeg: This is better. However, TMP3 is emulated on x86-32, and better not use it whenever is possible:
```
  OP1(SLJIT_MOV, TMP2, 0, SLJIT_MEM1(SLJIT_SP), OVECTOR(0));
  if (HAS_VIRTUAL_REGISTERS)
    {
    OP1(SLJIT_MOV, TMP1, 0, ARGUMENTS, 0);
    OP1(SLJIT_MOV, TMP1, 0, SLJIT_MEM1(TMP2), SLJIT_OFFSETOF(jit_arguments, str));
    }
  else
    {
    OP1(SLJIT_MOV, TMP1, 0, SLJIT_MEM1(ARGUMENTS), SLJIT_OFFSETOF(jit_arguments, str));
    }
  OP2U(SLJIT_SUB | SLJIT_SET_LESS, TMP2, 0, TMP1, 0);
  OP1(SLJIT_MOV, SLJIT_RETURN_REG, 0, SLJIT_IMM, PCRE2_ERROR_BAD_BACKSLASH_K);
  add_jump(compiler, &common->abort, JUMP(SLJIT_LESS));
  add_jump(compiler, &common->abort, CMP(SLJIT_GREATER, TMP2, 0, STR_PTR, 0));
```
(Or something similar)
The trick is that the flag register is set, then TMP1 is overwritten (but the flags are not), than jump based on the flag.

NWilson: OK, I've done that renaming. It seems a lot of effort for an obscure platform (x86-32). After reserving some registers for use by the matcher (like R1=STR_PTR), writing code with just two scratch registers feels really constrained.

Anyway, I'm happy if that's how you'd like it done.

src/pcre2_jit_compile.c
```diff
@@ -13772,6 +13772,34 @@ common->accept_label = LABEL();
...
+  OP1(SLJIT_MOV, TMP2, 0, SLJIT_MEM1(SLJIT_SP), OVECTOR(0));
+
+  /* (ovector[0] < jit_arguments->str)? */
+  OP2U(SLJIT_SUB | SLJIT_SET_LESS, TMP2, 0, TMP1, 0);
+  /* Unconditionally set R0 (aka TMP1), in between the comparison that needs to
+  use TMP1, but before the jump. */
+  OP1(SLJIT_MOV, SLJIT_RETURN_REG, 0, SLJIT_IMM, PCRE2_ERROR_BAD_BACKSLASH_K);
+  add_jump(compiler, &common->abort, JUMP(SLJIT_LESS));
+  /* (ovector[0] > STR_PTR)?  NB. ovector[1] hasn't yet been set to STR_PTR. */
+  add_jump(compiler, &common->abort, CMP(SLJIT_GREATER, TMP2, 0, STR_PTR, 0));
```
zherczeg, line 13800: The good is good. I am still unsure what would be the best way to handle \K. Throwing an error is one thing, but how this error should be handled by users?

NWilson: The expected way to handle error is by... fixing the regex? Or at least, stopping iteration if you're searching for matches. This is an incredibly obscure use-case: the user has deliberately or accidentally used subroutines to escape out of the protection offered by Perl's rule disallowing \K within lookaround.

If users write these regexes still... well, I suppose the user can always enable the PCRE flag to allow it again, and handle it in their application. But 99% of applications won't be able to handle this gracefully, so it's good to forbid it by default.

NWilson:
And this one, do you object to me merging?

zherczeg:
Nope. Lets merge it.

PR #813, opened 2025-10-07
Modernise FindReadline.cmake, and fix on Solaris

NWilson:
The CMake build fails on Solaris out of the box, because our Autoconf has clever logic to add `-ltermcap` as required by libreadline, but the CMake scripts just don't. And worse, CMake defaults to search for and link to all the third-party libraries, whereas Autoconf is set to not link them unless requested.

* Fix Solaris vanilla CMake build; remove CI override that forces use of termcap.
* Move up pcre2_use_system_extensions to the top of CMakeLists.txt
* Use cmake_push_check_state() instead of manual management of state
* Move _GNU_SOURCE detection into pcre2_use_system_extensions, to match Autoconf


CMakeLists.txt
```diff
@@ -537,19 +551,7 @@ if(PCRE2_SUPPORT_JIT)
 endif()
 
 if(PCRE2_SUPPORT_JIT_SEALLOC)
-  set(CMAKE_REQUIRED_DEFINITIONS -D_GNU_SOURCE)
-  check_symbol_exists(mkostemp stdlib.h REQUIRED)
-  unset(CMAKE_REQUIRED_DEFINITIONS)
-  if(${REQUIRED})
-    if(${CMAKE_SYSTEM_NAME} MATCHES Linux|NetBSD)
```
carenas, line 544: need to keep this restriction, sealloc was only tested on those two OS and only barely works.

NWilson: I think the restriction is still there - the OS check is present above as well, and it was just duplicated here.

NWilson: It's good feeback though. I added some additional checks for arguments that could be specified on the CMake commandline or set in the cache, but are unsupported. That wasn't a case we had guarded against much before.

sergio-nsk:
This change, the commit 0ee6302238faf8e99e3eb9ef8655a4c7b1f3f16c, breaks the build on Ubuntu 20 with ninja 1.10.0 and cmake 4.2.3.
```
-- Looking for mkostemp
CMake Error:
  Running

   '/usr/bin/ninja' '-C' 'out/u20/CMakeFiles/CMakeScratch/TryCompile-dbWCWO' '-t' 'recompact'

  failed with:

   ninja: error: build.ninja:50: bad $-escape (literal $ must be written as $$)





CMake Error at /usr/share/cmake-4.2/Modules/CheckSymbolExists.cmake:193 (try_compile):
  Failed to generate test project build system.
Call Stack (most recent call first):
  /usr/share/cmake-4.2/Modules/CheckSymbolExists.cmake:98 (__CHECK_SYMBOL_EXISTS_IMPL)
  third_party/pcre2/CMakeLists.txt:191 (check_symbol_exists)
```
This is because the global compile definitions has such entries below:
```
$<$<CONFIG:Debug>:DEBUG>;$<$<CONFIG:Debug>:_DEBUG>;$<$<NOT:$<CONFIG:Debug>>:NDEBUG>;$<$<NOT:$<CONFIG:Debug>>:_NDEBUG>
```

carenas:
> breaks the build on Ubuntu 20 with ninja 1.10.0 and cmake 4.2.3.

@sergio-nsk: assuming you meant Ubuntu 20.04, I hadn't been able to reproduce your problem:

```
# cat /etc/os-release | grep VERSION
VERSION="20.04.6 LTS (Focal Fossa)"
VERSION_ID="20.04"
VERSION_CODENAME=focal
# ninja --version
1.10.0
# cmake-4.2.3-linux-aarch64/bin/cmake -G Ninja -B build
# cd build && ../cmake-4.2.3-linux-aarch64/bin/cmake --build .
```

judging by the error message, I would expect the bug (if any) to be in `cmake`, not on our code too

sergio-nsk:
@carenas Please try this CMakeLists.txt in the parent directory:
```cmake
cmake_minimum_required(VERSION 3.15 FATAL_ERROR)

project(issue-813)

add_compile_definitions($<$<CONFIG:Debug>:DEBUG>)
add_subdirectory(pcre2)
```
Run it on Ubuntu 20:
```
$ cmake -S. -Bout -GNinja
-- The C compiler identification is Clang 12.0.0
-- The CXX compiler identification is Clang 12.0.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Using CMake version 4.2.3 (/usr/bin/cmake)
-- Could NOT find BZip2 (missing: BZIP2_LIBRARIES BZIP2_INCLUDE_DIR) 
-- Found ZLIB: /usr/lib/x86_64-linux-gnu/libz.so (found version "1.2.11")
-- Could NOT find Readline (missing: READLINE_LIBRARY READLINE_INCLUDE_DIR) 
-- Could NOT find Editline (missing: EDITLINE_LIBRARY EDITLINE_INCLUDE_DIR) 
-- Looking for assert.h
-- Looking for assert.h - found
-- Looking for dirent.h
-- Looking for dirent.h - found
-- Looking for sys/stat.h
-- Looking for sys/stat.h - found
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for unistd.h
-- Looking for unistd.h - found
-- Looking for windows.h
-- Looking for windows.h - not found
-- Detecting platform feature test macros
-- Detecting platform feature test macros - _GNU_SOURCE
-- Looking for mkostemp
CMake Error:
  Running

   '/usr/bin/ninja' '-C' '/home/ubuntu/Work/solid.tmp/third_party/out/CMakeFiles/CMakeScratch/TryCompile-D7KIMc' '-t' 'recompact'

  failed with:

   ninja: error: build.ninja:50: bad $-escape (literal $ must be written as $$)
    FLAGS = -std=gnu99   -D$<$<CONFIG:Debug>:DEBUG>
                           ^ near here



CMake Error at /usr/share/cmake-4.2/Modules/CheckSymbolExists.cmake:193 (try_compile):
  Failed to generate test project build system.
Call Stack (most recent call first):
  /usr/share/cmake-4.2/Modules/CheckSymbolExists.cmake:98 (__CHECK_SYMBOL_EXISTS_IMPL)
  pcre2/CMakeLists.txt:191 (check_symbol_exists)


-- Configuring incomplete, errors occurred!
```

NWilson:
Very odd. I can reproduce with that testcase. However, I don't understand it. The mkostemp check is failing with that very unexpected message. Somehow, the act of pushing and popping the feature check state is causing the compile definition to be set as a literal rather than a generator expression? I don't understand, I'll need to trace the CMake to see what's happening.

I do apologise for this regression. If you have any suggestions on how to fix it, I'd be grateful.

NWilson:
Oh oh! I'm a dummy, I see what I've done. There's a line where I propagate compile definitions... which is fine, except that I hadn't realised a) that it propagates customer definitions from the parent; and b) that you can't propagate generator expressions.

So, it was much more brittle than I realised.

I have put out a PR with the fix: #883

sergio-nsk:
> If you have any suggestions on how to fix it, I'd be grateful.

I don't have a fix, I have a work around in the parent cmake file:
```cmake
get_property(is_multi_config GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
if(is_multi_config)
    add_compile_definitions($<$<CONFIG:Debug>:DEBUG>)
elseif(CMAKE_BUILD_TYPE STREQUAL Debug)
    add_compile_definitions(DEBUG)
endif()
```

> that you can't propagate generator expressions.

I believe it may depend on the cmake version or its scripts. There is no such issue on Ubuntu 22 or newer. For some reasons, generator expressions are not expanded if the generator is not multi-configuration.

PR #814, opened 2025-10-08
Further updates to modernise CMake & PCRE_EXPORT documentation

NWilson:
There's nothing terribly wrong about what we're currently doing. In fact, it's arguably better than the Autoconf way it's being done, due to limitations of libtool. (CMake is just cleaner than libtool, and supports many more platforms.)

@carenas, what do you think? Is it better to be using GenerateExportHeader, or just carry on with our own manual detection of the `__attribute__((visibility("default")))`, and pass it in via the existing config.h file?

I'm 50/50 undecided myself. There's still lots of useful changes in this PR, but I'm considering cutting it right down, rather than doing all the export.h stuff in this draft.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -327,8 +327,9 @@ src/pcre2.h.generic: src/pcre2.h.in configure.ac
 # running configure in a new directory, it complains. For this reason, we move
 # config.status out of the way while doing the default configuration. The
 # resulting config.h is munged by perl to put #ifdefs round any #defines for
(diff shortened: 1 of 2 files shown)


NWilson:
What do you think @carenas? Is there any benefit at all to switching to CMake's GenerateExportHeader?

I think not, after considering it. I'll remove those changes from this PR.

NWilson:
- [x] Note to self - HAVE_VSCRIPT_SUN seems to be failing on our Solaris build. Hmm..

PR #818, opened 2025-10-14
Rework ARM64 SIMD support

zherczeg:
ARM64 SIMD code layout should be similar to other architectures


zherczeg:
Is macos the 64 bit ARM platform in the CI?

zherczeg:
The JIT test prints the cpu, but the output log removes the output of test programs. We don't have that much output, does it worth it?

NWilson:
> Is macos the 64 bit ARM platform in the CI?

Yes, and also a second one is the rather slow Linux with qemu, which isn't run automatically on PRs. You can run it manually though on the Actions tab (Dev).

NWilson:
> The JIT test prints the cpu, but the output log removes the output of test programs. We don't have that much output, does it worth it?

Which test specifically is that? We do check the return status of each test app, so we know they are not crashing or returning an error code

NWilson:
This PR looks cool! 

I'm not really qualified to review it though ..

zherczeg:
> Which test specifically is that?

The `pcre2_jit_test` prints the CPU related info when running, so you can see how the compiler is exactly configured. This is very useful for checking that the cpu and its available features are correctly detected.

zherczeg:
Probably I am stupid but I cannot an option to run "Dev / Multiarch". I suspect that is the qemu story.

NWilson:
> Probably I am stupid but I cannot an option to run "Dev / Multiarch". I suspect that is the qemu story.

I couldn't either... your branch is in your fork and I can't seem to run any jobs on it.

I copied your branch to this repo, and started a Dev job for it: ~~ https://github.com/PCRE2Project/pcre2/actions/runs/18494134606 ~~

Updated: https://github.com/PCRE2Project/pcre2/actions/runs/18497408852

NWilson:
> > Which test specifically is that?
> 
> The `pcre2_jit_test` prints the CPU related info when running, so you can see how the compiler is exactly configured. This is very useful for checking that the cpu and its available features are correctly detected.

Ah I see. I should update the CI jobs to print full test output, now I know there's a use for it.

ltrzesniewski:
> a second one is the rather slow Linux with qemu, which isn't run automatically on PRs.

There's a [`ubuntu-24.04-arm` partner runner image](https://github.com/actions/partner-runner-images) for Linux on ARM64 (I even use it in PCRE.NET).

zherczeg:
Thank you for the dev branch. I realized I made the code in release mode, and there was a minor assertion failure in debug mode. I passed an incorrect flag, which was ignored in release.

NWilson:
Good news - the build passed with that latest change.

zherczeg:
Thank you. If no objections, I plan to land this patch before the release.

PR #820, opened 2025-10-17
Fix build instructions for autoconf

Phantom1003:


NWilson:
You don't need to run autogen, since the instructions refer to building official released versions.

PR #821, opened 2025-10-17
Output CI test logs always

NWilson:
In a recent PR, Zoltan wanted to see the output from pcre2_jit_test in the CI logs, but it's not being stored currently.

It seems reasonable. I'll output it on all the test jobs.


zherczeg:
Nice! It is interesting, that some targets (Linux, alpine) has multiple tests in the log, while others (Mac) has a single one. It is easier to find the tests in the former.

zherczeg:
Thank you!

PR #822, opened 2025-10-17
Add OpenBSD and very old RHEL autoconf 2.69 CI builds

NWilson:
I have added a RHEL 8.6 build to exercise Autoconf 2.69, and raised the minimum requirement to match our testing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -344,15 +344,12 @@ jobs:
             ../maint/RunManifestTest install-dir ../maint/manifest-cmakeinstall-freebsd
             ../maint/RunSymbolTest install-dir/lib/ ../maint/
 
-  # XXX/TODO consider getting rid of: strcpy, strcat, sprintf (to fix OpenBSD warnings)
   openbsd:
     name: OpenBSD
     runs-on: ubuntu-latest
-    # XXX remove 'pull_request' before merging
     if: |
       (github.event_name == 'workflow_dispatch' && (inputs.job_id == 'all' || inputs.job_id == 'openbsd')) ||
-      github.event_name == 'push' ||
-      github.event_name == 'pull_request'
+      github.event_name == 'push'
     steps:
       - name: Checkout
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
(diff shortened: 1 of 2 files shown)


NWilson:
@carenas I'm going to need your help with the OpenBSD build. The JIT is failing to allocate memory, and I don't know how to fix it. We don't have any docs on OpenBSD...

.github/workflows/build.yml
```diff
@@ -343,6 +344,63 @@ jobs:
...
+
+            export MALLOC_OPTIONS="USRJGFC>>" # XXX I have no idea - should I set this?
+
+            cp -rp . ../build-autoconf
+            cp -rp . ../build-cmake
+
+            echo "== Autoconf =="
+            cd ../build-autoconf
+
+            ./configure CFLAGS="$CFLAGS_GCC_STYLE" --enable-jit --enable-pcre2-16 --enable-pcre2-32 --enable-Werror
```
carenas: CPPFLAGS=-DSLJIT_WX_EXECUTABLE_ALLOCATOR

RunGrepTest
```diff
@@ -859,8 +859,13 @@ if [ $? -ne 0 ]; then
   echo "RC=2" >>testtrygrep
 else
 
-  (cd $srcdir; unset LC_ALL; LC_CTYPE=locale.bad $valgrind $vjs $pcre2grep abc /dev/null) >>testtrygrep 2>&1
+  (cd $srcdir; unset LC_ALL; export LC_CTYPE=locale.bad; $valgrind $vjs $pcre2grep abc /dev/null >>$builddir/testtrygrep 2>&1) >testtemp1grep 2>&1
```
carenas: This is likely broken in at least Solaris.

POSIX shell requires export and setting the variables in two different commands

NWilson: Of course, thank you for the reminder of that pitfall.

NWilson:
Thank you Carlo. In future, I should probably do some CMake & Autoconf tests to set that SLJIT_WX_EXECUTABLE_ALLOCATOR flag automatically. And, on Linux, there is a way to do runtime detection of `W^X`, so Linux distributions should probably ship a build of PCRE2 which has both allocators, and checks for the policy the first time it does an allocation to choose which allocator to use.

I won't tackle that now though.

PR #825, opened 2025-10-24
Use correctly-cased CMake variables for better compatibility

### LINKED ISSUE #824: CMake dependencies _FOUND tests are wrong. CMake variables are case sensitive
Hi, 

You are using CMake `find_package(<PackageName> ...)` to find dependencies. Based on documentation https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature this call will define `<PackageName>_FOUND` variable. 

```
find_package(BZip2)
find_package(ZLIB)
find_package(Readline)
find_package(Editline)
```

But later in pcre2 CMakeLists.txt, these are used as `<PACKAGENAME>_FOUND` (uppercase). CMake variables are case sensitive. 

So for example `find_package(BZip2)` will define `BZip2_FOUND`, not `BZIP2_FOUND` as You use for example here https://github.com/PCRE2Project/pcre2/blob/master/CMakeLists.txt#L501C6-L501C17. So it will raise a FATAL_ERROR.

```cmake
if(PCRE2_SUPPORT_LIBBZ2)
  if(BZIP2_FOUND)
    include_directories(${BZIP2_INCLUDE_DIR})
  else()
    message(
      FATAL_ERROR
      " libbz2 not found. Set BZIP2_INCLUDE_DIR to a compatible header\n"
      " or set BZip2_ROOT to a full bzip2 installed tree, as needed."
    )
  endif()
endif()
```

Here is how CMake builtin finder for BZip2 works https://cmake.org/cmake/help/latest/module/FindBZip2.html#result-variables

NWilson:
* Use the correct casing for `<PackageName>_FOUND`
* Add a CI test with CMake very latest (currently 4.1.2). This job also sets DCMAKE_POLICY_VERSION_MINIMUM to verify that we don't break any future policies, greater than our minimum version 3.15.
* Tidy the ncurses support for libreadline to use READLINE_LIBRARIES

Fixes #824


cmake/FindReadline.cmake
NWilson: Note to self - I could suppress this spammy line of logging

PR #826, opened 2025-10-24
Write padding values to ensure pcre2_serialize_encode() outputs defined values

NWilson:
Fixes low-severity valgrind error reported in GHSA-q7rw-r7qq-2hx6.

### DIFF
--- a/src/pcre2_compile_class.c
+++ b/src/pcre2_compile_class.c
@@ -1802,17 +1802,14 @@ if ((xclass_props & XCLASS_REQUIRED) != 0)
       PUT(code, 0, (uint32_t)(char_lists_size >> 1));
       code += LINK_SIZE;
 
-#if defined PCRE2_DEBUG || defined SUPPORT_VALGRIND
+      /* If we added padding to align the list, initialize the bytes to
+      defined values, so the library is valgrind-clean. It could also
+      be a security concern for clients calling into PCRE2 via bindings
+      from a memory-safe language, if pcre2_serialize_encode() exposes
+      uninitialized memory that may contain sensitive information. */
+
       if ((char_lists_size & 0x2) != 0)
-        {
-        /* In debug the unused 16 bit value is set
-        to a fixed value and marked unused. */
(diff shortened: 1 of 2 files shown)


src/pcre2_compile_class.c
```diff
@@ -1802,17 +1802,14 @@ if ((xclass_props & XCLASS_REQUIRED) != 0)
...
+      uninitialized memory that may contain sensitive information. */
+
       if ((char_lists_size & 0x2) != 0)
-        {
-        /* In debug the unused 16 bit value is set
-        to a fixed value and marked unused. */
-        ((uint16_t*)data)[-1] = 0x5555;
-#ifdef SUPPORT_VALGRIND
-        VALGRIND_MAKE_MEM_NOACCESS(data - 2, 2);
-#endif
```
carenas, line 1813: even if the data is unconditionally initialized it should be made inaccessible for reads, since it is not valid, so this part of the code is still needed IMHO.

carenas (changes requested):

src/pcre2_compile_class.c
NWilson: On the other hand, the test then fails. The data can't be no-access, because the serialisation code reads it. But if it is uninitialised (for valgrind) then we can't use valgrind to assert that all the memory has been set. 

That's a nasty trade off. I wonder if there's a way to benefit both ways...?

carenas: Technically; the fact that the serialization code is reading it is a bug that was introduced when this feature was added in 10.45, and therefore we should fix that as well by making sure that when the code is serialized all holes are skipped instead.

PhilipHazel: Or you could just define the holes as something and specify that they always contain zeros or some other value.

zherczeg: I feel we overcomplicate this. Just write any value there.

NWilson: I agree with Philip and Zoltan. My PR was just writing a value in. That's all we need.

PR #827, opened 2025-10-25
fixup

carenas:
re-add valgrind tag which AFAIK doesn't seem to fail, at least with the current tests


NWilson:
I was puzzled by this too. It doesn't make sense to me: if you mark the memory as "no-access", surely valgrind _should_ generate an error when the serialization code does a memcpy from these bytes?

I think it's just a valgrind bug. Valgrind has special code to allow you to do memcpy from bad regions, and it propagates the badness (taint propagation). So it should catch the memcpy from a "no-access" region but it doesn't, because it's just propagating the taint.

We _could_ do `VALGRIND_MAKE_MEM_UNDEFINED` instead... but that would then cause the test to fail.

I think I'll just leave it as it is, sorry.

PR #829, opened 2025-10-25
Fix incorrect version in docs

ltrzesniewski:
I noticed the version is incorrect since these features were released in 10.47.

(I know I don't need to fix all the file types, but I'd hate to submit an inconsistent PR 😇)


NWilson:
Oops! The 10.46 was an emergency release. We'd already done most of the work for the next release of master, and I'd written the docs assuming it would be 10.46. But then the security issue came along, and I did a hotfix, and forgot to update these two references to 10.46, after it was basically renamed to 10.47.

PR #830, opened 2025-10-25
Update to Unicode 17

### LINKED ISSUE #823: Update to Unicode 17
Unicode releases are roughly annual, and v17 came out in Sept 2025.

I can't see any changes that have any obvious compatibility concerns. It should be a simple routine update to the data files.

NWilson:
Fixes #823


testdata/testinput26
```diff
@@ -2562,7 +2562,7 @@
 
 # Character not in script
 /^\p{Tangut}/utf
-    \x{18d09}
+    \x{18d20}
```
NWilson, line 2565: One of these tests in testinput26 was easy to update - they just added a few more characters to this script, so it was easy to bump the character being tested for "not in the script".

But there's a failing test higher up for Han, and I haven't yet worked out what's going on. The character in the test is still unassigned, but PCRE2 now seems to think it is the Han script.

I need to look into why.

NWilson: I checked. They did expand the Han range to the whole page, including unallocated code points.

PR #834, opened 2025-10-28
pcre2test: keep track of internal API with sljit platform names (WIP)

carenas:
DO NOT MERGE

an early draft to help discussion on design, not even tested

See zherczeg/sljit#349 for context

### DIFF
--- a/doc/pcre2api.3
+++ b/doc/pcre2api.3
@@ -1296,8 +1296,9 @@ documentation for more details.
 .sp
   PCRE2_CONFIG_JITTARGET
 .sp
-The \fIwhere\fP argument should point to a buffer that is at least 64 code
-units long. (The exact length required can be found by calling
+The \fIwhere\fP argument should point to a buffer that is at least
+.\" DEFINE PCRE2_JITTARGET_BUFFSIZE_LIMIT
+71 code units long. (The exact length required can be found by calling
 \fBpcre2_config()\fP with \fBwhere\fP set to NULL.) The buffer is filled with a
 string that contains the name of the architecture for which the JIT compiler is
 configured, for example "x86 32bit (little endian + unaligned)". If JIT support
--- a/src/pcre2test.c
+++ b/src/pcre2test.c
@@ -314,6 +314,7 @@ needs to be when compiling one of the libraries). */
 #include "pcre2.h"
 #include "pcre2posix.h"
 #include "pcre2_internal.h"
+#include "pcre2_api_internal.h"
 
 /* We need access to some of the data tables that PCRE2 uses. The previous
 definition of PCRE2_PCRE2TEST makes some minor changes in the files. The
@@ -3046,7 +3047,7 @@ fprintf(f, "Unicode version %s", buf);
 static void
 print_jit_target(FILE *f)
 {
-char buf[VERSION_SIZE];
+char buf[PCRE2_JITTARGET_BUFFSIZE_LIMIT];
 config_str(PCRE2_CONFIG_JITTARGET, buf);
 fputs(buf, f);
 }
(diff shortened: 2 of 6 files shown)


doc/pcre2api.3
```diff
@@ -1296,8 +1296,9 @@ documentation for more details.
 .sp
   PCRE2_CONFIG_JITTARGET
 .sp
-The \fIwhere\fP argument should point to a buffer that is at least 64 code
-units long. (The exact length required can be found by calling
+The \fIwhere\fP argument should point to a buffer that is at least
+.\" DEFINE PCRE2_JITTARGET_BUFFSIZE_LIMIT
+71 code units long. (The exact length required can be found by calling
```
NWilson, line 1301: I was going to say we should _remove_ the documented buffer length, not add more mechanisms for exposing it.

The problem is that applications often can't control what version of PCRE2 they are used with at runtime, for example, someone writes the software, then another person upgrades the dependencies (OS packages, or anything else).

We should instead say, "don't assume that the buffer size will always be the same; use the pcre2_config() with where set to NULL to find out the required size".

src/pcre2test.c
```diff
@@ -3046,7 +3047,7 @@ fprintf(f, "Unicode version %s", buf);
 static void
 print_jit_target(FILE *f)
 {
-char buf[VERSION_SIZE];
+char buf[PCRE2_JITTARGET_BUFFSIZE_LIMIT];
 config_str(PCRE2_CONFIG_JITTARGET, buf);
```
NWilson, line 3051: Let's just dynamically allocate the buffer and not have any new sizing macros at all.

(That's the change I was going to make in my PR.)

carenas: While I agree that is what users of PCRE2 should do, here I was intentionally using the static buffer because this is a test application that should break if that constrain is not kept.

NWilson: If we remove the constraint, then there's nothing to check :)

doc/pcre2api.3
carenas: As you had found with the release of 10.47, what users consider an "API", extends to the error numbers or text and even buffer sizes.

While I agree that having documented buffer sizes is not very reliable, that is what we have right now and while I can see us "recommending" that they are not used, retiring them seems like something that will require that their use first be "deprecated" before we can safely assume no one is doing so.

NWilson: What we have now is a documented size of 64, which isn't big enough on RISC-V.

We have to change something, we shouldn't just leave it as it is (I think you agree). But updating the docs to 71 bytes is just making the same mistake again.

NWilson: I'm not very good with GitHub search. I tried this:

```
PCRE2_CONFIG_JITTARGET NOT is:fork NOT path:**/src/pcre2_config.c NOT /PCRE2_CONFIG_JITTARGET\s*2\s*$/ NOT path:**/*.html NOT path:**/*.3
```

Thankfully most users are more sensible than we are, and use a buffer of 128, or 256, or malloc the required size (PHP does that).

Not many people are using the PCRE2_CONFIG_JITTARGET API.

carenas: > But updating the docs to 71 bytes is just making the same mistake again.

my bad; this is just a draft that I pushed so it could be clearer what the whole change was like, and should include one extra commit already in sljit's master (which I didn't bother including, since a limit of 71 doesn't need it) and sljit's [PR349](https://github.com/zherczeg/sljit/pull/349), which hasn't been finalized either.

NWilson:
I will close this in favour of Carlo's second PR, which has conflicts with this.

Issue #831, opened 2025-10-26
Passing `-march=rv64gcb_zicond` parameter to compiler with JIT  enabled causes tests to fail on RISC-V


Puqns67:
This issue occurs on both `gcc 15.2.0` and `clang 21.1.4`.

[LastTest.log](https://github.com/user-attachments/files/23151300/LastTest.log)

<details>
<summary>Build & test log in 3b9197727e70bffded674d24100e7b4c80e7ef58</summary>
<pre>
puqns67@azusa ~/C/pcre2 (master)> cmake -Bbuild -GNinja -DPCRE2_SUPPORT_JIT=ON
-- Using CMake version 4.1.2 (/usr/bin/cmake)
-- The C compiler identification is Clang 21.1.4
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/lib/llvm/21/bin/clang - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found BZip2: /usr/lib64/libbz2.so (found version "1.0.8")
-- Looking for BZ2_bzCompressInit
-- Looking for BZ2_bzCompressInit - found
-- Found ZLIB: /usr/lib64/libz.so (found version "1.3.1")
-- Looking for readline
-- Looking for readline - found
-- Found Readline: /usr/lib64/libreadline.so
-- Could NOT find Editline (missing: EDITLINE_LIBRARY EDITLINE_INCLUDE_DIR) 
-- Looking for assert.h
-- Looking for assert.h - found
-- Looking for dirent.h
-- Looking for dirent.h - found
-- Looking for sys/stat.h
-- Looking for sys/stat.h - found
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for unistd.h
-- Looking for unistd.h - found
-- Looking for windows.h
-- Looking for windows.h - not found
-- Detecting platform feature test macros
-- Detecting platform feature test macros - _GNU_SOURCE
-- Looking for mkostemp
-- Looking for mkostemp - found
-- Looking for memfd_create
-- Looking for memfd_create - found
-- Looking for secure_getenv
-- Looking for secure_getenv - found
-- Performing Test HAVE_REALPATH
-- Performing Test HAVE_REALPATH - Success
-- Performing Test HAVE_ATTRIBUTE_UNINITIALIZED
-- Performing Test HAVE_ATTRIBUTE_UNINITIALIZED - Success
[...]

PR #835, opened 2025-10-28
pcre2test: dynamically allocate buffer for JITTARGET

carenas:
Only lightly tested, could be used as a basic "solution" for #831, that could be easily backported for 10.47.

### DIFF
--- a/doc/pcre2api.3
+++ b/doc/pcre2api.3
@@ -1,4 +1,4 @@
-.TH PCRE2API 3 "19 October 2025" "PCRE2 10.48-DEV"
+.TH PCRE2API 3 "29 October 2025" "PCRE2 10.48-DEV"
 .SH NAME
 PCRE2 - Perl-compatible regular expressions (revised API)
 .sp
@@ -1297,13 +1297,14 @@ documentation for more details.
   PCRE2_CONFIG_JITTARGET
 .sp
 The \fIwhere\fP argument should point to a buffer that is suitably aligned and
-at least 128 code units long. (The exact length required can be found by calling
-\fBpcre2_config()\fP with \fBwhere\fP set to NULL.) The buffer is filled with a
-string that contains the name of the architecture for which the JIT compiler is
-configured, for example "x86 32bit (little endian + unaligned)". If JIT support
-is not available, PCRE2_ERROR_BADOPTION is returned, otherwise the number of
-code units used is returned. This is the length of the string, plus one unit
-for the terminating zero.
+wide enough to hold the full string. The exact length required is returned when
+calling \fBpcre2_config()\fP with \fBwhere\fP set to NULL first. The buffer is
+filled with a string that contains the name of the architecture for which the JIT
+compiler is configured at build time, for example, a 64-bit ARM CPU that supports
+the Armv8.1 extension writes "ARM-64 (LSE) 64bit (little endian + unaligned)".
+If JIT support is not available, PCRE2_ERROR_BADOPTION is returned, otherwise
+the number of code units used is returned. This is the length of the string,
+plus one code unit for the NUL character.
 .sp
   PCRE2_CONFIG_LINKSIZE
 .sp
@@ -4606,6 +4607,6 @@ Cambridge, England.
 .rs
 .sp
 .nf
-Last updated: 19 October 2025
+Last updated: 29 October 2025
 Copyright (c) 1997-2024 University of Cambridge.
 .fi


doc/pcre2api.3
```diff
@@ -1296,8 +1296,8 @@ documentation for more details.
 .sp
   PCRE2_CONFIG_JITTARGET
 .sp
-The \fIwhere\fP argument should point to a buffer that is at least 64 code
-units long. (The exact length required can be found by calling
+The \fIwhere\fP argument should point to a buffer that is suitable aligned and
+at least 128 code units long. (The exact length required can be found by calling
```
zherczeg: I would not say anything about "at least".

src/pcre2test.c
```diff
@@ -3046,9 +3054,20 @@ fprintf(f, "Unicode version %s", buf);
 static void
 print_jit_target(FILE *f)
 {
-char buf[VERSION_SIZE];
-config_str(PCRE2_CONFIG_JITTARGET, buf);
-fputs(buf, f);
+PCRE2_SIZE len = pcre2_config(PCRE2_CONFIG_JITTARGET, NULL);
+const int bytes_per_codepoint = test_mode /  8;
+char *buf = malloc(len * 4);
```
zherczeg: Why do you need the `* 4`? This is an ascii string. I think we can promise that it will always be an ascii string.

carenas: We are getting the string through the PCRE2 API, which copies a PCRE2_SPTR into a `void *`, so we have to convert it back into an ASCII C string.

zherczeg: Does `config_str` did this before? Anyway, instead of copying `PCRE2_BIG_ENDIAN` macro just read the input as UCHAR values and the cpu will handle the endianness. You need to do this only if `sizeof(UCHAR) != 1`.

doc/pcre2api.3
carenas: "at least" is just being copied from the original wording, and agree is weird.

the next patch on this series changes the wording so it encourages the use of a dynamically allocated buffer instead, but since that "breaks API" I am not sure if it should be included in the minimal patch needed to "fix" the original issue and that should be independent enough for backport, since I don't think this is worth a "brown paper" release.

src/pcre2test.c
carenas: While I also don't like the endianess specific logic, it is there to address the other bug this API has, which is that it could crash in some cases or perform badly because it doesn't care about alignment. 1 byte reads are always safe, but wider ones are not.

zherczeg: This is the test code. And malloc returns with aligned data when the size is aligned. So reading it with character sizes are always correct (and aligned).

src/pcre2test.c
```diff
@@ -3037,7 +3037,19 @@ config_str(PCRE2_CONFIG_UNICODE_VERSION, buf);
...
+    (defined(_POWER) && defined(__64BIT__))) && !defined(__LITTLE_ENDIAN__)) || \
+  (defined(__ppc__) || defined(__powerpc__) || defined(_ARCH_PPC) || defined(_ARCH_PWR) || defined(_ARCH_PWR2) || defined(_POWER)) || \
+  (((defined(__mips__) && !defined(_LP64)) || defined(__mips64)) && !defined(__MIPSEL__)) || \
+  ((defined(__aarch64__) || defined(__arm__) || defined(__thumb__) || defined(_ARM) || defined(_M_ARM) || defined(_M_ARMT)) && defined(__ARM_BIG_ENDIAN)) || \
+  (defined(__sparc__) || defined(__sparc)) || \
+  (defined(__BYTE_ORDER__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__) || \
+  (defined(__BIG_ENDIAN__) && __BIG_ENDIAN__) || (defined(BIG_ENDIAN) && BIG_ENDIAN)
+#define PCRE2_BIG_ENDIAN 1
+#else
+#define PCRE2_LITTLE_ENDIAN 1
```
NWilson: We don't need anything to do with endianness, or alignment, or anything. We just need to change the config_str function to return a malloc'd buffer.

carenas: I see; got confused refactoring my original changes over the `pcretest_inc.h` refactor.

I also got a nicer documentation update that might be worth squashing over so we can skip the awkward 128 byte bump, but will need to take a break before pushing it (I am in PST), so let me know if that is a problem.

src/pcre2test.c
carenas: good point, it is also true that the JIT (after SPARC was removed and with all the targets being unaligned), wouldn't be affected, but we still have the bug with the other strings though, as it is really `pcre2_config()`, not the test code, the one doing the unaligned reading/writing.

src/pcre2test_inc.h
```diff
@@ -564,18 +564,23 @@ return 0;
...
   exit(1);
   }
 
-while (r1-- > 0) where[r1] = (char)buf[r1];
+buf = malloc(needed_len * sizeof(PCRE2_UCHAR));
+r2 = pcre2_config(what, buf);
+PCRE2_ASSERT(r2 == needed_len);
+
+while (r2-- > 0) where[r2] = (char)buf[r2];
+free(buf);
```
NWilson, line 593: The caller is still using a VERSION_SIZE buffer. I think you need to malloc a char buffer and return it for the caller to free. You can then remove the VERSION_SIZE macro.

Otherwise, I also like this PR. Great!

carenas: While I agree that VERSION_SIZE should be eventually removed, I want to keep this change minimal, so it can be safely backported to 10.47. Hopefully this version is still good enough.

NWilson (approved):

src/pcre2test_inc.h
NWilson: If you want this as a separate commit, that's OK by me. I'll do a follow-up to remove VERSION_SIZE immediately after this.

NWilson: Thank you for addressing my concerns. The code here appears to be correct & working. I like it.

src/pcre2test_inc.h
```diff
@@ -558,24 +558,41 @@ return 0;
 
 Arguments:
   what       the item to read
-  where      the 8-bit buffer to receive the string
+  where      the 8-bit buffer to receive the string (NULLABLE)
+  size       sizeof(where) or 0 tp ask for the buffer to be allocated
```
carenas: typo

doc/pcre2api.3
```diff
@@ -1296,14 +1296,15 @@ documentation for more details.
...
-for the terminating zero.
+The \fIwhere\fP argument should point to a buffer that is suitably aligned and
+wide enough to hold the full string. The exact length required is returned when
+calling \fBpcre2_config()\fP with \fBwhere\fP set to NULL first. The buffer is
+filled with a string that contains the name of the architecture for which the JIT
+compiler is configured at build time, for example, a 64-bit ARM CPU that supports
+the Armv8.1 extension writes "ARM-64 (LSE) 64bit (little endian + unaligned)".
+If JIT support is not available, PCRE2_ERROR_BADOPTION is returned, otherwise
+the number of code units used is returned. This is the length of the string,
+plus one code unit for the NUL character.
```
carenas, line 1307: this is missing an historical overview and an explanation of why the use of static buffers is discouraged.

NWilson: I'm happy with it. I can revise it in the next commit

PR #841, opened 2025-11-03
Fix character list generator

zherczeg:


NWilson:
Great, thank you so much Zoltan for looking into this! The fix is in the sort of area where I expected it to be.

But I'd need to spend some time to review exactly how the structures work before I can double-check the fix. It looks good though.

zherczeg:
I would like to land this fix before I completely forget about it.

NWilson:
!! I did forget about it, my apologies. Please, do merge it.

NWilson:
This commit is a definite candidate for a backport to older versions.

PR #848, opened 2025-11-20
Use named fields for match frame members

zherczeg:
I hoped the memory can be reduced this way, but no luck so far. The size before `eptr` is 80 bytes before and after the patch, since char_repeat uses all members. However, due to using a struct type, we might be able to reduce the size of some members in the future.

### DIFF
--- a/src/pcre2_match.c
+++ b/src/pcre2_match.c
@@ -169,9 +169,9 @@ enum { RM200=200, RM201, RM202, RM203, RM204, RM205, RM206, RM207,
 /* Define short names for general fields in the current backtrack frame, which
 is always pointed to by the F variable. Occasional references to fields in
 other frames are written out explicitly. There are also some fields in the
-current frame whose names start with "temp" that are used for short-term,
-localised backtracking memory. These are #defined with Lxxx names at the point
-of use and undefined afterwards. */
+current frame that are used for short-term, localised backtracking memory.
+These are members of the fields union and #defined with Lxxx names at the
+point of use and undefined afterwards. */
 
 #define Fback_frame        F->back_frame
 #define Fcapture_last      F->capture_last
@@ -180,7 +180,6 @@ of use and undefined afterwards. */
 #define Feptr              F->eptr
 #define Fgroup_frame_type  F->group_frame_type
 #define Flast_group_offset F->last_group_offset
-#define Flength            F->length
 #define Fmark              F->mark
 #define Frdepth            F->rdepth
 #define Fstart_match       F->start_match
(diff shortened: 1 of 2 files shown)


src/pcre2_match.c
```diff
@@ -1293,13 +1292,14 @@ fprintf(stderr, "++ %2ld op=%3d %s\n", Fecode - mb->start_code, *Fecode,
...
-#define Lc           F->temp_32[2]
-#define Loc          F->temp_32[3]
+#define Llength      F->fields.char_repeat.length
+#define Loclength    F->fields.char_repeat.oclength
+#define Lstart_eptr  F->fields.char_repeat.start_eptr
+#define Lcharptr     F->fields.char_repeat.charptr
+#define Lmin         F->fields.char_repeat.min
+#define Lmax         F->fields.char_repeat.max
+#define Lc           F->fields.char_repeat.c
+#define Loc          F->fields.char_repeat.oc
```
NWilson, line 1302: Do you know why some of the defines are "Lsomething" and some are "Fsomething"? Is it just inconsistent naming?

src/pcre2_match.c
```diff
@@ -6451,14 +6456,14 @@
       pattern. */
 
       case OP_ASSERT_SCS:
-      F->temp_sptr[0] = mb->end_subject;
-      mb->end_subject = P->temp_sptr[0];
-      mb->true_end_subject = mb->end_subject + P->temp_size;
-      Feptr = P->temp_sptr[1];
+      F->fields.op_assert_scs.saved_end_subject = mb->end_subject;
+      mb->end_subject = P->fields.op_assert_scs.saved_end_subject;
```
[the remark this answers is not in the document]
NWilson: We can just ignore this. Clang's static analyzer is usually very helpful, but it has some spammy false positives in this file, due to the goto-based architecture. It thinks it's possible for there to be NULLs, because it can't follow all the conditions across the gotos.

NWilson (approved):
I like it! Better names is good.

src/pcre2_match.c
zherczeg: I suspect L is local, and F is frame. So L-s may or many not present in a frame, but F-s are always present and have the same meaning. Of course only @PhilipHazel knows the exact meaning. :)

PhilipHazel: @Zherczeg is correct. The naming is definitely deliberate. There is this comment in the code:

/* Define short names for general fields in the current backtrack frame, which
is always pointed to by the F variable. Occasional references to fields in
other frames are written out explicitly. There are also some fields in the
current frame whose names start with "temp" that are used for short-term,
localised backtracking memory. These are #defined with Lxxx names at the point
of use and undefined afterwards. */

zherczeg: I have updated the comments.

PR #849, opened 2025-11-21
Reduce the memory consumption of heapframe

zherczeg:
Heavyweight optimization patch for the interpreter. The `heapframe` size is reduced by 16 bytes.


zherczeg:
If no plans for review, I will land this patch.

NWilson:
Of course, thanks Zoltan. Go ahead!

PR #852, opened 2025-11-27
Support possessive repeats for backreferences

zherczeg:


NWilson:
Quite happily, it seems that the study code (find_minlength) already supports `OP_CRPOS.*` after `OP_(DN)?REFI?`

zherczeg:
Thank you for the review!

PR #858, opened 2025-12-02
Add support for partial matching to pcre2_substitute

### LINKED ISSUE #846: Can't use PARTIAL_HARD matches with SUBSITUTE_MATCHED
The new checks introduced in #807 seem to be a bit too strict.

If I pass `PCRE2_PARTIAL_HARD` to `pcre2_match`, I can't any longer use the match with `pcre2_substitute` with `PCRE2_SUBSTITUTE_MATCHED`, because if I do pass `PCRE2_PARTIAL_HARD` in my options, I get the error `PCRE2_ERROR_BADOPTION` (fair enough!) whereas if I don't pass it, I get `PCRE2_ERROR_DIFFSUBSOPTIONS`.

Should `PCRE2_PARTIAL_HARD` (and `PCRE2_PARTIAL_SOFT`) be excluded from the check for `PCRE2_ERROR_DIFFSUBSOPTIONS`, just like `PCRE2_NO_UTF_CHECK` in the recent commit 574f5683?

NWilson:
Fixes #846

### DIFF
--- a/testdata/testoutput2
+++ b/testdata/testoutput2
@@ -23671,14 +23671,18 @@ Failed: error -72: substitute subject differs from prior match call
 # --------------
 
 /(a)b+/
+\= Expect to fail with "bad option"
     ab\=ph,replace=FOO
 Failed: error -34: bad option value
+\= Expect to fail with "partial match"
     ab\=ph,substitute_replacement_only,replace=FOO
 Failed: error -2: partial match
+\= Expect success
     abc\=ph,substitute_replacement_only,replace=FOO
  1: FOO
     zabc\=ph,substitute_replacement_only,replace=>$&|$1|$`<
  1: >ab|a|z<
+\= Expect to fail with PCRE2_ERROR_PARTIALSUBS
     abc\=ph,substitute_replacement_only,replace=>$_<
 Failed: error -76 at offset 3 in replacement: replacement $' or $_ not supported with partial match
         here: >$_ |<--| <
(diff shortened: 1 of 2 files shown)


testdata/testoutput2
```diff
@@ -23666,6 +23666,44 @@ Failed: error -72: substitute subject differs from prior match call
     fooBAR
  1: X::textY
 
+# --------------
+# Tests for substitution with partial
+# --------------
+
+/(a)b+/
```
carenas, line 23673: It is usually nicer to sort the tests, so the ones that are meant to pass go first, and then the ones that are meant to fail are set after a comment:

```
\= Expect error
    ab\=ph,replace=FOO
...
```

NWilson: Thank you. I have reordered one test case, and labelled them with the expected outcomes.

rrthomas:
Thanks very much for this. It looks good on a read-through. The restrictions make sense. (I guess it would be possible if desired to reroute the match & substitution through code that works on the entire input if I wanted to support the use of `$_` and `$'`, having detected that condition with the `ERROR_PARTIALSUBS` error.)

I'll see if I can build my code against this branch now.

rrthomas:
The only thing I would question so far is the wording: relative to the current state, it's correct to say "Add support for partial matching", but relative to the *status quo ante* 10.47, it's more "Restore support for partial matching, now properly documented and implemented": after all, it worked fine before, though certain uses would have silently failed.

And I don't suppose I was the only person previously relying on this behavior!

rrthomas:
Yup, works fine for me. A thousand thanks!

NWilson:
Thank you for those comments, and confirming that it works.

Also, I do acknowledge that I did break your flow in 10.47, which had been previously working. I apologise again.  (It's possible however that no-one else had discovered this flow, due to the fact that it required PCRE2_SUBSTITUTE_MATCHED to allow partial matching.)

rrthomas:
It's less that I feel ignored (I don't!) and more that it might surprise future pedantic readers of the changelog (like me!) who end up confused and side-tracked by the idea that this functionality is brand new in 10.48.

I put "future me" firmly in this class of people, BTW, once I've forgotten this whole issue about 5 minutes after it gets fixed.

Issue #874, opened 2026-01-16
JIT matching failure with large patterns


6d7a:
### Issue

JIT-compiled patterns fail to match in PCRE2 10.45+ while non-JIT matching succeeds. The issue appears to be related to the `scan_prefix` function refactoring in commit `6f2da25f` ("Non-recursive scan prefix in JIT"), at least that is the earliest commit that prompts our test case to fail.

### Minimal Repro
The following test case demonstrates the issue:

```c
#define PCRE2_CODE_UNIT_WIDTH 8
#include <pcre2.h>
#include <stdio.h>
#include <string.h>

int main(void) {
    pcre2_code *re;
    pcre2_match_data *match_data;
    int errorcode;
    PCRE2_SIZE erroroffset;
    int rc;
    
    const char *pattern = 
        "(?:(?:(?i:(?P<_var_name>(channel|line)[\\w\\-.]{0,40}(secret|token)[\\w\\-.]{0,40})"
        "(?:(?:(\\\\\"(?![ \\t[\\]\"'()]{0,10}(,|>)))?[ \\t[\\]\"'()]{0,50}"
        "(?!(,|>)[ \\t[\\]\"'()]{0,10}\\\\\")))(?:(?::)|(?:=)|(?:\\|\\|)|(?:=>)|(?::=)|(?:<-)|"
        "(?:,)|(?:>)|(?:\\()|(?:[\\s+-]{0,250}[\\w.:-]{0,20}value=)|"
        "(?:[\\s+-]{0,250}[,\\n][\\s+-]{0,250}[\\w.:-]{0,20}(?:value|\"value\")[ \\t]{0,20}:)|"
        "(?:>[\\s+-]{0,250}<value>)|(?:\\.sendKeys\\ *\\()|"
        "(?:[ \\t]{0,50}=[ \\t]{0,50}\\n[ \\t]{0,50})|"
        "(?:[ \\t]{0,50}\\n[ \\t]{0,50}=[ \\t]{0,50}))"
        "(?:(?:(?:[ \\t[\\]\"'`()]{0,50}?([bru]{1,2}[\"']{1,3})?)(\\\\\")?)))"
        "(?P<apikey>[a-zA-Z0-9]{43})))(?![A-Za-z0-9!#(*.?[^~_+])";
    
    // invalid test API token
    const char *subject = 
        "@@ -0,0 +1,37 @@\n"
        "#include <WiFiClientSecure.h>\n"
[...]

PR #875, opened 2026-01-16
Fix repeat reset for scan prefix

zherczeg:
If this fixes the problem in #874, maybe this could be a candidate for maintenance patch.


zherczeg:
The static analyzer complains about non signed packages. Probably it needs some update.

src/pcre2_jit_compile.c
```diff
@@ -6249,6 +6249,7 @@ while (TRUE)
       repeat = GET2(cc, 1);
       if (repeat <= 0)
         {
+        repeat = 1;
```
carenas, line 6252: ~~considering that this code was released with 10.41, is the bug really a regression for 10.45?~~
nevermind, had to remember that we have a non linear history now:

```
$ git describe --tags 6f2da25f0
pcre2-10.44-145-g6f2da25f
$ git describe 6f2da25f0
pcre2-10.40-466-g6f2da25f
```

NWilson: I will recommend it for backporting as far back as the bug goes. The original report just says 10.45+, but if it affects earlier, I'll check the patch and see what's needed.

zherczeg:
I plan to land this patch. Does anybody has any objection?

NWilson:
I'm happy, I approved

src/pcre2_jit_compile.c
carenas: the original report was correct about the versions affected, but seem to had gotten the wrong changeset as a culprit.

the code that was patched was released with 10.45 (which is what 10.44 tag + 145 commits) means in the output of `git describe --tags`

PR #882, opened 2026-02-12
oss-fuzz: remove callout_count

carenas:
This could be used as an "alternative solution" for the issue referred in #881

Posted as a draft, as it hasn't been even tested and I have no access to that issue to confirm it actually "solves" it

### DIFF
--- a/src/pcre2_fuzzsupport.c
+++ b/src/pcre2_fuzzsupport.c
@@ -273,17 +273,6 @@ abort();
 #endif  /* SUPPORT_DIFF_FUZZ */
 #endif  /* SUPPORT_JIT */
 
-/* This is the callout function. Its only purpose is to halt matching if there
-are more than 100 callouts, as one way of stopping too much time being spent on
-fruitless matches. The callout data is a pointer to the counter. */
-
-static int callout_function(pcre2_callout_block *cb, void *callout_data)
-{
-(void)cb;  /* Avoid unused parameter warning */
-*((uint32_t *)callout_data) += 1;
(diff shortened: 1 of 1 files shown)


addisoncrump:
The issue that it "solves" was actually unrelated to the reported fix. I was unable to find the corresponding issue that it claimed to solve, actually, so I expect it was hallucinated or corresponds to a testcase we otherwise don't see.

addisoncrump:
@oliverchang Can you provide additional context as to _how_ #881 came about? I have made an assumption above, but I would hope that there is a corresponding issue you can point to.

NWilson:
The code you're removing here looks OK to me? If there is a crash, then we would ideally fix it rather than remove the test code.

NWilson:
For what it's worth, I wasn't able to reproduce this one.

(There was another OSS-Fuzz crash that's rather hard to exercise, on 32-bit CPUs only, and I have a fix I need to submit for that.)

addisoncrump (approved):
Looking over this, AFAICT this might increase the frequency of timeouts (though, presumably, the internal timeout mechanisms in PCRE2 have improved since this code was written) and we will not exercise callout functionality, which could be a testing blind spot but a fairly low-importance one. OSS-Fuzz will let us know if there are more issues, and this doesn't touch any user-facing code so there's no harm in experimenting a bit.

NWilson:
@carenas We couldn't reproduce, so I intend to close this. Thank you for your work trying to fix the issue.

If there still is a problem in the fuzzer's handling of callout_count, please open an Issue or re-open the PR.

PR #883, opened 2026-02-21
Fix propagation of generator expressions from parent directory

NWilson:
The CMake build would fail if invoked from a parent directory with a generator expression set on COMPILE_DEFINITIONS:

```
cmake_minimum_required(VERSION 3.15 FATAL_ERROR)

project(issue-813)

add_compile_definitions($<$<CONFIG:Debug>:DEBUG>)
add_subdirectory(pcre2)
```

This is a regression.

My code was overly broad; I propagated much more than I intended.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,14 +178,19 @@ check_include_file(windows.h HAVE_WINDOWS_H)
 pcre2_use_system_extensions()
 
 cmake_push_check_state(RESET)
-# Somewhat awkward code to propagate the _GNU_SOURCE definition (added to
-# COMPILE_DEFINITIONS by pcre2_use_system_extensions()).
+# Propagate the _GNU_SOURCE definition (added to COMPILE_DEFINITIONS by
+# pcre2_use_system_extensions()) so that these check_symbol_exists() calls
+# can find symbols only exposed with _GNU_SOURCE.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -178,13 +178,19 @@ check_include_file(windows.h HAVE_WINDOWS_H)
...
+# skip anything else (e.g. generator expressions inherited from a parent
+# project), because try_compile() cannot handle generator expressions.
 get_directory_property(_pcre2_compile_definitions COMPILE_DEFINITIONS)
 if(DEFINED _pcre2_compile_definitions)
   set(CMAKE_REQUIRED_DEFINITIONS "")
   foreach(_def IN LISTS _pcre2_compile_definitions)
-    list(APPEND CMAKE_REQUIRED_DEFINITIONS "-D${_def}")
+    if(_def STREQUAL "_GNU_SOURCE")
+      list(APPEND CMAKE_REQUIRED_DEFINITIONS "-D${_def}")
+    endif()
```
sergio-nsk: I would replace the `foreach` with `list(FIND _pcre2_compile_definitions "_GNU_SOURCE" _idx)`.

NWilson:
Thank you for reviewing. I have also added a CI test so we do not regress again on this particular behaviour.

PR #888, opened 2026-03-01
Fix memory leak on invalid blocksize in pcre2_deserialize_decode

cuiweixie:


src/pcre2_serialize.c
NWilson: It looks like there is the same leak here, too.

NWilson (changes requested):
Thank you very much for this contribution!

I'm not a big lover of "goto", but it might be useful here to use a "goto cleanup" pattern for the three branches that need to do cleanup.

There's a "unittest()" function in pcre2test_inc.h, where I have been slowly adding test cases and coverage for the remaining untested error branches.

Would it be possible to add test cases there, for these error returns?

If it helps, there is a coverage report in the CI job: https://github.com/PCRE2Project/pcre2/actions/runs/22279856593/artifacts/5607727786

I am happy for use of AI for adding test cases. (In fact, now that I have Claude code, I should let it go and fix all the missed lines in the coverage report!)

cuiweixie:
@NWilson  please help to trigger the  ci again.

src/pcre2_serialize.c
NWilson, line 243: I think your change has a double-free? You free the dst_re in the cleanup block, but also, the same object could already have been placed in `codes[i-1]`.

Unless I'm misreading, this suggests one missing testcase, and a need to do `dst_re = NULL` at the point when we assign `codes[i] = dst_re`, to transfer ownership and avoid the chance of a double-free.

Thank you very much, the change is otherwise looking good.

PR #893, opened 2026-03-11
pcre2test: detect setrlimit instead of hardcoding platform list

puffnfresh:
Replace the growing platform preprocessor check for -S flag support with AC_CHECK_FUNCS/check_symbol_exists detection of setrlimit. This automatically handles any platform lacking setrlimit (WASI, VMS, etc.) without maintaining a manual list.

Haiku and z/OS are kept as explicit exclusions because they provide setrlimit but don't support RLIMIT_STACK.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,6 +196,7 @@ endif()
 check_symbol_exists(mkostemp stdlib.h HAVE_MKOSTEMP) # glibc 2.7
 check_symbol_exists(memfd_create "sys/mman.h" HAVE_MEMFD_CREATE) # glibc 2.27
 check_symbol_exists(secure_getenv "stdlib.h" HAVE_SECURE_GETENV) # glibc 2.17
+check_symbol_exists(setrlimit "sys/resource.h" HAVE_SETRLIMIT)
 cmake_pop_check_state()
 
 check_c_source_compiles(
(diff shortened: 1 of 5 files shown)


NWilson:
Thank you, there are several of these in pcre2test and pcre2grep which I would love to clear up. Much appreciated! I will have to do a little bit of platform checking testing before I merge, but in principle, this is good.

NWilson:
Hi @puffnfresh.

Thank you very much for this WASI fix!

I would like to merge it. I have now had the chance to do some testing of my own.

I pushed a small fix to your branch, excluding WIN32 as before. This _may_ not be needed... but I have tried a few versions of MinGW, MSYS2, etc, and I'm nervous a regression is possible. Plus - the `#if HAVE_SETRLIMIT` block where you do the header include is only present on non-Win32.

Secondly, I seem not to have permissions to push merge commits to your branch, so I can't update this PR with a merge from main.

I would like to add the following to `build.zig`, however it will conflict unless the merge is done:

```zig
            .HAVE_SETRLIMIT = is_unix and !is_mingw,
```

NWilson (changes requested):
Needs merge with main, and build.zig update

NWilson:
I have merged this without the `build.zig` update. I will do that myself after merging.

NWilson:
Thank you again for this contribution!

PR #894, opened 2026-03-20
Update build.zig to compile on the latest stable Zig 0.15.2

bernardassan:
Add a named LazyPath called `pcre2.h`, which library users can import and add to their projects' include directory, like https://github.com/bernardassan/Glib/blob/49ef9727f43d41011baa1f3961076a29ebebb515/build/glib/root.zig#L93

Add some more config_h options

Correctly set PCRE2_EXPORT

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -12,6 +12,7 @@ pub fn build(b: *std.Build) !void {
     const rt = target.result;
 
     const linkage = b.option(std.builtin.LinkMode, "linkage", "whether to statically or dynamically link the library") orelse @as(std.builtin.LinkMode, if (rt.isGnuLibC()) .dynamic else .static);
+    const sanitize_c = b.option(std.zig.SanitizeC, "sanitize_c", "whether to build with undefined behaviour sanitizer enabled") orelse .off;
     const codeUnitWidth = b.option(CodeUnitWidth, "code-unit-width", "Sets the code unit width") orelse .@"8";
     const jit = b.option(bool, "support_jit", "Enable/disable JIT compiler support") orelse false;
 
@@ -39,18 +40,16 @@ pub fn build(b: *std.Build) !void {
             .HAVE_SYS_TYPES_H = true,
             .HAVE_UNISTD_H = is_unix or is_mingw,
             .HAVE_WINDOWS_H = rt.os.tag == .windows,
-            .HAVE_MKOSTEMP = is_unix,
+
             .HAVE_MEMFD_CREATE = is_musl or is_glibc or is_freebsd,
             .HAVE_SECURE_GETENV = is_musl or is_glibc or is_freebsd,
-            .HAVE_REALPATH = is_unix,
 
             // all compilation is using the Zig bundled c compiler
-            .HAVE_ATTRIBUTE_UNINITIALIZED = true,
-            .HAVE_VISIBILITY = true,
             .HAVE_BUILTIN_ASSUME = null,
             .HAVE_BUILTIN_MUL_OVERFLOW = true,
             .HAVE_BUILTIN_UNREACHABLE = true,
-            .INTEL_CET_ENABLED = null,
+            .HAVE_ATTRIBUTE_UNINITIALIZED = true,
+            .HAVE_VISIBILITY = true,
 
             .SUPPORT_PCRE2_8 = codeUnitWidth == .@"8",
             .SUPPORT_PCRE2_16 = codeUnitWidth == .@"16",
@@ -78,6 +77,7 @@ pub fn build(b: *std.Build) !void {
     const lib_mod = b.createModule(.{
         .target = target,
         .optimize = optimize,
+        .sanitize_c = sanitize_c,
         .link_libc = true,
     });
 
@@ -126,6 +126,9 @@ pub fn build(b: *std.Build) !void {
             "src/pcre2_ucd.c",
             "src/pcre2_valid_utf.c",
             "src/pcre2_xclass.c",
+        },
+        .flags = &.{
+            "-fvisibility=hidden",
         },
     });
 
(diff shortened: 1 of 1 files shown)


build.zig
```diff
@@ -7,50 +7,74 @@ pub const CodeUnitWidth = enum {
...
-
+            .HAVE_DIRENT_H = is_unix or is_mingw,
+            .HAVE_SYS_STAT_H = true,
+            .HAVE_SYS_TYPES_H = true,
+            .HAVE_UNISTD_H = is_unix or is_mingw,
+            .HAVE_WINDOWS_H = rt.os.tag == .windows,
+            .HAVE_MKOSTEMP = is_unix,
+            .HAVE_MEMFD_CREATE = is_musl or is_glibc or is_freebsd,
+            .HAVE_SECURE_GETENV = is_musl or is_glibc or is_freebsd,
+            .HAVE_REALPATH = is_unix,
```
NWilson: Several of these added options don't need to be set.

Some are just added unavoidably by autoconf.

The list of those we actually use is in config-cmake.h.in. Indeed, that's the file that we've configuring here, so there's no need to set options not present in that file.

So the list here should have exactly the same as config-cmake.h.in.

Thank you though for adding some that were missed out.

build.zig
```diff
@@ -59,25 +83,18 @@ pub fn build(b: *std.Build) !void {
...
     lib_mod.addCMacro("PCRE2_CODE_UNIT_WIDTH", @tagName(codeUnitWidth));
-    if (linkage == .static) {
-        lib_mod.addCMacro("PCRE2_STATIC", "");
+    switch (linkage) {
+        .static => lib_mod.addCMacro("PCRE2_STATIC", ""),
+        .dynamic => {},
     }
+    lib_mod.addConfigHeader(config_h);
+    lib_mod.addIncludePath(pcre2_h_dir.getDirectory());
+    lib_mod.addIncludePath(b.path("src"));
```
NWilson, line 100: Just to confirm, I don't know the implications of adding a directory to the lib_mod's includes.

Does that affect all downstream consumers? Could you test or confirm whether clients using PCRE2 will inherit this include path? We have had that problem in other build systems - for example Bazel, where we by mistake added `src/` to the list of include paths that library clients would search.

build.zig
```diff
@@ -7,50 +7,74 @@ pub const CodeUnitWidth = enum {
...
-            .SUPPORT_PCRE2_32 = codeUnitWidth == CodeUnitWidth.@"32",
+            .SUPPORT_PCRE2_8 = codeUnitWidth == .@"8",
+            .SUPPORT_PCRE2_16 = codeUnitWidth == .@"16",
+            .SUPPORT_PCRE2_32 = codeUnitWidth == .@"32",
             .SUPPORT_UNICODE = true,
             .SUPPORT_JIT = jit,
 
-            .PCRE2_EXPORT = null,
+            .PCRE2_EXPORT = switch (linkage) {
+                .dynamic => "__attribute__ ((visibility (\"default\")))",
```
NWilson: Does this make any difference? In C, this makes no difference unless you are actually passing `-fvisibility=hidden` to the compiler.

I see this: https://github.com/ziglang/zig/issues/9762

Do you observe an effect when you set PCRE2_EXPORT like this? Does the resulting library actually contain a different set of symbols?

build.zig
```diff
@@ -146,47 +177,33 @@ pub fn build(b: *std.Build) !void {
...
+
+        const posixLib = b.addLibrary(.{
+            .name = "pcre2-posix",
+            .root_module = posixLib_mod,
+            .linkage = linkage,
+        });
+
+        pcre2test_mod.linkLibrary(posixLib);
 
         posixLib.installHeader(b.path("src/pcre2posix.h"), "pcre2posix.h");
```
NWilson, line 227: Should the POSIX lib also use the addNamedLazyPath mechanism?

NWilson:
Thank you for this contribution! I have raised some questions, but in principle, if it builds for you and on our CI, I am happy to accept it.

build.zig
bernardassan: I have remove `HAVE_MKOSTEMP`, `INTEL_CET_ENABLED` and `HAVE_REALPATH` which aren't in config-cmake.h.in

build.zig
bernardassan: No, the includes are private to the module and don't affect downstream consumers. In the zig build system, you have to manually export the include directory using a named `LazyPath`/`WriteFile` Step for it to be accessible to downstream users. And the users must manually opt in to using the exported directories by importing them using `namedLazyPath`/`namedWriteFiles` fn and adding it to their modules include directories using `addIncludePath` for it to be used in their project so nothing happens automagically

build.zig
bernardassan: Indeed, `-fvisibility=hidden` is required for it to take effect, and adding it changes the symbols set.

build.zig
bernardassan: I don't need `pcre2posix.h`, that's why I didn't export it. I am a bit averse to adding it because of the recommendation to use the native API https://pcre2project.github.io/pcre2/doc/pcre2posix/

build.zig
bernardassan: Let me throughly test my project with this change and see if its worth it then PR it

bernardassan:
I will update this PR in the morning, done some changes already

bernardassan:
Updated PR with suggestions from the review

bernardassan:
@NWilson, can you have a look at the improvements made in your free time?

NWilson:
Thank you again @bernardassan. I have pushed one small change to set the flags consistently across the different zig modules.

I tested with `nm -D` on the shared library produced by zig and observed that the `_pcre2_default_tables` symbol was public, because the `-fvisibility` flag was only set on one of the addCSourceFiles calls.

That led to me passing the flags consistently to all of them.

I will merge your changes + my small addition.

Please do feel free to come back with any further changes if you object or disagree with any of the tweaks I made (I am, after all, a C developer not a Zig developer).

bernardassan:
@NWilson looks good to me. Maybe you will become a Zig developer someday in the future 😄.

PR #901, opened 2026-04-29
Fix bounds checks for truncated UTF-8 in pattern conversion

TristanInSec:
Add bounds validation after GETCHARLENTEST/GETCHARINCTEST in convert_posix() and convert_glob_parse_range() to detect when a multi-byte UTF-8 sequence extends past the end of the pattern buffer. Return PCRE2_ERROR_UTF8_ERR1 when remaining bytes are insufficient for the declared sequence length.

Five call sites are affected: one in convert_posix() (GETCHARLENTEST) and four in convert_glob_parse_range() (GETCHARINCTEST). Without these checks, a truncated UTF-8 lead byte at the end of a pattern causes reads past the buffer boundary when using pcre2_pattern_convert() with PCRE2_CONVERT_UTF.

Found while writing a fuzz harness for pcre2_pattern_convert().

### DIFF
--- a/src/pcre2_convert.c
+++ b/src/pcre2_convert.c
@@ -192,6 +192,11 @@ while (plength > 0)
   c = *posix;
 #else
   GETCHARLENTEST(c, posix, clength);
+  if ((PCRE2_SIZE)clength > plength)
+    {
+    *bufflenptr = plength;
+    return PCRE2_ERROR_UTF8_ERR1;
+    }
 #endif
   posix += clength;
   plength -= clength;
@@ -704,6 +709,12 @@ while (pattern < pattern_end)
   char_start = pattern;
   GETCHARINCTEST(c, pattern);
 
+  if (pattern > pattern_end)
+    {
+    *from = pattern_end;
+    return PCRE2_ERROR_UTF8_ERR1;
+    }
+
   if (c == CHAR_RIGHT_SQUARE_BRACKET)
     {
     convert_glob_write(out, c);
@@ -752,12 +763,24 @@ while (pattern < pattern_end)
     char_start = pattern;
     GETCHARINCTEST(c, pattern);
 
+    if (pattern > pattern_end)
+      {
+      *from = pattern_end;
+      return PCRE2_ERROR_UTF8_ERR1;
+      }
+
     if (pattern >= pattern_end) break;
 
     if (escape != 0 && c == escape)
       {
       char_start = pattern;
       GETCHARINCTEST(c, pattern);
+
+      if (pattern > pattern_end)
+        {
+        *from = pattern_end;
+        return PCRE2_ERROR_UTF8_ERR1;
+        }
       }
     else if (c == CHAR_LEFT_SQUARE_BRACKET && *pattern == CHAR_COLON)
       {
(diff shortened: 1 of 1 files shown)


NWilson:
Thank you very much! I will look at this as soon as I can, as this is potentially of high severity.

zherczeg:
Is this a new feature patch?

TristanInSec:
This is a bug fix, not a new feature. pcre2_convert() can read past the end of the pattern buffer when a truncated UTF-8 sequence sits at the boundary. The existing length checks don't account for multi-byte sequences that extend beyond plength. The patch adds bounds checks to prevent the overread.

zherczeg:
The current operation is intentional, changing it is a new feature. Question is do we want it?

NWilson:
This pull request has no associated tests.

The functions you have modified appear to be guarded by UTF-8 validation - so is this PR like the other one, where the bug only occurs if malformed input is provided with PCRE2_NO_UTF_CHECK?

TristanInSec:
@NWilson No, this is different from PR #902.

`pcre2_pattern_convert()` has no `PCRE2_NO_UTF_CHECK` equivalent. The function API does not offer any UTF-8 validation step. An application that accepts POSIX or glob patterns from external input and passes them to `pcre2_pattern_convert()` has no way to request validation before conversion.

The out-of-bounds read triggers on any truncated UTF-8 sequence at the end of the pattern buffer, with no flag needed to disable checks, because there are no checks to disable.

@zherczeg Regarding "intentional": the current behavior for valid UTF-8 input is unchanged by this patch. The patch only affects the case where the pattern ends with a truncated multi-byte sequence (e.g., a lead byte with missing continuation bytes). In that case, the GETCHARLENTEST macro reads past the end of the buffer. The fix returns PCRE2_ERROR_UTF8_ERR1, consistent with how pcre2_compile handles the same condition.

I can add tests. Would you prefer tests in the existing pcre2test framework, or standalone?

NWilson:
> pcre2_pattern_convert() has no PCRE2_NO_UTF_CHECK equivalent. The function API does not offer any UTF-8 validation step. An application that accepts POSIX or glob patterns from external input and passes them to pcre2_pattern_convert() has no way to request validation before conversion.

Really? I think it does.

If you have reproduction steps for a crash, which does not depend on passing PCRE2_CONVERT_NO_UTF_CHECK, then I would like to see them, ideally in the form of the existing pcre2test framework.

However, I do doubt that it is possible.

TristanInSec:
@NWilson You are correct. I missed the `PCRE2_CONVERT_NO_UTF_CHECK` / `valid_utf()` gate at the top of `pcre2_pattern_convert()`. The out-of-bounds read does require `PCRE2_CONVERT_NO_UTF_CHECK` to be set, same as PR #902.

This is a defense-in-depth fix: the GETCHARLENTEST/GETCHARINCTEST macros should not read past the buffer boundary even when UTF validation is skipped. But I acknowledge the severity is lower than I stated. Apologies for the incorrect characterization.

I will add pcre2test-style reproduction tests that demonstrate the overread with `PCRE2_CONVERT_NO_UTF_CHECK`. Would you still like the bounds checks, given the reduced severity?

NWilson:
If there is no bug when the flag is not passed, then we will keep the code as it is. 

Thank you for your offer and investigation.

PR #902, opened 2026-04-29
Add bounds checks for truncated/malformed UTF-8 in 8-bit mode

TristanInSec:
When PCRE2_NO_UTF_CHECK is used with pcre2_match, pcre2_dfa_match,
or pcre2_substitute, truncated UTF-8 sequences at the end of the
subject can cause GETCHAR macros to read past the subject buffer.
Malformed sequences within the buffer can also produce codepoint
values above 0x10FFFF, which then index out of bounds in the
Unicode property tables.

This patch adds:

1. A `UTF8CLENTOOLONG` helper macro in pcre2_internal.h that checks
   whether a lead byte's implied sequence length would exceed the
   remaining buffer.

2. Pre-GETCHAR truncation checks using this macro before all subject
   reads in the match engine, DFA engine, script run checker,
   extended grapheme handler, substitute function, and xclass handler.

3. Post-GETCHAR codepoint range checks (c > 0x10FFFF) before
   UCD property table lookups, to guard against malformed sequences
   that assemble garbage codepoints within buffer bounds.

4. Bounded BACKCHAR scans to prevent underflow past the subject
   start when the subject begins with continuation bytes (0x80-0xBF).

Affected files: pcre2_match.c, pcre2_dfa_match.c, pcre2_xclass.c,
pcre2_extuni.c, pcre2_script_run.c, pcre2_substitute.c.

### DIFF
--- a/src/pcre2_dfa_match.c
+++ b/src/pcre2_dfa_match.c
@@ -732,7 +732,16 @@ for (;;)
     {
     clen = 1;        /* Number of data items in the character */
 #ifdef SUPPORT_UNICODE
-    GETCHARLENTEST(c, ptr, clen);
+    if (utf && UTF8CLENTOOLONG(ptr, end_subject))
+      {
+      clen = 0;
+      c = NOTACHAR;
+      }
+    else
+      {
+      GETCHARLENTEST(c, ptr, clen);
+      if (c > 0x10ffffu) { clen = 0; c = NOTACHAR; }
+      }
 #else
     c = *ptr;
 #endif  /* SUPPORT_UNICODE */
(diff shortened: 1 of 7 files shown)


NWilson:
This one we will not fix. 

The documentation for PCRE2_NO_UTF_CHECK is clear: undefined behaviour results of you say the data is valid and it isn't, and turn off checks. 

This is the application's fault. 

I sympathise with the concern, and if we were doing it again, I would probably not offer this option myself. But I don't see a strong need to remove it now.

zherczeg:
PCXRE2 has two modes: a faster for trusted utf input and a slower for untrusted.

TristanInSec:
Understood, and I respect the position on the API contract. Just one note for the record: other widely-deployed C libraries (OpenSSL, zlib, libpng) have moved toward hardening even documented-UB paths, since callers routinely pass network-sourced data through performance flags they believe are safe. The bounds checks here add minimal overhead to the fast path. If the position changes down the line, the PR is there. Thanks for considering it.

PhilipHazel:
The use case for adding PCRE2_NO_UTF_CHECK (long ago) is when the same subject string is being scanned many times, either for many different patterns or for finding all matches of one pattern. If the subject is very long - megabytes or more; this is not unknown - the performance improvement can be substantial. However, like all tools, this feature can be misused.

TristanInSec:
Thank you for the detailed explanation, @PhilipHazel and @NWilson. The performance use case makes sense. I will close this PR.

PR #907, opened 2026-05-03
Support for direct installation via `zig fetch`

mrdimidium:
Modern versions of zig (at least 0.15 and above) support a small package manager. This PR adds a manifest to the root directory and updates the build so that the manifest is included in the release archives.

I also took the liberty of updating the zig installation in CI to a reusable action from one of the zig core team members. This will prevent compiler version desynchronization and allow you to add builds to master versions if you wish. Let me know if this is unacceptable -- I'll return the snap or offer installation in ziglang.org with a small bash script (but with worse caching).

P.S. Thanks for supporting Zig here, it really made my job easier.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,7 +1,11 @@
 .{
     .name = .pcre2,
     .version = "10.48.0-DEV",
-    .fingerprint = 0x914b3d63ebab9e21, // Changing this has security and trust implications.
+
+    // Do NOT change this for new releases.
+    // DO change this if making a fork of PCRE2 which is used by any Zig clients.
+    .fingerprint = 0x914b3d63ebab9e21,
+
     .minimum_zig_version = "0.15.2",
     .dependencies = .{},
     .paths = .{
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -1,0 +1,14 @@
+.{
+    .name = .pcre2,
+    .version = "10.48.0-DEV",
+    .fingerprint = 0x914b3d63ebab9e21, // Changing this has security and trust implications.
+    .minimum_zig_version = "0.15.2",
+    .dependencies = .{},
+    .paths = .{
+        "build.zig",  "build.zig.zon",
+        "src/",       "doc/",
+        "HACKING",    "INSTALL",
+        "README.md",  "LICENCE.md",
+        "AUTHORS.md", "SECURITY.md",
+    },
+}
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -9,10 +9,11 @@
     .minimum_zig_version = "0.15.2",
     .dependencies = .{},
     .paths = .{
-        "build.zig",  "build.zig.zon",
-        "src/",       "doc/",
-        "HACKING",    "INSTALL",
-        "README.md",  "LICENCE.md",
-        "AUTHORS.md", "SECURITY.md",
+        "build.zig",   "build.zig.zon",
+        "src/",        "doc/",
+        "deps/",       "HACKING",
+        "INSTALL",     "README.md",
+        "LICENCE.md",  "AUTHORS.md",
+        "SECURITY.md",
     },
 }
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -634,9 +634,9 @@ jobs:
       (github.event_name == 'workflow_dispatch' && (inputs.job_id == 'all' || inputs.job_id == 'zebrilus')) ||
       github.event_name == 'push'
     steps:
-      - name: Setup
-        run: |
-          sudo snap install zig --classic --beta
+      # The zig version from minimum_zig_version in build.zig.zon will be installed.
+      - name: Install zig
+        uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
 
       - name: Checkout
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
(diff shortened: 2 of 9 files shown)


build.zig.zon
```diff
@@ -0,0 +1,14 @@
+.{
+    .name = .pcre2,
+    .version = "10.48.0-DEV",
+    .fingerprint = 0x914b3d63ebab9e21, // Changing this has security and trust implications.
```
NWilson: Could you include instructions on how to update this? Is it done automatically, and on what schedule?

NWilson (approved):
This looks fine, thank you. All the zig support is simply as submitted by other zig users, so if this is useful, then I'm happy to host it. 

I will test it works on my machine and in the co jobs before merging. I like to be able to build the different platforms myself.

build.zig.zon
mrdimidium: It's never updated. It's a unique package ID, used to avoid collisions with identical names. The only scenario where you'd want to update it is when forking the project.

mrdimidium: In more detail, zig doesn't offer a package registry; packages are installed directly via HTTP (for example, from Github releases). A large, unique ID identifies the package, so changing it could indicate an attempt to impersonate a different package. The comment states that no PR should ever change it.

build.zig.zon
```diff
@@ -0,0 +1,14 @@
...
+    .minimum_zig_version = "0.15.2",
+    .dependencies = .{},
+    .paths = .{
+        "build.zig",  "build.zig.zon",
+        "src/",       "doc/",
+        "HACKING",    "INSTALL",
+        "README.md",  "LICENCE.md",
+        "AUTHORS.md", "SECURITY.md",
```
NWilson: I think you want to include some additional directories here:

Possibly testdata and RunTest(.bat)

Probably deps, which includes source code for the JIT compiler, which is a supported option in the Zig build.

NWilson: It looks like zig fetch supports Git submodules, but I don't know how exactly the hash would be calculated based on the hashed contents here.

If you included the `deps/` directory (ie the Git submodules) in the package hash, then the package importing pcre2 would calculate a different hash depending on whether or not the submodules were specified to be fetched or not.

build.zig.zon
```diff
@@ -0,0 +1,18 @@
+.{
+    .name = .pcre2,
+    .version = "10.48.0-DEV",
+
+    // Do NOT change this for new releases.
+    // DO change this if making a fork of PCRE2 which is used by any Zig clients.
```
NWilson, line 6: I made this comment more prescriptive to tell myself or other maintainers what to do with this field.

Let me know if these instructions are wrong.

build.zig.zon
mrdimidium: Interesting. I didn't include deps, expecting that if the user needed jit, they would install it separately and link it into Zig, but now I see that deps is included as files in the release archives.

I expect zig to calculate the hash based on the contents, but this will still lead to different results when installing from a release and from a repository.

build.zig.zon
mrdimidium: That's right, thank you.

build.zig.zon
mrdimidium: The most idiomatic way to specify juajit is through build.zig.zon . I'll try to suggest `build.zig` to the upstream author if they're willing to accept it.

mrdimidium: @NWilson, Please take a look at [this commit](https://github.com/PCRE2Project/pcre2/pull/907/changes/ea8ea9430686227c0e7f7566f1aeee4497a5f288). I explicitly installed the dependency from the same commit so it would work with git, HTTP, and any commit archive.

[The file copying trick](https://github.com/PCRE2Project/pcre2/pull/907/changes#diff-f87bb3596894756629bc39d595fb18d479dc4edf168d93a911cadcb060f10fccR154) is necessary because the dependency is installed somewhere in zig-pkg, and we copy the file so it finds `../deps/...`. This could be done more carefully by passing the path to the preprocessor, but I was hesitant to change the code for the sake of a zig build.

The downside is that you need to update the commit in the submodule and in build.zig.zon simultaneously.

NWilson: I would suggest that perhaps you don't want to use sljit as a Zig dependency, since it's not a Zig library. It really is part of PCRE2 (effectively), and it's just stored as a Git submodule for convenience.

If you want to use tarballs, then use a release tarball from GitHub Releases. If you want to use Git to checkout a tag, that's also fine. But I would dissuade you from using GitHub's per-commit tarballs, because they are not signed or reproducible (unlike our signed release tarballs and signed Git release tags).

mrdimidium: That makes sense. I replaced the dependency with an explicit inclusion of deps. I'm not sure the library user needs the test data and should depend on it. Am I wrong?

NWilson: Personally, I would always recommend running a library's test suite after building it from source. Maybe I'm paranoid! The pcre2test binary is built by Zig and can be used to verify the library functions, so I'd suggest you include the tests.

But I don't mind either way. I don't think the tests are hooked up to run automatically after the Zig build, so a Zig user could delete the testdata and not notice anything.

mrdimidium: Frankly, I'd be surprised by the tests running during builds, especially with cross-compilation.

I'd be happy to offer more build options in CI (at least cross-compilation and multiple compiler versions) if you're open to the extra workload.

NWilson: I don't need it - if you're happy for the tests not to be run by Zig, that's fine by me.

Don't do anything extra for me.

NWilson:
Thank you again Nikolay. Apologies for the delay merging.

PR #909, opened 2026-05-12
Use CU2BYTES for byte sizing in two allocation sites

iliaal:
Two allocation sites multiply by `PCRE2_CODE_UNIT_WIDTH` (the bit width: 8, 16, or 32) where the byte-count helper `CU2BYTES(x) = (x) * (PCRE2_CODE_UNIT_WIDTH/8)` is intended. The result over-allocates by the code-unit byte width: 8x in 8-bit mode, 16x in 16-bit, 32x in 32-bit.

Subsequent `memcpy` calls already use `CU2BYTES` correctly, so no out-of-bounds write occurs. The over-allocation leaks the difference until the buffer is freed, on every call.

Sites:
- `src/pcre2_substring.c:214` (`pcre2_substring_get_bynumber`)
- `src/pcre2_convert.c:1219` (`pcre2_pattern_convert`)

Repo-wide grep for `* PCRE2_CODE_UNIT_WIDTH` outside preprocessor conditionals turned up only these two sites. Found via internal audit of PCRE2 10.47.

Verified clean against the existing test suite via `cmake` build + `RunTest` (29 tests, all pass). Tests 24 and 25 exercise `pcre2_pattern_convert`; substring tests in test 1 exercise `pcre2_substring_get_bynumber`.

### DIFF
--- a/src/pcre2_convert.c
+++ b/src/pcre2_convert.c
@@ -1215,9 +1215,11 @@ for (int i = 0; i < 2; i++)
   /* Allocate memory for the buffer, with hidden space for an allocator at
   the start. The next time round the loop runs the conversion for real. */
 
-  allocated = PRIV(memctl_malloc)(sizeof(pcre2_memctl) +
-    (*bufflenptr + 1)*PCRE2_CODE_UNIT_WIDTH, (pcre2_memctl *)ccontext);
-  if (allocated == NULL)
+  if (*bufflenptr > ((PCRE2_SIZE_MAX - sizeof(pcre2_memctl)) /
+        CU2BYTES(1)) - 1 ||
+      (allocated = PRIV(memctl_malloc)(sizeof(pcre2_memctl) +
+        CU2BYTES(*bufflenptr + 1),
+        (pcre2_memctl *)ccontext)) == NULL)
     {
     *bufflenptr = 0;  /* Error offset */
     return PCRE2_ERROR_NOMEMORY;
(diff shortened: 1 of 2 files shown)


NWilson:
Thank you! This looks legitimate.

If you used AI, would you mind sharing what tool (and prompt) you used?

I will confirm it manually myself.

I'm trying to get a bit better on hygiene around integer overflows - so ideally we would improve the code further:

```
  if (*bufflenptr > ((PCRE2_SIZE_MAX - sizeof(pcre2_memctl)) /
        CU2BYTES(1)) - 1 ||
      (allocated = PRIV(memctl_malloc)(sizeof(pcre2_memctl) +
        CU2BYTES(*bufflenptr + 1),
        (pcre2_memctl *)ccontext)) == NULL)
    {
    *bufflenptr = 0;  /* Error offset */
    return PCRE2_ERROR_NOMEMORY;
    }
```

(In fact, I clearly need a helper for allocate-with-multiply-check. But that's for another PR.)

Would you be able to add the check to both sites you've improved?

iliaal:
Yeah, I'll take a look.

AI was used to draft PR & validate finding when looking at something else. For reference: https://github.com/iliaal/whetstone/tree/main/plugins/whetstone/skills/ia-code-review
https://github.com/iliaal/whetstone/tree/main/plugins/whetstone/skills/ia-meta-prompting (Adversarial)

Were used across multiple LLM (Claude, Codex, Deepseek v4 Pro) cross checks

iliaal:
Added the guard at both sites.

NWilson:
Those are some very interesting skills! I have access to Claude & GPT. I would be interested in doing a more thorough audit of all of PCRE2... at some point. I'll bear in mind that your tool might be useful.

PR #910, opened 2026-05-12
Widen class_ranges::range_list_size from uint16_t to uint32_t

iliaal:
`class_ranges::range_list_size` is computed in `pcre2_compile_class.c` as a `size_t` from `parse_class()`, then stored into the struct as `uint16_t` via three explicit narrowing casts. The underlying allocation at line 542 uses the full `size_t` value, so no out-of-bounds write occurs. Downstream consumers (e.g. `compile_class_not_nested` reading `cranges->range_list_size - 1` and `cranges->range_list_size - 2` at lines 1150-1151) walk fewer entries than were allocated when the count exceeds 65535, producing wrong match results.

Reachable on 16-bit or 32-bit lib builds with `LINK_SIZE >= 3`, where `MAX_PATTERN_SIZE` grows enough for a single character class to produce more than 65535 sorted-range entries. Not reachable on the default 8-bit lib with `LINK_SIZE = 2`. Discovered during an internal audit of PCRE2 10.47.

Widening the struct field to `uint32_t` and updating all three narrowing casts:

- `src/pcre2_intmodedep.h:781` field declaration
- `src/pcre2_compile_class.c:552, 621, 743` producer casts

The neighbouring `char_lists_types` field stays `uint16_t` (it holds flag bits, not a count).

No new test added. Reaching the trigger requires more than 65535 sorted-range entries in a single character class, which would inflate `testdata/testinput*` significantly and only run in non-8-bit lib mode. Happy to add a test if reviewers want one. Existing `RunTest` suite verified clean against the patch (29 tests, all pass).

### DIFF
--- a/src/pcre2_compile_class.c
+++ b/src/pcre2_compile_class.c
@@ -549,7 +549,7 @@ cranges->header.next = NULL;
 #ifdef PCRE2_DEBUG
 cranges->header.type = CDATA_CRANGE;
 #endif
-cranges->range_list_size = (uint16_t)range_list_size;
+cranges->range_list_size = range_list_size;
 cranges->char_lists_types = 0;
 cranges->char_lists_size = 0;
 cranges->char_lists_start = 0;
@@ -618,7 +618,7 @@ ptr = buffer;
 while (ptr < dst && ptr[1] < 0x100) ptr += 2;
 if (dst - ptr < (2 * (6 - 1)))
   {
-  cranges->range_list_size = (uint16_t)(dst + 2 - buffer);
+  cranges->range_list_size = (size_t)(dst + 2 - buffer);
   return cranges;
   }
 
@@ -740,7 +740,7 @@ PCRE2_ASSERT((uint16_t*)dst <= next_char);
 cranges->char_lists_size =
   (size_t)((uint8_t*)(buffer + total_size) - (uint8_t*)next_char);
 cranges->char_lists_start = (size_t)((uint8_t*)next_char - (uint8_t*)buffer);
-cranges->range_list_size = (uint16_t)(dst - buffer);
+cranges->range_list_size = (size_t)(dst - buffer);
 return cranges;
 }
 
--- a/src/pcre2_intmodedep.h
+++ b/src/pcre2_intmodedep.h
@@ -778,7 +778,7 @@ typedef struct class_ranges {
   compile_data header;             /* Common header */
   size_t char_lists_size;          /* Total size of encoded char lists */
   size_t char_lists_start;         /* Start offset of encoded char lists */
-  uint16_t range_list_size;        /* Size of ranges array */
+  size_t range_list_size;          /* Size of ranges array */
   uint16_t char_lists_types;       /* The XCL_LIST header of char lists */
   /* Followed by the list of ranges (start/end pairs) */
 } class_ranges;


NWilson:
Hmm. This is delicate code. I'll have to check it manually later in the week (when I have time).

Thank you for the contribution.

I think a test would be very helpful. There are test files that are only run for given LINK_SIZE and bitmodes. It would be good to see a genuine pattern that exhibits a match error due to this truncation.

If the uint16_t can overflow, do you believe that the uint32_t could overflow too? (It would take gigabytes of test input to trigger it, I presume.) The natural fix may be a size_t.

iliaal:
Will work on a test, it seemed to me uint32 would've be enough, but perhaps size_t is safer, let me dig into this further

iliaal:
Pushed the size_t widening per your suggestion. The test investigation produced something I didn't expect.

`cranges->range_list_size` is written at three sites: 552 (pre-merge from `parse_class`), 622 (small-tail return), 745 (post-char_lists return). It's read at 1150 and 1585, both downstream of `compile_optimize_class`.

The 552 store can exceed 65535 (probed: 65540 for a class of 32770 disjoint codepoints). But it's only the final stored value when the function early-returns via 558 (`range_list_size == 0`) or 564 (`<= 2`). Any larger pre-merge value falls through to merging + char_lists conversion, and the field is overwritten at 622 or 745. Both are bounded by the ASCII residual (max 128 disjoint ranges = 256 entries), since char_lists consumes everything `>= 0x100`.

With my probe pattern the reader at 1150 sees 0 (every range migrated to char_lists), regardless of whether the field is `uint16_t`, `uint32_t`, or `size_t`. So the original PR description is wrong: no pattern demonstrates a match error from this truncation in current code. The size_t widening is defensive only, removing a cast at 552 that would matter only if a reader were inserted between 552 and the 622/745 overwrite.

Up to you whether to merge as defensive cleanup or close.

zherczeg:
I made this code, but I don't remember the exact strategy. I am sure I chose uint16 for a reason, but don't remember the details. The idea was something, that from 0-65536 the maximum number of ranges is 65536/2: [0..0] [2..2] [4..4] [6..6]...

iliaal:
> I made this code, but I don't remember the exact strategy. I am sure I chose uint16 for a reason, but don't remember the details. The idea was something, that from 0-65536 the maximum number of ranges is 65536/2: [0..0] [2..2] [4..4] [6..6]...

The reality even uint16_t works because of prior checks, as per earlier note, so this change is **purely defensive** and realistically not even needed given current work flow.

Update: worth nothing the obvious (albeit minor) memory cost of switching to uint32 or size_t

zherczeg:
There are two uint16_t values, which from a 32 bit value. If we change the size of the first uint16_t value, an extra padding is also added after the second value.

zherczeg:
The ascii case, I think I remember now. Ascii ranges was handled separately, and you are right, max 128 range is present. The other ranges are compressed into a rather complex data structure.

NWilson:
> With my probe pattern the reader at 1150 sees 0 (every range migrated to char_lists), regardless of whether the field is `uint16_t`, `uint32_t`, or `size_t`. So the original PR description is wrong: no pattern demonstrates a match error from this truncation in current code. The size_t widening is defensive only, removing a cast at 552 that would matter only if a reader were inserted between 552 and the 622/745 overwrite.
> 
> Up to you whether to merge as defensive cleanup or close.

Thank you for clarifying.

As I suspected, Zoltan did get it right initially, and the cast "should" be safe. If there isn't an assert at the point of cast, there certainly should be - and then the fuzzers would detect any large inputs that violate the assert, in future.

I am happy with the widening though, even if it's not needed.

Choices:
* Keep uint16_t but ensure debug assert validates no overflow
* Widen to size_t even if current code does not require it

I think I should take some time to refresh my mind on the data structures and make a decision (or Zoltan can advise and I'll go with his choice, if he has one).

NWilson:
I'm afraid this one also is going to be bumped to the next release. My apologies.

NWilson:
I have reviewed this code again over the last week.

I do not believe any of the casts can overflow.

Thank you @iliaal for this PR; however I will close it now that it has been investigated.

PR #912, opened 2026-05-20
Fix setting STR_END in JIT

zherczeg:
When OP_VREVERSE is present in a non-atomic LOOKBEHIND ASSERTION, the STR_END must be set before the backtracking path is executed.


albertoblue87-netizen (commented):
Hecho

PR #921, opened 2026-07-15
alpha: Add JIT support

mattst88:
Add JIT compiler support for DEC Alpha (AXP) 64-bit, in four commits:

Base
  - Document and package the backend: add sljit Alpha sources to EXTRA_DIST, update the JIT-supported-platforms list. The deps/sljit submodule bump is deferred until the sljit backend is merged upstream.

CMPBGE scanning:
  - Implement `fast_forward_char_simd`, `fast_requested_char_simd`, and `fast_forward_char_pair_simd` using the Alpha `CMPBGE` instruction for 8-byte-at-a-time character scanning, analogous to the SIMD paths on other architectures. The first-match step uses a base-ISA binary search (NEGQ+AND to isolate the lowest set bit, then three AND+SELECT pairs), identical to glibc's Alpha strchr.

memchr:
  - For exact single-character 8-bit scanning, call glibc's `memchr()` instead of the inline CMPBGE loop. glibc's Alpha memchr uses CMPBGE with 3-cache-line prefetching, alignment, and software pipelining that the JIT cannot easily replicate.

Prefetching:
  - Emit `PREFETCH_L1` (LDL R31) 192 bytes ahead in all three CMPBGE scanning loops. Guarded against short haystacks. Unconditional prefetch caused a ~15% regression on inputs shorter than 192 bytes.

Validated on a real HP AlphaServer ES47 (1 GHz EV7): `pcre2_jit_test passes` all tests in 8/16/32-bit modes; RunTest passes with and without JIT.

The last two commits could be dropped from this PR and considered separately, if desired.

Requires https://github.com/zherczeg/sljit/pull/364

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -642,6 +642,11 @@ jobs:
           # cross compilers do not depend on the target's libc headers, and it
           # cannot always be derived from the triple.
           #
+          # Alpha's libc package is libc6.1-dev-alpha-cross, because Alpha uses
+          # libc6.1 rather than libc6, but it provides the libc6-dev-alpha-cross
+          # name built from the "libc" field. Only gcc-14 is packaged for it.
+          - { name: alpha,       triple: alpha-linux-gnu,             qemu: alpha,       libc: alpha,      container: "ubuntu:26.04", gcc_suffix: "-14" }
+          #
           # Ubuntu armhf defaults to Thumb-2, so ARM mode needs asking for.
           - { name: arm,         triple: arm-linux-gnueabihf,         qemu: arm,         libc: armhf,      container: "ubuntu:26.04", cflags: -marm }
           - { name: arm-thumb2,  triple: arm-linux-gnueabihf,         qemu: arm,         libc: armhf,      container: "ubuntu:26.04", cflags: -mthumb }
@@ -675,7 +680,7 @@ jobs:
         run: |
           apt-get -qq update
           apt-get -qq install -y git autoconf automake libtool make \
-                                 gcc-${{ matrix.config.triple }} \
+                                 gcc${{ matrix.config.gcc_suffix }}-${{ matrix.config.triple }} \
                                  libc6-dev-${{ matrix.config.libc }}-cross \
                                  qemu-user
 
@@ -703,7 +708,7 @@ jobs:
           ./configure \
             --build=`./config.guess` \
             --host=${{ matrix.config.triple }} \
-            CC=${{ matrix.config.triple }}-gcc \
+            CC=${{ matrix.config.triple }}-gcc${{ matrix.config.gcc_suffix }} \
             CFLAGS="${{ matrix.config.cflags }} $CFLAGS_GCC_STYLE" \
             --disable-shared \
             --enable-jit \
--- a/Makefile.am
+++ b/Makefile.am
@@ -530,6 +530,7 @@ EXTRA_DIST += \
   deps/sljit/sljit_src/sljitConfigInternal.h \
   deps/sljit/sljit_src/sljitLir.c \
   deps/sljit/sljit_src/sljitLir.h \
+  deps/sljit/sljit_src/sljitNativeALPHA_64.c \
   deps/sljit/sljit_src/sljitNativeARM_32.c \
   deps/sljit/sljit_src/sljitNativeARM_64.c \
   deps/sljit/sljit_src/sljitNativeARM_T2_32.c \
--- a/doc/html/pcre2jit.html
+++ b/doc/html/pcre2jit.html
@@ -54,6 +54,7 @@ platforms:
 <pre>
   ARM 32-bit (v7, and Thumb2)
   ARM 64-bit
+  DEC Alpha 64-bit
   IBM s390x 64 bit
   Intel x86 32-bit and 64-bit
   LoongArch 64 bit
--- a/doc/pcre2jit.3
+++ b/doc/pcre2jit.3
@@ -29,6 +29,7 @@ platforms:
 .sp
   ARM 32-bit (v7, and Thumb2)
   ARM 64-bit
+  DEC Alpha 64-bit
   IBM s390x 64 bit
   Intel x86 32-bit and 64-bit
   LoongArch 64 bit
--- a/maint/manifest-tarball
+++ b/maint/manifest-tarball
@@ -58,6 +58,7 @@ drwxr-xr-x tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/allocator_src
 -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitConfigInternal.h
 -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitLir.c
 -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitLir.h
+-rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitNativeALPHA_64.c
 -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitNativeARM_32.c
 -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitNativeARM_64.c
 -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitNativeARM_T2_32.c
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -607,6 +607,15 @@ typedef struct compare_context {
 #define ARGUMENTS     SLJIT_S4
 #define RETURN_ADDR   SLJIT_R4
 
+/* Number of scratch registers the generated code may use. The Alpha CMPBGE
+   scanning loops keep up to five replicated character constants live across a
+   loop, so they use scratch registers above RETURN_ADDR and must declare them. */
+#if (defined SLJIT_CONFIG_ALPHA && SLJIT_CONFIG_ALPHA)
+#define SCRATCH_REGISTERS 10
+#else
+#define SCRATCH_REGISTERS 5
+#endif
+
 #if (defined SLJIT_CONFIG_X86_32 && SLJIT_CONFIG_X86_32)
 #define HAS_VIRTUAL_REGISTERS 1
 #else
@@ -13789,7 +13798,7 @@ common->compiler = compiler;
 
 /* Main pcre2_jit_exec entry. */
 SLJIT_ASSERT((private_data_size & (sizeof(sljit_sw) - 1)) == 0);
-sljit_emit_enter(compiler, 0, SLJIT_ARGS1(W, W), 5 | SLJIT_ENTER_VECTOR(SLJIT_NUMBER_OF_SCRATCH_VECTOR_REGISTERS), 5, private_data_size);
+sljit_emit_enter(compiler, 0, SLJIT_ARGS1(W, W), SCRATCH_REGISTERS | SLJIT_ENTER_VECTOR(SLJIT_NUMBER_OF_SCRATCH_VECTOR_REGISTERS), 5, private_data_size);
 
 /* Register init. */
 reset_ovector(common, (re->top_bracket + 1) * 2);
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -609,7 +609,9 @@ typedef struct compare_context {
 
 /* Number of scratch registers the generated code may use. The Alpha CMPBGE
    scanning loops keep up to five replicated character constants live across a
-   loop, so they use scratch registers above RETURN_ADDR and must declare them. */
+   loop, so they use scratch registers above RETURN_ADDR and must declare them.
+   Registers above SLJIT_NUMBER_OF_TEMPORARY_REGISTERS are callee-saved, so
+   every JIT function pays to spill them: keep this as small as it can be. */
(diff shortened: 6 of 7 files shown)


zherczeg:
Is this architecture is still supported by HP? Alpha EV7 is released in 2003.
I want to drop mips from the codebase, because the last MIPS cpu was released in 2015.

mattst88:
> Is this architecture is still supported by HP? Alpha EV7 is released in 2003.

It is not supported by HP, no.

> I want to drop mips from the codebase, because the last MIPS cpu was released in 2015.

Is it causing maintenance issues? I have a passing interest in MIPS and could have a look.

Dropping it based on release date doesn't seem like the right metric.

zherczeg:
The reasoning is that the jit compiler is normally used in high performance environments. If you have old machines, they are likely not used in production environments. They simply used for "showing that they still works". Supporting them is not worth the maintenance efforts. MIPS is not the first arch which was dropped, sparc and TileGX was also supported at some point of time.

Anyway, if somebody is willing to maintaining these architectures, I have no problem supporting these outdated cpus. The problem is that if the maintainer disappears after a few years. I have no resources to maintain them.

mattst88:
Thanks. That makes complete sense to me and I fully agree.

I've been working to support old/weird architectures for 15+ years now, so I'm happy to contribute to sljit and pcre2 in that capacity. (Though I cannot promise I won't suddenly be hit by a bus)

I'll start by checking on the status of the test suite on mips tomorrow. I found an issue indicating it might be more than a little broken :)

NWilson:
I don't have any particular influence over sljit. I have been struggling with personal to keep up with PCRE2 recently - so anything that keeps maintenance down would be nice.

Dead architectures like MIPS or Alpha or Sparc seem totally reasonable to drop. Retrocomputing can focus on keep old software alive on old systems, but doesn't need to bring new software to such systems. Or at least, I don't the time or energy to contribute to that effort personally.

I feel like there are several bits of code in PCRE2 (including some z/OS support and VMS) which is clearly broken, and I'm quite confident doesn't compile anymore. It's always painful to "pull the trigger" and delete it.

I'm afraid that old sljit architectures would end in the same place: code contributed by one expert, used by one or two users, and then after a couple of years be dead, and with little way for us to know if it still works.

mattst88:
The sljit code has landed upstream. Updated this branch to pull that in first.

NWilson:
@zherczeg I can ask Claude and GPT to review this, but otherwise... it's just a total black box to me. I know nothing about Alpha.

I presume, since you accepted the PR in sljit, that you're happy to accept the Alpha support in PCRE2?

NWilson:
We need to add the new file to the tarball manifest.

I cannot push this change, because maintainer updates to your branch are not turned on, so you'll have to do it yourself.

```diff
--- a/maint/manifest-tarball
+++ b/maint/manifest-tarball
@@ -57,6 +57,7 @@ drwxr-xr-x tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/allocator_src
 -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitConfigInternal.h
 -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitLir.c
 -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitLir.h
+-rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitNativeALPHA_64.c
 -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitNativeARM_32.c
 -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitNativeARM_64.c
 -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitNativeARM_T2_32.c
```

zherczeg:
While alpha support is acceptable for sljit, I am not sure about PCRE2. Even with the sljit core code, it should work. I would not go adding a lot of specialized code for PCRE2 though. @NWilson what do you think?

NWilson:
The problem is we have no easy way of reviewing or fixing any issues. I'm really not that excited about adding an architecture that's not been on sale for 20 years.

On the hand... I'm not mean, and if a user wants it, I guess I can take it.

I have asked GPT to do a very thorough review of all the new Alpha code, and it couldn't find any issues.

@mattst88 Is there the possibility of cutting down on the amount of Alpha-specific specialisation so the PCRE2 code that's specific to Alpha is shorter?

mattst88:
We could certainly only land the first two commits and get basic Alpha support. I'll run some benchmarks (or find where I recorded benchmark results while writing this series) and post them.

Would it be valuable (in general) to run the PCRE2 tests under QEMU for supported platforms? That's what we do in [libunwind](https://github.com/libunwind/libunwind) and it has helped immensely. I'd be happy to work on that and then we can circle back to this PR once we've built some confidence (in the approach and in me).

NWilson:
We already run our CI tests on a list of processor architectures, and all OSes that we can get our hands on.

S/390X, PPC64, ARMv7, AArch64, RiscV64...

So we've even got big-endian and 32-bit platforms in there.

Unfortunately the GitHub Actions runner we use (run-on-arch-action) doesn't have Alpha. Not surprising really.

Honestly - it's over to you to keep it running. If you want to contribute a CI job for that, great. But otherwise, it's just your testing keeping it from bitrot.

Personally - I would add the absolute minimum number of lines of code to PCRE2. Why optimise for a platform with so few users?

mattst88:
Okay, I've opened https://github.com/PCRE2Project/pcre2/pull/939 which adds QEMU testing for currently-untested platforms.

> Personally - I would add the absolute minimum number of lines of code to PCRE2. Why optimise for a platform with so few users?

Understood. For me, it's a fun challenge. You're certainly under no obligation to indulge me in that. But, I do enjoy contributing upstream, so if you all are willing, I'd prefer the code to be upstream. Happy to contribute to the project outside of that goal as well, of course!

NWilson:
Very well then. I accept it! With your CI additions as well, we'll even be able to actually run the code, which is great.

And we're always grateful for contributions. I like your enthusiasm!

NWilson:
This does need a small fix though:

> We need to add the new file to the tarball manifest.
> 
> I cannot push this change, because maintainer updates to your branch are not turned on, so you'll have to do it yourself.
> 
> ```diff
> --- a/maint/manifest-tarball
> +++ b/maint/manifest-tarball
> @@ -57,6 +57,7 @@ drwxr-xr-x tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/allocator_src
>  -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitConfigInternal.h
>  -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitLir.c
>  -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitLir.h
> +-rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitNativeALPHA_64.c
>  -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitNativeARM_32.c
>  -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitNativeARM_64.c
>  -rw-r--r-- tarball-dir/pcre2-SNAPSHOT/deps/sljit/sljit_src/sljitNativeARM_T2_32.c
> ```

mattst88:
Okay, here's the story with the follow-on commits. They're pretty important. I made a benchmark that checks a bunch of different cases (which I'll make a PR for, in case it appears to be generally useful).

- `ff-char-absent`, `ff-lit-absent`, `ff-range-absent`: the first code unit never
  occurs, so the scan runs the whole buffer with no false starts. This is pure
  fast-forward cost.
- `ff-class-dense`: dense first character, every candidate fails immediately.
  Worst case for a vectorized scan.
- `ff-lit-found`, `ff-caseless-*`, `ff-pair-common`, `re-email`, `re-alt`,
  `re-longword`: mixed and realistic.
- `ctl-anchored`, `ctl-startline`: no fast-forward, used as a noise floor.

This is the table of operations in the benchmark:

```c
static const struct benchmark benchmarks[] = {
  /* Pure fast-forward cost: the first code unit never occurs, so the JIT
   * scans the whole buffer and reports NOMATCH without a single false start. */
  { "ff-char-absent",      "Q",                    "single absent char" },
  { "ff-lit-absent",       "zqjx",                 "absent literal, char pair start" },
  { "ff-lit-found",        "wombat",               "rare literal, present" },

  /* Character-range fast-forward. */
  { "ff-range-absent",     "[XYZ]",                "absent range" },
  { "ff-range-digits",     "[0-9]{6}",             "sparse digit run" },
  { "ff-class-dense",      "[aeiou]qzx",           "dense first char, all false starts" },

  /* Caseless and pair-based fast-forward. */
  { "ff-caseless-absent",  "(?i)zqjx",             "caseless absent literal" },
  { "ff-caseless-found",   "(?i)Hello",            "caseless literal, present" },
  { "ff-pair-common",      "th",                   "very common char pair" },

  /* Realistic mixed patterns. */
  { "re-email",            "[a-z0-9._%+-]+@[a-z0-9.-]+\\.[a-z]{2,}", "email address" },
  { "re-longword",         "\\b\\w{12,}\\b",       "long word boundary scan" },
  { "re-alt",              "(?:wombat|numbat|quokka)", "alternation of literals" },

  /* Controls that do not use fast-forward at all; these should be flat across
   * every commit in the series and act as a noise floor. */
  { "ctl-anchored",        "\\A\\w+",              "anchored, no fast-forward" },
  { "ctl-startline",       "(?m)^the\\b",          "multiline start-of-line" }
};
```

Here are the numbers, per commit. These are throughput in MB/s:

| bench | base (no JIT) | sljit | +cmpbge | +prefetch | +memchr | overall |
|---|---|---|---|---|---|---|
| ff-char-absent | 1255.6 | 139.8 | 436.2 | 866.7 | 1225.2 | 0.98x |
| ff-lit-absent | 1254.2 | 184.5 | 312.9 | 683.5 | 680.0 | 0.54x |
| ff-lit-found | 55.5 | 199.1 | 278.2 | 516.5 | 527.2 | 9.50x |
| ff-range-absent | 100.7 | 134.4 | 134.8 | 134.2 | 133.0 | 1.32x |
| ff-range-digits | 99.8 | 134.0 | 134.5 | 133.8 | 132.5 | 1.33x |
| ff-class-dense | 7.1 | 32.3 | 34.2 | 34.3 | 33.5 | 4.71x |
| ff-caseless-absent | 629.4 | 198.7 | 318.2 | 617.1 | 614.6 | 0.98x |
| ff-caseless-found | 0.5 | 168.3 | 191.9 | 267.9 | 262.6 | 570.89x (suspect) |
| ff-pair-common | 19.6 | 66.2 | 114.1 | 133.3 | 119.6 | 6.11x |
| re-email | 2.7 | 39.2 | 41.4 | 41.1 | 40.9 | 15.26x |
| re-longword | 3.0 | 10.9 | 10.9 | 10.9 | 10.9 | 3.62x |
| re-alt | 13.3 | 173.7 | 313.5 | 561.6 | 560.0 | 42.04x |
| ctl-startline | 80.2 | 95.4 | 205.0 | 289.0 | 294.1 | 3.67x |
| ctl-anchored | 1339751.5 | 2856440.1 | 2910311.8 | 2651481.6 | 2542647.1 | 1.90x |


|                             commit                                   | pcre2 lines |
|----------------------------------------------------------------------|-------------|
| pre-series base — old sljit pin, no Alpha backend → interpreter only | —           |
| sljit bump: Alpha JIT works, no pcre2 SIMD                           | 0           |
| docs/packaging only (codegen identical to above)                     | 3           |
| CMPBGE fast-forward scanning                                         | +513        |
| guarded prefetch in CMPBGE loops                                     | +24         |
| memchr for exact single-char scan                                    | +76         |

`9a3be4af` (CMPBGE, 513 lines) gives 37% to 112% on single-character, pair, caseless and alternation fast-forward, and on multiline start-of-line. It does nothing for character ranges (-1.4%) and little for the dense false-start case (+5%).

`b1842a38` (prefetch, 24 lines) gives a further 17% to 119% on every pattern CMPBGE helped. It is by a wide margin the best return per line in the series.

`3658b9ec` (memchr, 76 lines) gives 41.8% on exactly one shape, the exact single character scan, and nothing measurable elsewhere. It costs 2.1% on `ff-class-dense`. Its value is that it closes the last deficit against the interpreter on that shape, from 0.70x to 0.98x.

mattst88:
> This does need a small fix though:

Thank you! I've fixed that and repushed. Note that the PR currently now contains the patch from https://github.com/PCRE2Project/pcre2/pull/939 so that I could add another patch that adds CI support for alpha on top. I expect the other PR to land before this and rebase out of this PR.

NWilson:
Wow. This is the dedication of a true hobbyist (I presume). You don't work with Alpha commercially, do you? Anyway, those performance stats are awesome.

We'll take it!

mattst88:
Thank you!

Yep, true hobbyist. Just got interested in weird old computers a long time ago and never grew out of it. (I've recently begun restoring the content of the AlphaLinux.org wiki which I originally created in 2008 and was lost when a hard drive failed in ~2011 -- see https://alphalinux.org/wiki/). If this lands, it'll be a news item for the [dozens](https://knowyourmeme.com/photos/1645190-arrested-development) of Alpha users.

Dozens!

NWilson (approved):

NWilson:
@mattst88 I'm ready for this now! I just needs a merge to fix conflicts (no rebase necessary, I'll squash-merge it regardless).

NWilson:
@mattst88 You didn't enable 'maintainers can push'.

I merged with main - I'd happily push if you let me. Otherwise you'll have to merge in main yourself

mattst88:
Thank you. Rebased!

PR #922, opened 2026-07-19
Fix: Fetch SLJIT through the zig package manager for JIT builds

nankishr:
Zig fetched git packages do not include submodule contents, which caused the following error while building pcre2 with JIT mode on:

image

So I made this tiny patch to fetch the pinned SLJIT revision through zig's package manager when JIT support is enabled and stage the fetched sources alongside `pcre2_jit_compile.c`. This works without modifying the C source or affecting non-JIT builds of pcre2.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -634,14 +634,14 @@ jobs:
       (github.event_name == 'workflow_dispatch' && (inputs.job_id == 'all' || inputs.job_id == 'zebrilus')) ||
       github.event_name == 'push'
     steps:
-      # The zig version from minimum_zig_version in build.zig.zon will be installed.
-      - name: Install zig
-        uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
-
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
-          submodules: true
+          submodules: false
+
+      # The zig version from minimum_zig_version in build.zig.zon will be installed.
+      - name: Install zig
+        uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
 
       - name: Build
         run: zig build -Dsupport_jit
--- a/build.zig
+++ b/build.zig
@@ -16,6 +16,24 @@ pub fn build(b: *std.Build) !void {
     const codeUnitWidth = b.option(CodeUnitWidth, "code-unit-width", "Sets the code unit width") orelse .@"8";
     const jit = b.option(bool, "support_jit", "Enable/disable JIT compiler support") orelse false;
 
+    const jit_source = if (jit) blk: {
+        const sljit = b.lazyDependency("sljit", .{}) orelse
+            return error.MissingDependency;
+
+        const files = b.addWriteFiles();
+
+        _ = files.addCopyDirectory(
+            sljit.path("sljit_src"),
+            "deps/sljit/sljit_src",
+            .{},
+        );
+
+        break :blk files.addCopyFile(
+            b.path("src/pcre2_jit_compile.c"),
+            "src/pcre2_jit_compile.c",
+        );
+    } else b.path("src/pcre2_jit_compile.c");
+
     const pcre2_h_dir = b.addWriteFiles();
     const pcre2_h = pcre2_h_dir.addCopyFile(b.path("src/pcre2.h.generic"), "pcre2.h");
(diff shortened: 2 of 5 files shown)


build.zig.zon
```diff
@@ -7,7 +7,12 @@
     .fingerprint = 0x914b3d63ebab9e21,
 
     .minimum_zig_version = "0.15.2",
-    .dependencies = .{},
+    .dependencies = .{
+        .sljit = .{
+            .url = "git+https://github.com/zherczeg/sljit.git#45f910b78c6605ebf5b53d3ec7cb00f2312fe417",
+            .hash = "N-V-__8AAIUFRABdZku3bW_1S5GOC4iaFh17LsRurEDR1lvv",
```
carenas, line 13: wouldn't this require updating whenever a new version of the sljit submodule is committed?

a `git clone` is meant to be followed by `git submodule update --init` or equivalent; could zig or its package manager do that?

nankishre: Hi, yes the sljit commit hash would have to be updated whenever gitmodules changes aswell and I'm not aware of a helper cmd which has that feature.

Rn, the maintainers would have to run
`zig fetch --save=sljit "git+https://github.com/zherczeg/sljit.git#$(git ls-tree HEAD deps/sljit | awk '{print $3}')"`
every time sljit's commit # is changed. If they aren't interested in doing that maybe I could setup a workflow to check for and update the zig package that if that's okay.

carenas: > I could setup a workflow to check for and update the zig package that if that's okay.

There is a script in `maint/UpdateAlways` where that logic could be added. It is used in CI to do automatic modifications in the source tree, and could therefore update this file if a newer version of the dependency was committed previously.

nankishre: > There is a script in `maint/UpdateAlways` where that logic could be added. It is used in CI to do automatic modifications in the source tree, and could therefore update this file if a newer version of the dependency was committed previously.

Hi thank you for directing me to that. I've made the necessary changes and hope it's good to go!

nankishre:
Squashed some commits ^^

NWilson:
Hi @nankishre! I have tested this out, and made a few improvements. I needed help from AI because I don't really know Zig.

I believe we can update to Zig 0.16.0 now (and probably should do).

Over to you now, to review my changes!

.github/workflows/dev.yml
```diff
@@ -634,14 +634,14 @@ jobs:
...
-
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
-          submodules: true
+          submodules: false
+
+      # The zig version from minimum_zig_version in build.zig.zon will be installed.
+      - name: Install zig
+        uses: mlugg/setup-zig@d1434d08867e3ee9daa34448df10607b98908d29 # v2.2.1
```
NWilson, line 644: It couldn't be installing the version from build.zig.zon, if checkout hasn't yet run in order to fetch that file!

build.zig
```diff
@@ -14,7 +14,41 @@ pub fn build(b: *std.Build) !void {
...
+
+        const files = b.addWriteFiles();
+
+        // Recreate the layout used by pcre2_jit_compile.c's relative include.
+        _ = files.addCopyDirectory(
+            sljit.path("sljit_src"),
+            "deps/sljit/sljit_src",
+            .{},
+        );
+        _ = files.add("src/.empty", "");
```
NWilson, line 48: I didn't want to do a dummy copy of src/pcre2_jit_compile.c, so I have instead added the dependency as an include path.

build.zig.zon
```diff
@@ -6,8 +6,14 @@
...
     .fingerprint = 0x914b3d63ebab9e21,
 
-    .minimum_zig_version = "0.15.2",
-    .dependencies = .{},
+    .minimum_zig_version = "0.16.0",
+    .dependencies = .{
+        .sljit = .{
+            .url = "git+https://github.com/zherczeg/sljit.git#45f910b78c6605ebf5b53d3ec7cb00f2312fe417",
+            .hash = "N-V-__8AAIUFRABdZku3bW_1S5GOC4iaFh17LsRurEDR1lvv",
+            .lazy = true,
```
NWilson, line 14: I believe the lazy flag is needed here - otherwise it's fetched always, regardless of what build.zig says.

.github/workflows/dev.yml
nankishre: Oops apologies that was an oversight from my side.

nankishre:
LGTM!

NWilson:
Thank you NaN!

PR #926, opened 2026-07-31
Ignore JIT if a function is not compiled

zherczeg:


zherczeg:
I could not merge the two invalid utf code paths because of small differences. Instead, the `use_jit` is set to FALSE if the requested JIT function is not compiled.

zherczeg:
Something is wrong with the code coverage public key.

NWilson:
> Something is wrong with the code coverage public key.

This a problem with the codecov plugin. They fixed it, but we had to update to the latest version of the CI widget.

src/pcre2_match.c
NWilson: This comment about "fall through to the interpreter" needs to be updated

src/pcre2_match.c
```diff
@@ -7088,8 +7085,9 @@ JIT executable instead of the rest of this function. Most options must be set
 at compile time for the JIT code to be usable. */
 
 #ifdef SUPPORT_JIT
-use_jit = (re->executable_jit != NULL &&
-          (options & ~PUBLIC_JIT_MATCH_OPTIONS) == 0);
+if (re->executable_jit != NULL &&
+    (options & ~PUBLIC_JIT_MATCH_OPTIONS) == 0)
+  use_jit = PRIV(jit_check_exec)(re->executable_jit, options);
```
NWilson: You could keep the old structure, if you want, and just add an `&& PRIV(jit_check_exec)(...)` at the end. I don't mind.

src/pcre2_internal.h
```diff
@@ -2330,8 +2331,9 @@ extern BOOL         _pcre2_is_newline(PCRE2_SPTR, uint32_t, PCRE2_SPTR,
                       uint32_t *, BOOL);
 extern void         _pcre2_jit_free_rodata(void *, void *);
 extern void         _pcre2_jit_free(void *, pcre2_memctl *);
-extern size_t       _pcre2_jit_get_size(void *);
 const char *        _pcre2_jit_get_target(void);
+extern size_t       _pcre2_jit_get_size(void *);
+extern BOOL         _pcre2_jit_check_exec(void *, uint32_t);
```
NWilson, line 2334: Not alphabetical - very minor (and above)

src/pcre2_jit_misc_inc.h
```diff
@@ -231,4 +231,29 @@ return executable_sizes[0] + executable_sizes[1] + executable_sizes[2];
...
+#else  /* SUPPORT_JIT */
+executable_functions *functions = (executable_functions *)executable_jit;
+int index = 0;
+
+if ((options & PCRE2_PARTIAL_HARD) != 0)
+  index = 2;
+else if ((options & PCRE2_PARTIAL_SOFT) != 0)
+  index = 1;
+
+return functions->executable_funcs[index] != NULL;
```
NWilson: I would add a one-line comment in pcre2_jit_match, where this logic was taken from, which says `/* Note the selection of executable_funcs should stay in sync with _pcre_jit_check_exec */`

NWilson (commented):
Great, it looks good overall. I have some small suggestions but I don't think there are any actual bugs in here. Thanks!

src/pcre2_match.c
zherczeg: I just deleted the `use_jit` variable.

zherczeg:
Thank you for the review! Patch is updated

src/pcre2_jit_misc_inc.h
```diff
@@ -200,17 +200,28 @@ if (jit_stack != NULL)
 
 
 /*************************************************
-*               Get target CPU type              *
+*          Checks function compilation           *
 *************************************************/
 
-const char*
-PRIV(jit_get_target)(void)
```
zherczeg, line 207: This function is moved below to follow the alphabetic order

NWilson:
Excellent! Thanks.

PR #928, opened 2026-08-03
pcre2_compile: tighten calculation of re_blocksize

### LINKED ISSUE #918: Heap-buffer-overflow WRITE in pcre2_compile_32 (OSS-Fuzz #488713937)
## Summary

OSS-Fuzz has identified a heap-buffer-overflow (WRITE 4) in `pcre2_compile_32`, reported as [OSV-2026-343](https://osv.dev/vulnerability/OSV-2026-343). The issue has been public since 2026-03-03 but does not appear to have a corresponding GitHub issue or fix.

## Details

- **Crash type:** Heap-buffer-overflow WRITE 4
- **Crash state:** `pcre2_compile_32` / `pcre2_fuzzsupport.c`
- **OSS-Fuzz report:** https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=488713937
- **OSV entry:** https://osv.dev/vulnerability/OSV-2026-343
- **Severity:** HIGH
- **Introduced in:** commit 24f9d8df0b8ddabc217ec4e7856a678e09f52773
- **Affected versions:** 10.45, 10.45-RC1, 10.46, 10.47

## Questions

1. Is this issue already tracked internally via the OSS-Fuzz dashboard?
2. Is a fix in progress, or is this considered low priority given that the crash state references `pcre2_fuzzsupport.c`?

## Context

This was surfaced by an OSV-based vulnerability scan of Homebrew packages. PCRE2 is a transitive dependency of many core tools (git, zsh, ripgrep, etc.), so visibility on the fix timeline would be appreciated.

carenas:
Fixes: #918

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -10295,7 +10295,7 @@ PCRE2_SPTR ptr;                       /* Current pointer in pattern */
 uint32_t *pptr;                       /* Current pointer in parsed pattern */
 
 PCRE2_SIZE length = 1;                /* Allow for final END opcode */
-PCRE2_SIZE usedlength;                /* Actual length used */
+INT64_OR_DOUBLE usedlength;           /* Actual length used */
 PCRE2_SIZE re_blocksize;              /* Size of memory block */
 PCRE2_SIZE parsed_size_needed;        /* Needed for parsed pattern */
 
@@ -10853,8 +10853,7 @@ block for storing the compiled pattern and names table. Integer overflow should
 no longer be possible because nowadays we limit the maximum value of
 cb.names_found and cb.name_entry_size. */
 
-re_blocksize =
-  CU2BYTES((PCRE2_SIZE)cb.names_found * (PCRE2_SIZE)cb.name_entry_size);
+usedlength = CU2BYTES((size_t)cb.names_found * cb.name_entry_size);
 
 #if defined SUPPORT_WIDE_CHARS
 if (cb.char_lists_size != 0)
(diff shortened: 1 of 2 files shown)


src/pcre2_compile.c
```diff
@@ -10295,7 +10295,7 @@ PCRE2_SPTR ptr;                       /* Current pointer in pattern */
 uint32_t *pptr;                       /* Current pointer in parsed pattern */
 
 PCRE2_SIZE length = 1;                /* Allow for final END opcode */
-PCRE2_SIZE usedlength;                /* Actual length used */
+INT64_OR_DOUBLE usedlength;           /* Actual length used */
```
zherczeg, line 10298: I would not use double. Just throw an error if the number is too big.

carenas: For all purposes (specially since the codebase is C99 and most CPUs are 64bit) this is equivalent to `int64_t` which is of course what we need to check for integer overflows, specially when size_t is 32bit

src/pcre2_compile.c
```diff
@@ -10853,30 +10853,39 @@ block for storing the compiled pattern and names table. Integer overflow should
 no longer be possible because nowadays we limit the maximum value of
 cb.names_found and cb.name_entry_size. */
 
-re_blocksize =
-  CU2BYTES((PCRE2_SIZE)cb.names_found * (PCRE2_SIZE)cb.name_entry_size);
+usedlength = CU2BYTES((size_t)cb.names_found * cb.name_entry_size);
```
zherczeg, line 10856: What about checking the error with: `if (CU2BYTES(cb.names_found) > ~(PCRE2_SIZE)0 / CU2BYTES(name_entry_size))`

carenas: that would be imposible, since both of those are `uint16_t`, and AFAIK name_entry_size can't be over 128, but agree that the way the checking is done is convoluted and seem to leave gaps.

zherczeg: The arguments are converted to size_t, I don't understand what is impossible.

NWilson:
@carenas, would you mind if I had a go at this issue too? I'm digging deeper, and want to clean up and fix quite a few unchecked additions in the same area.

I don't want to ignore your work here, but I would prefer to make a less targeted (broader) fix.

src/pcre2_compile.c
NWilson: I'm actually planning to get rid of INT64_OR_DOUBLE. It's only really used in one place, and it isn't actually required there, so ideally it would just disappear.

src/pcre2_compile.c
carenas: an integer overflow is impossible because both variables are uint16_t and more importantly they are further restricted, so names_found <= 10000 and name_entry_size <= 128

PR #938, opened 2026-08-08
Use all available CPUs in CI instead of a fixed -j3

mattst88:
Every CI job hard-codes `-j3` for `make` and `ctest`. GitHub's hosted Linux runners currently have four CPUs, and larger or self-hosted runners have more, so this leaves capacity unused.

There is no portable way to ask for the CPU count, so each platform gets the spelling it supports:

| Platform | Expression |
|---|---|
| Linux hosts, containers, MSYS2, run-on-arch | `$(nproc)` |
| macOS | `$(sysctl -n hw.ncpu)` |
| FreeBSD, OpenBSD | `NPROC=`sysctl -n hw.ncpu`` once per VM script |
| Solaris | `NPROC=`getconf NPROCESSORS_ONLN`` |
| Windows (PowerShell) | `-j $env:NUMBER_OF_PROCESSORS` |

On the BSDs and Solaris the count is computed once at the top of the VM script and reused, rather than repeating the command at each call site.

One commit per platform, so the unusual ones can be reviewed or dropped individually.

No functional change: the same jobs run the same builds and tests.

Two things deliberately left alone:

- The z/OS `ctest -j3`. I don't know whether `nproc` exists in the zopen environment, and that line has a pre-existing syntax error (`--output-on-failure; &&`) which I've included a patch to fix
- The Solaris compilation steps, which were already serial and stay that way. Only its `ctest` changes.


NWilson:
Thank you! These changes are a nice maintenance improvement.

PR #939, opened 2026-08-08
Test the remaining JIT backends under QEMU

mattst88:
PCRE2 has hand-written code for every architecture SLJIT supports, but CI only exercises some of it. The existing `ptarmigan` job covers s390x, ppc64le, armv7, aarch64 and riscv64. This adds a job covering the backends that nothing currently touches:

| Target | Backend |
|---|---|
| arm (`-marm`) | `sljitNativeARM_32.c` |
| arm (`-mthumb`) | `sljitNativeARM_T2_32.c` |
| powerpc | `sljitNativePPC_32.c` |
| powerpc64 (big-endian) | `sljitNativePPC_64.c` |
| mips, mipsel | `sljitNativeMIPS_32.c` |
| mips64, mips64el | `sljitNativeMIPS_64.c` |
| mips32r6el, mips64r6el | as above, `SLJIT_MIPS_REV >= 6` paths |
| loongarch64 | `sljitNativeLOONGARCH_64.c` |

Two of these are easy to overlook. Ubuntu's armhf compiler defaults to Thumb-2, so `armv7` in `ptarmigan` never exercises the ARM-mode backend. And ppc64le is ELFv2 while ppc64 big-endian is ELFv1, with function descriptors and a different call sequence, so the two are not interchangeable.

## Why a new job rather than more rows in ptarmigan

`ptarmigan` uses `run-on-arch-action`, which boots a container of the target architecture. That action supports only armv6, armv7, aarch64, riscv64, s390x and ppc64le, so none of the targets above can be added to it. (It also emulates with QEMU, so this is not a change in kind, just in mechanism.)

Cross-compiling and emulating only the test programs is also faster than emulating a whole toolchain.

`RunTest` already accepts a `-sim` prefix for exactly this, so no test scripts change. Shared libraries are disabled so the test programs are real executables rather than libtool wrapper scripts, which `qemu-user` cannot run.

## Cost

Eleven jobs. Measured locally on a fast desktop, each is about 2.5 minutes of compute: the cross-build runs at native speed, `RunTest` costs ~10x native under emulation (15-21s) and `pcre2_jit_test` ~20-29x (82-113s). Allowing for a slower runner plus container pull, apt and autogen, expect 6-10 minutes per job. The timeout is set to 20 minutes.

Gated to `push` and `workflow_dispatch` only, matching `ptarmigan`, so pull requests are unaffected.

## Testing

Cross-built and run under qemu-user locally for arm, loongarch64, powerpc, powerpc64 and mips64. `RunTest` and `pcre2_jit_test` pass on all five, including both big-endian PowerPC targets.

Not verified locally: the 32-bit MIPS targets, the little-endian MIPS variants and the release 6 rows, for want of toolchains, and the Thumb-2 row. Package availability for all rows was checked against the Ubuntu archive indices rather than by building, so the first CI run may need fixups.

## Notes

Two implementation details that I should note:

- The cross compilers do not depend on the target's libc headers, so `libc6-dev-<arch>-cross` is installed explicitly. That needs the Debian architecture name, which is not always derivable from the triple (`mipsisa32r6el-linux-gnu` needs `libc6-dev-mipsr6el-cross`).
- MIPS release 6 has no separate `qemu-user` binary, so `QEMU_CPU` names the emulated CPU instead. The defaults predate release 6.

The MIPS cross compilers are no longer built for Ubuntu 26.04, so those rows use a 24.04 container.

### DIFF
--- a/.github/workflows/dev.yml
+++ b/.github/workflows/dev.yml
@@ -23,6 +23,7 @@ on:
         - chaffinch
         - fruitbat
         - ptarmigan
+        - coelacanth
         - zebrilus
         - bee
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/dev.yml
```diff
@@ -626,6 +627,100 @@ jobs:
...
+          # PowerPC big-endian; ptarmigan covers ppc64le, which is a different
+          # ABI to ppc64.
+          - { name: powerpc,     triple: powerpc-linux-gnu,           qemu: ppc,         libc: powerpc,    container: "ubuntu:26.04" }
+          - { name: powerpc64,   triple: powerpc64-linux-gnu,         qemu: ppc64,       libc: ppc64,      container: "ubuntu:26.04" }
+          # The MIPS cross compilers are no longer built for Ubuntu 26.04,
+          # hence the older container.
+          - { name: mips,        triple: mips-linux-gnu,              qemu: mips,        libc: mips,       container: "ubuntu:24.04" }
+          - { name: mipsel,      triple: mipsel-linux-gnu,            qemu: mipsel,      libc: mipsel,     container: "ubuntu:24.04" }
+          - { name: mips64,      triple: mips64-linux-gnuabi64,       qemu: mips64,      libc: mips64,     container: "ubuntu:24.04" }
+          - { name: mips64el,    triple: mips64el-linux-gnuabi64,     qemu: mips64el,    libc: mips64el,   container: "ubuntu:24.04" }
```
zherczeg, line 661: Does this mean ubuntu dropped mips support?

mattst88: Ubuntu never had MIPS support, as far as I know. This PR is using Ubuntu because it's convenient and provides access to Debian's cross toolchain packages (which is what these are).

What I've gathered is that there is currently no maintainer for the mips cross toolchain packages in Debian, so they haven't been updated in whatever version of Debian feeds into Ubuntu 26.04.

.github/workflows/dev.yml
```diff
@@ -626,6 +627,100 @@ jobs:
...
+        run: |
+          ./configure \
+            --build=`./config.guess` \
+            --host=${{ matrix.config.triple }} \
+            CC=${{ matrix.config.triple }}-gcc-14 \
+            CFLAGS="${{ matrix.config.cflags }} $CFLAGS_GCC_STYLE" \
+            --disable-shared \
+            --enable-jit \
+            --enable-pcre2-16 \
+            --enable-pcre2-32
```
NWilson: Could we use:

`--enable-Werror` to pick up warnings;
`--enable-debug` to turn on runtime assertions?

I generally want all CI builds to be built with assertions enabled

.github/workflows/dev.yml
```diff
@@ -626,6 +627,100 @@ jobs:
...
+
+      - name: Build
+        run: make -j$(nproc)
+
+      - name: Test (main test script)
+        run: ./RunTest -sim qemu-${{ matrix.config.qemu }}
+
+      - name: Test (JIT test program)
+        run: qemu-${{ matrix.config.qemu }} ./pcre2_jit_test
```
NWilson, line 723: We could add a comment:

`# RunGrepTest is skipped as it does not have a "-sim" flag yet`

I'll probably never add it. But a comment doesn't hurt.

.github/workflows/dev.yml
```diff
@@ -626,6 +627,100 @@ jobs:
...
+      github.event_name == 'push'
+    env:
+      # Where qemu-user looks for the target's shared libraries.
+      QEMU_LD_PREFIX: /usr/${{ matrix.config.triple }}
+    steps:
+      - name: Setup
+        run: |
+          apt-get -qq update
+          apt-get -qq install -y git autoconf automake libtool make \
+                                 gcc-14-${{ matrix.config.triple }} \
```
NWilson: Is the a virtual package for `gcc-$triple`, instead of hardcoding GCC 14?

Also below, `CC=$triple-gcc-14` is hardcoded, and I wonder if there's a symlink without the version number?

.github/workflows/dev.yml
```diff
@@ -626,6 +627,100 @@ jobs:
...
+          # Ubuntu armhf defaults to Thumb-2, so ARM mode needs asking for.
+          - { name: arm,         triple: arm-linux-gnueabihf,         qemu: arm,         libc: armhf,      container: "ubuntu:26.04", cflags: -marm }
+          - { name: arm-thumb2,  triple: arm-linux-gnueabihf,         qemu: arm,         libc: armhf,      container: "ubuntu:26.04", cflags: -mthumb }
+          - { name: loongarch64, triple: loongarch64-linux-gnu,       qemu: loongarch64, libc: loong64,    container: "ubuntu:26.04" }
+          # PowerPC big-endian; ptarmigan covers ppc64le, which is a different
+          # ABI to ppc64.
+          - { name: powerpc,     triple: powerpc-linux-gnu,           qemu: ppc,         libc: powerpc,    container: "ubuntu:26.04" }
+          - { name: powerpc64,   triple: powerpc64-linux-gnu,         qemu: ppc64,       libc: ppc64,      container: "ubuntu:26.04" }
+          # The MIPS cross compilers are no longer built for Ubuntu 26.04,
+          # hence the older container.
```
NWilson: If MIPS has been dropped by Ubuntu, could we just leave it out ourselves? The usage with PCRE2 must be really, really low.

Ubuntu 24 has two years of support left. If we're going to remove MIPS at that point, I wonder whether it's worth adding it at all.

NWilson (commented):
Cool! I'll trying running. I've added a few comments and queries; let me know what you think.

.github/workflows/dev.yml
mattst88: The story is a bit more complex than "Ubuntu dropped MIPS" since Ubuntu itself never supported MIPS. See my other comment here: https://github.com/PCRE2Project/pcre2/pull/939#discussion_r3742560043

I'd suggest landing the MIPS CI jobs even if you're planning to remove the code (and it seems you're both antsy to do so). I'd like to at least fix the N32 ABI support, which is currently broken (O32 and N64 ABIs work fine) before its sent off into the sunset.

.github/workflows/dev.yml
mattst88: Yes, definitely. Fixed.

.github/workflows/dev.yml
mattst88: Dropped `-14` from both package name and `CC=`. Thanks!

.github/workflows/dev.yml
mattst88: Done, thanks!

.github/workflows/dev.yml
```diff
@@ -624,23 +625,125 @@ jobs:
...
             ninja
-            ctest -j$(nproc) --output-on-failure && (cat ./Testing/Temporary/LastTest.log || true)
+            ctest -j3 --output-on-failure && (cat ./Testing/Temporary/LastTest.log || true)
 
-  zebrilus:
-    # Tests with: Zig compiler. A "zebrilus" is known as a "zigzag heron".
-    name: Zig
+  coelacanth:
+    # Tests with: cross-compilation and qemu-user, for the JIT backends that no
+    # other job covers. A coelacanth is a fish that was thought to be extinct.
```
NWilson, line 632: Also - I absolutely love your choice of name for the build stage.

NWilson:
I am holding off merging this until the builds all pass.

There seem to be some MIPS bugs in sljit which prevent the CI from passing all the tests.

I know nothing about MIPS, but GPT fixed the tests so I sent in a PR to Zoltan. If you know anything about it, maybe you could review it!

NWilson (approved):

PR #940, opened 2026-08-09
Add a testcase for #926

NWilson:


src/pcre2test_inc.h
```diff
@@ -1995,9 +1995,19 @@ switch(cmd)
...
-  if ((coptlist[i].value != 0) == if_inverted)
+  if (coptlist[i].type == CONF_FIX)
+    {
+    config_value = coptlist[i].value;
+    }
+  else /* if (coptlist[i].type == CONF_INT) */
+    {
+    (void)pcre2_config(coptlist[i].value, &config_value);
+    }
+  if ((config_value != 0) == if_inverted)
```
zherczeg, line 2007: It feels there is an `else` before the `if`. Please add a newline.

zherczeg (approved):
LGTM

testdata/testinput17
```diff
@@ -336,4 +336,14 @@
     zabcz\=callout_error=1
     zabcz\=callout_error=1,substitute_matched
 
+# Check UTF validation when falling back from a partially-compiled JIT pattern
+# to the interpreter
+
+#if unicode
```
NWilson, line 342: This was a cunning (?) thing I added. We have many test files... but not for the case must-support-JTI AND must-support-Unicode.

Instead of making a new file for every combination of conditions, I'm starting to make the tests use conditional constructs.

PR #941, opened 2026-08-10
Scan character classes with SIMD in the JIT on x86-64

mattst88:
`fast_forward_start_bits()` tests the start bitmap one code unit at a time on
every architecture, so a pattern whose first character is a class scans far more
slowly than one starting with a literal, even though the JIT already has a
vectorized scan for the literal case.

This adds a hook for it and implements it for x86-64 with SSE2, both for a class
at the start of the match and for one at a relative offset, so that `.{3}[QXZ]`
is searched as well as `[QXZ]`.

### Commits

**Add a helper to split a class bitmap into ranges.** `optimize_class_ranges()`
already reduced a 256 bit bitmap to the code unit values at which membership
changes. That scan moves into `extract_class_ranges()`, placed before the
include of `pcre2_jit_simd_inc.h` so the vectorized scan can use it too, and
takes a limit on the number of entries rather than the fixed
`MAX_CLASS_RANGE_SIZE`. No functional change.

**Emit the x86 vector compares through one helper.**
`fast_forward_char_pair_sse2_compare()` assembled each instruction byte by byte,
patching the prefix in place for the AVX2 form, so the same four bytes were put
together seven times over in slightly different ways. The encoding moves into
`emit_vector_op()`, which takes an opcode and three register indices and picks
the two operand SSE2 form or the three operand AVX2 one from the register width.
It also emits a REX prefix, which the byte-by-byte code had no need for, so that
later users can reach the upper vector registers. The emitted bytes are
unchanged for every combination of compare type, register width, step and
register assignment.

**Scan character classes with SIMD in the JIT on x86-64.** The bitmap is reduced
to a list of ranges at compile time; the scan then tests each range with the
usual unsigned range idiom, where subtracting the low bound makes the range
start at zero and a saturating subtract of the span leaves zero exactly for the
bytes inside it, since anything below the low bound wraps to a value larger than
the span. That is `PSUBB`, `PSUBUSB` and `PCMPEQB` per range, `OR`'d together,
with `PCMPEQB` alone for a single character. Bitmaps needing more than four
ranges keep the existing scan.

**Scan for a character class at a relative offset in the JIT.** `scan_prefix()`
already walks the class bitmap at each position, but expands it into the `chars`
array and keeps at most `MAX_DIFF_CHARS` characters, so a larger class left
nothing behind but "any character". It now records the bitmap itself alongside
the array. A position keeps its bitmap only while one and the same class reaches
it, since a second alternative, a literal or an unbounded class all make it
describe less than the position accepts.

`fast_forward_first_n_chars()` then picks the sparsest such position, but only
where it would otherwise give up: a position holding one or two characters gets
a cheaper scan of its own, and a run of them gets the shift table, both of which
beat testing a handful of ranges.

### Numbers

Measured over a 4MB subject on an i7-1370P, scanning for a class that does not
occur:

| pattern | before | after |
| --- | --- | --- |
| `[QXZ]` | 1445 MB/s | 25961 MB/s |
| `[0-9]{6}` | 3337 MB/s | 33440 MB/s |
| `.{3}[QXZ]` | 450 MB/s | 25798 MB/s |
| `.{2}[0-9]{6}` | 647 MB/s | 33424 MB/s |

Every other pattern measured is unchanged, including `a.{4}[QXZ]` at 1486 MB/s,
which keeps its single character scan rather than being taken over by the class
scan.

### Notes

Vectorizing only pays for a sparse class. The byte-at-a-time loop stops at the
first code unit in the class, so where the class is dense it stops almost
immediately, while a vector loop has already tested a whole block. Measured on
`\b\w{12,}\b`, whose class accepts 63 code units, the vector scan was 38.6%
slower. Classes accepting more than 48 code units are therefore left alone, and
`\b\w{12,}\b` and `.{3}\w{12}` are unchanged.

The bounds of each range live in a vector register of their own, so the scan
needs twelve of them, more than the Win64 ABI leaves as scratch registers. It is
limited to the other x86-64 ABIs, as `fast_forward_char_pair_simd()` already is,
with a compile time assertion tying the range limit to
`SLJIT_NUMBER_OF_SCRATCH_VECTOR_REGISTERS`.

SSE4.2's `PCMPESTRI` was measured as an alternative. It handles up to eight
ranges in one instruction and so is flat in the number of ranges, but it is
slower than this sequence for one or two ranges, which is what real classes
mostly are: 11.5 GB/s against 28.5 GB/s for a single range. It also has no
256-bit form, so it would foreclose widening this loop to AVX2 later.
`PCMPISTRI`, the faster of the two, cannot be used at all, because it treats a
zero byte as the end of the subject.

RunTest and pcre2_jit_test pass at all three code unit widths, on x86-32, and
with the SIMD scans compiled out.

### DIFF
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -381,6 +381,13 @@ typedef struct fast_forward_char_data {
   sljit_u8 count;
   /* Number of last UTF-8 characters in the chars array. */
   sljit_u8 last_count;
+  /* TRUE when something other than the class in class_bits also reaches this
+     position, which makes class_bits unusable. */
+  sljit_u8 mixed;
+  /* Bitmap of the characters accepted in the current position, when the only
+     thing reaching it is a single class. NULL otherwise. Unlike the chars
+     array this is not limited to MAX_DIFF_CHARS characters. */
+  const sljit_u8 *class_bits;
   /* Available characters in the current position. */
   PCRE2_UCHAR chars[MAX_DIFF_CHARS];
 } fast_forward_char_data;
@@ -5851,6 +5858,23 @@ if (last)
   chars->last_count++;
 }
 
+/* Records that the characters accepted in the current position come from the
+class bitmap in bits, or from something which is not a class when bits is NULL.
+A position keeps its bitmap only while one and the same class reaches it. */
+
+static SLJIT_INLINE void set_prefix_class_bits(fast_forward_char_data *chars, const sljit_u8 *bits)
+{
+if (bits != NULL && !chars->mixed
+    && (chars->class_bits != NULL ? chars->class_bits == bits : chars->count == 0))
+  {
+  chars->class_bits = bits;
+  return;
+  }
+
+chars->class_bits = NULL;
+chars->mixed = TRUE;
+}
+
 /* Value can be increased if needed. Patterns
 such as /(a|){33}b/ can exhaust the stack.
 
@@ -5885,6 +5909,7 @@ BOOL last, any, class, caseless, accept_above;
 int stack_ptr, step_count, repeat, len, len_save;
 sljit_u32 chr; // Any unicode character.
 sljit_u8 *bytes, *bytes_end, byte;
+const sljit_u8 *class_start;
 PCRE2_SPTR alternative, cc_save, oc;
 #if defined SUPPORT_UNICODE && PCRE2_CODE_UNIT_WIDTH == 8
 PCRE2_UCHAR othercase[4];
@@ -6195,6 +6220,7 @@ while (TRUE)
     do
       {
       chars->count = 255;
+      set_prefix_class_bits(chars, NULL);
       chars++;
       }
     while (--repeat > 0 && chars < chars_end);
@@ -6214,6 +6240,9 @@ while (TRUE)
 #else
     accept_above = FALSE;
 #endif
+    /* Where it accepts them, the bitmap does not describe the class on its
+    own. */
+    class_start = accept_above ? NULL : bytes;
     cc += 1 + 32 / sizeof(PCRE2_UCHAR);
 
     SLJIT_ASSERT(last == TRUE && repeat == 1);
@@ -6263,6 +6292,8 @@ while (TRUE)
 
     do
       {
+      set_prefix_class_bits(chars, class_start);
+
       if (accept_above)
         chars->count = 255;
       else if (chars->count != 255)
@@ -6350,6 +6381,7 @@ while (TRUE)
       len--;
 
       chr = *cc;
+      set_prefix_class_bits(chars, NULL);
       add_prefix_char(*cc, chars, len == 0);
 
       if (caseless)
@@ -6481,6 +6513,66 @@ return TRUE;
 
 #endif /* JIT_HAS_FAST_FORWARD_CHAR_PAIR_SIMD */
 
+#ifdef JIT_HAS_FAST_FORWARD_START_BITS_SIMD
+
+/* Counts the code units a class bitmap accepts. */
+
+static int count_class_bits(const sljit_u8 *bits)
+{
+int i, count = 0;
+
+for (i = 0; i < 32; i++)
+  {
+  sljit_u8 byte = bits[i];
+
+  if (byte == 0xff)
+    {
+    count += 8;
+    continue;
+    }
+
+  while (byte != 0)
+    {
+    count += byte & 0x1;
+    byte >>= 1;
+    }
+  }
+
+return count;
+}
+
+static BOOL check_fast_forward_start_bits_simd(compiler_common *common, fast_forward_char_data *chars, int max)
+{
+int i, count, best_i = -1, best_count = 0;
+
+/* The scan tests a whole block of code units before it can stop, so the
+sparser the class the more it gains over testing them one at a time, and the
+sparsest position is the one to pick. Positions holding one or two characters
+are left to the scans which specialize in those. */
+for (i = 0; i < max; i++)
+  {
+  if (chars[i].class_bits == NULL || chars[i].count <= 2)
+    continue;
+
+  count = count_class_bits(chars[i].class_bits);
+
+  if (best_i < 0 || count < best_count)
+    {
+    best_i = i;
+    best_count = count;
+    }
+  }
+
+if (best_i < 0)
+  return FALSE;
+
+/* scan_prefix() keeps a class bitmap only where it describes the class on its
+own, so the class accepts nothing above 255. */
+return fast_forward_start_bits_simd(common, chars[best_i].class_bits, best_i, FALSE);
+}
+
+#endif /* JIT_HAS_FAST_FORWARD_START_BITS_SIMD */
+
 static void fast_forward_first_char2(compiler_common *common, PCRE2_UCHAR char1, PCRE2_UCHAR char2, sljit_s32 offset)
 {
 DEFINE_COMPILER;
@@ -6585,6 +6677,8 @@ for (i = 0; i < MAX_N_CHARS; i++)
   {
   chars[i].count = 0;
   chars[i].last_count = 0;
+  chars[i].mixed = FALSE;
+  chars[i].class_bits = NULL;
   }
 
 max = scan_prefix(common, common->start, chars);
@@ -6699,7 +6793,16 @@ SLJIT_ASSERT(offset == -1 || (chars[offset].count >= 1 && chars[offset].count <=
 if (range_right < 0)
   {
   if (offset < 0)
+    {
+    /* No position holds few enough characters to be worth searching for one at
+    a time, but a position may still hold a class sparse enough to vectorize. */
+#ifdef JIT_HAS_FAST_FORWARD_START_BITS_SIMD
+    if (JIT_HAS_FAST_FORWARD_START_BITS_SIMD && common->mode == PCRE2_JIT_COMPLETE
+        && check_fast_forward_start_bits_simd(common, chars, max))
+      return TRUE;
+#endif
     return FALSE;
+    }
   /* Works regardless the value is 1 or 2. */
   fast_forward_first_char2(common, chars[offset].chars[0], chars[offset].chars[1], offset);
   return TRUE;
@@ -7002,7 +7105,7 @@ jump_list *matches = NULL;
 
 #ifdef JIT_HAS_FAST_FORWARD_START_BITS_SIMD
 if (JIT_HAS_FAST_FORWARD_START_BITS_SIMD && common->mode == PCRE2_JIT_COMPLETE
-    && fast_forward_start_bits_simd(common, start_bits))
+    && fast_forward_start_bits_simd(common, start_bits, 0, TRUE))
   return;
 #endif
 
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -7000,6 +7000,12 @@ struct sljit_jump *found = NULL;
 #endif
 jump_list *matches = NULL;
 
+#ifdef JIT_HAS_FAST_FORWARD_START_BITS_SIMD
+if (JIT_HAS_FAST_FORWARD_START_BITS_SIMD && common->mode == PCRE2_JIT_COMPLETE
+    && fast_forward_start_bits_simd(common, start_bits))
+  return;
+#endif
+
 if (common->match_end_ptr != 0)
   {
   OP1(SLJIT_MOV, TMP1, 0, SLJIT_MEM1(SLJIT_SP), common->match_end_ptr);
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -6389,6 +6389,46 @@ CMPTO(SLJIT_EQUAL, reg, 0, SLJIT_IMM, 0xdc00, label);
 }
 #endif
 
+/* Splits a 256 bit class bitmap into the code unit values at which membership
+of the class changes: entry n is the first code unit whose membership differs
+from that of entry n-1, and the membership below the first entry is given by
+bit zero of the bitmap. An empty or a full bitmap therefore has no entries.
+Returns the number of entries, or -1 when there are more than max_ranges of
+them. */
+
+static int extract_class_ranges(const sljit_u8 *bits, int *ranges, int max_ranges)
+{
+sljit_u8 bit, cbit, all;
+int i, byte, length = 0;
+
+bit = bits[0] & 0x1;
+/* All bits will be zero or one (since bit is zero or one). */
+all = (sljit_u8)-bit;
+
+for (i = 0; i < 256; )
+  {
+  byte = i >> 3;
+  if ((i & 0x7) == 0 && bits[byte] == all)
+    i += 8;
+  else
+    {
+    cbit = (bits[byte] >> (i & 0x7)) & 0x1;
+    if (cbit != bit)
+      {
+      if (length >= max_ranges)
+        return -1;
+      ranges[length] = i;
+      length++;
+      bit = cbit;
+      all = (sljit_u8)-cbit; /* sign extend bit into byte */
+      }
+    i++;
+    }
+  }
+
+return length;
+}
+
 #include "pcre2_jit_simd_inc.h"
 
 #ifdef JIT_HAS_FAST_FORWARD_CHAR_PAIR_SIMD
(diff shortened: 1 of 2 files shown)


mattst88:
This was something that I noticed could be improved when investigating one of the regressions on Alpha.

Not sure if I should add support for necessary ops to sljit first and then use them here or whether the approach this patch takes is okay. Happy to do either.

src/pcre2_jit_simd_inc.h
```diff
@@ -52,6 +52,52 @@ typedef enum {
...
+  sljit_u32 low;
+  sljit_u32 high;
+} start_bits_range;
+
+/* Split a start bitmap into inclusive ranges of set bits, for the vectorized
+   start-bits scans below. Returns the number of ranges, or -1 if there are more
+   than max_ranges of them, in which case the caller must fall back to testing
+   the bitmap one code unit at a time. */
+
+static int extract_start_bits_ranges(const sljit_u8 *bits, start_bits_range *ranges, int max_ranges)
```
zherczeg, line 75: I think there is a similar function in the code somewhere. Not sure it can be used here.

src/pcre2_jit_simd_inc.h
```diff
@@ -373,6 +419,194 @@ if (common->utf && offset > 0)
...
+  emit_sse2_op(compiler, 0x74, dst_ind, low_ind);
+  return;
+  }
+
+/* PSUBB dst, low */
+emit_sse2_op(compiler, 0xf8, dst_ind, low_ind);
+/* PSUBUSB dst, span */
+emit_sse2_op(compiler, 0xd8, dst_ind, span_ind);
+/* PCMPEQB dst, zero */
+emit_sse2_op(compiler, 0x74, dst_ind, zero_ind);
```
zherczeg: This looks like a good idea

src/pcre2_jit_simd_inc.h
```diff
@@ -373,6 +419,194 @@ if (common->utf && offset > 0)
...
+#define X86_START_BITS_MAX_RANGES 4
+
+/* Largest number of accepted code units before the class counts as dense and
+   the byte-at-a-time scan is left to do the job. */
+#define X86_START_BITS_MAX_COVERED 48
+
+/* Emit an SSE2 instruction operating on two vector registers. A REX prefix is
+   needed for xmm8 and above, which the start-bits scan reaches as soon as it
+   holds the bounds of more than two ranges. */
+static SLJIT_INLINE void emit_sse2_op(struct sljit_compiler *compiler, sljit_u8 opcode,
```
zherczeg: Maybe such helpers could be used to improve code readability.

zherczeg (commented):
This looks like a good improvement.

The other similar functions also support relative offset, so not only `[range]` but also `.{n}[range]` is also supported where n is < 15. I.e. the first few characters are too complex to search, but there is a simple range after them.

NWilson:
This looks nice.

However, I recommend that we bump it to the next release, rather than merge it and release it in a hurry.

I'm just trying to get all the required bugfixes in, and stabilise the branch ahead of the upcoming release.

mattst88:
Thanks for the review. The two inline comments are answered in their threads;
this covers the rest.

> The other similar functions also support relative offset, so not only `[range]` but also `.{n}[range]` is also supported where n is < 15.

This is the fourth commit of the series. `scan_prefix()` already walks the class
bitmap at each position, but expands it into the `chars` array and keeps at most
`MAX_DIFF_CHARS` characters, so a larger class left nothing behind but "any
character". It now records the bitmap itself alongside the array, and a position
keeps its bitmap only while one and the same class reaches it, since a second
alternative, a literal or an unbounded class all make it describe less than the
position accepts.

`fast_forward_first_n_chars()` picks the sparsest such position, but only where
it would otherwise give up. A position holding one or two characters gets a
cheaper scan of its own, and a run of them gets the shift table, and both beat
testing a handful of ranges, so this takes no work away from them. `a.{4}[QXZ]`
keeps its single character scan, unchanged at 1486 MB/s.

`.{3}[QXZ]` goes from 450 to 25798 MB/s and `.{2}[0-9]{6}` from 647 to 33424
MB/s over a 4MB subject on an i7-1370P. A dense class at an offset, `.{3}\w{12}`,
is left alone as before.

Two things worth flagging separately.

The scan as I first posted it was broken on Win64. It holds the bounds of each
range in a vector register of its own, so it needs twelve of them, but
`SLJIT_NUMBER_OF_SAVED_FLOAT_REGISTERS` is 10 there and only five vector scratch
registers are requested, so it reached registers the ABI requires to be
preserved. It is now limited to the other x86-64 ABIs the same way
`fast_forward_char_pair_simd()` is, with a compile time assertion tying the range
limit to `SLJIT_NUMBER_OF_SCRATCH_VECTOR_REGISTERS` so the budget cannot
silently overflow again.

I also tried handing the bound registers out densely rather than two per range,
so that a class of single characters stays below xmm8 and needs no REX prefix.
It saves four bytes on `[QXZ]`, but costs 7.7% on `.{3}[QXZ]`, stable over
repeated runs, purely from the loop body shifting relative to its alignment. Not
worth it, so it is not in the series.

The branch is force-pushed and rebased on current main. Tested at all three code
unit widths, on x86-32, and with the SIMD scans compiled out.

src/pcre2_jit_simd_inc.h
mattst88: It was the scan at the top of `optimize_class_ranges()`. The first commit of the
series now moves that into `extract_class_ranges()`, which takes a limit on the
number of entries instead of the fixed `MAX_CLASS_RANGE_SIZE`, and sits before
the include of `pcre2_jit_simd_inc.h` so both users see it.

`extract_start_bits_ranges()` then only has to turn the transition list into
inclusive pairs. `optimize_class_ranges()` is otherwise unchanged: the
membership of the highest code unit, which its loop used to leave behind in
`bit`, is read from the bitmap instead.

src/pcre2_jit_simd_inc.h
mattst88: Done in the second commit of the series. `emit_sse2_op()` became
`emit_vector_op()`, taking an opcode and three register indices and picking the
two operand SSE2 form or the three operand AVX2 one from the register width, and
`fast_forward_char_pair_sse2_compare()` is rewritten on top of it, losing about
60 lines of hand assembled prefix bytes.

The one place the two shapes genuinely differ is the two character compare,
where the 256 bit form does with one instruction what the 128 bit form needs two
for, so that case no longer falls into the shared switch.

Since the AVX2 path is disabled at runtime I could not exercise it the usual way,
so I compiled the old and the new function side by side against a recording stub
and compared their output: 98304 cases, every combination of compare type,
register width, step and register assignment, byte identical.

NWilson:
I am happy, but haven't reviewed in detail.

I did ask GPT 5.6 to dive in and it didn't find any safety or correctness issues.

The branch is ready for merges again.

@zherczeg, please review and merge whenever you are happy with this.

NWilson:
@zherczeg Matt's benchmarks look convincing, and GPT reviewed the code and didn't find bugs. I read it myself of course, but I don't know x86 SIMD well enough to find any subtle mistakes myself.

Let me know if you would like me to hold the PR (since it affects JIT).

If I don't hear, I think I'll merge it.

src/pcre2_jit_compile.c
```diff
@@ -6439,6 +6511,58 @@ return TRUE;
...
+
+/* Counts the code units a class bitmap accepts. */
+
+static int count_class_bits(const sljit_u8 *bits)
+{
+int i, count = 0;
+
+for (i = 0; i < 32; i++)
+  {
+  sljit_u8 byte = bits[i];
```
zherczeg, line 6526: I would add a special case for 0xff values

src/pcre2_jit_simd_inc.h
```diff
@@ -373,6 +431,236 @@ if (common->utf && offset > 0)
...
+
+count = extract_start_bits_ranges(start_bits, ranges, X86_START_BITS_MAX_RANGES);
+if (count <= 0)
+  return FALSE;
+
+/* A vector scan only pays when it gets to scan. The byte-at-a-time loop stops
+   at the first code unit in the class, so for a dense class it usually stops
+   immediately, while this loop has already tested a whole block. Count the
+   code units the class accepts and leave the dense ones alone: \w covers 63 of
+   them and is common enough in real subjects to matter. */
```
zherczeg: So the problem is setting up this scan is too costly for dense classes? Can this be computed by extract_start_bits_ranges and return 0?

src/pcre2_jit_simd_inc.h
```diff
@@ -373,6 +431,236 @@ if (common->utf && offset > 0)
...
+emit_vector_op(compiler, reg_type, 0x6f, dst_ind, -1, src_ind);
+
+if (single)
+  {
+  /* PCMPEQB dst, low */
+  emit_vector_op(compiler, reg_type, 0x74, dst_ind, dst_ind, low_ind);
+  return;
+  }
+
+/* PSUBB dst, low */
```
zherczeg: Is the code limited to 8 bit characters? In theory, 16/32 bit characters could be compared.

src/pcre2_jit_simd_inc.h
```diff
@@ -373,6 +431,236 @@ if (common->utf && offset > 0)
...
+emit_vector_op(compiler, reg_type, 0xf8, dst_ind, dst_ind, low_ind);
+/* PSUBUSB dst, span */
+emit_vector_op(compiler, reg_type, 0xd8, dst_ind, dst_ind, span_ind);
+/* PCMPEQB dst, zero */
+emit_vector_op(compiler, reg_type, 0x74, dst_ind, dst_ind, zero_ind);
```
zherczeg: I would propose two optimizations here.

1) You should use `PCMPGTB` and friends here. The zero point will be SIMM_MAX (127, 32767, ...).
Let's assume the character is stored in 'c', and the `[20-30]` range needs to be checked
`c + (SIMM_MAX - 30) > (SIMM_MAX - (30 - 20) - 1)`

This check requires two instructions instead of three.

2) The last compare does not need to copy `src_ind` to a temporary register, it can use it directly.

I don't know the speed of mask move, but using it after every compare could free a vector register, and use the 'or' operation in normal registers.

zherczeg (changes requested):

src/pcre2_jit_compile.c
mattst88: Done: a byte of 0xff now adds eight at once instead of going round the bit
loop.

src/pcre2_jit_simd_inc.h
mattst88: It is not the setup: the constants are loaded once, outside the loop. The
cost is in how soon each loop can stop. The code unit at a time loop stops at
the first code unit in the class, and for a dense class that is usually one of
the first few it looks at, while the vector loop has always tested a whole
block of 16 bytes by then. On \b\w{12,}\b, whose class accepts 63 code units,
the vector scan was 38.6% slower.

But yes, it fits better there. extract_start_bits_ranges() now takes a limit
on the number of code units as well as on the number of ranges, and returns 0
when the bitmap accepts more than that, so the caller only checks the count.
The reasoning moved to the comment on X86_START_BITS_MAX_COVERED.

src/pcre2_jit_simd_inc.h
mattst88: Not any more. With the PCMPGT form from your other comment the sequence is the
same at every width, using the word and doubleword instructions, so the width
restriction is gone. Above 8 bits, bit 255 of the start bitmap stands for
every code unit from 255 upwards, as in fast_forward_start_bits(), so there
the range which reaches 255 is extended to the largest code unit. A class
bitmap at an offset gets no such extension, since scan_prefix() only keeps
one where it describes the class on its own; the scan takes a flag saying
which kind of bitmap it has.

Scanning a 4M code unit subject for a class that does not occur, in millions
of code units per second on an i7-1370P, [QXZ] goes from 1444 to 13549 at 16
bits and from 1436 to 6339 at 32 bits, and [0-9]{6} from 3593 to 18228 and
from 3152 to 8406. The scan at an offset gets the same: .{3}[QXZ] goes from
455 to 12427 at 16 bits and from 457 to 5852 at 32 bits, and .{2}[0-9]{6}
from 647 to 18297 and from 646 to 8214. Nothing else measured moves.

While testing this I found a separate bug, which predates this series and is
in 10.47: in the 16-bit and 32-bit libraries scan_prefix() only took a
negated class to accept the code units above 255 when bit 255 of its bitmap
was set, so [^\x00-\x40\x42-\xff] fast-forwarded to the next 'A' and missed
\x{100}. The fix is in #980 on its own so that it can be backported, and
this series now builds on it, since the scan at an offset relies on the
same rule for which classes accept the code units above 255.

src/pcre2_jit_simd_inc.h
mattst88: Done, thank you. Each range is now PADD and PCMPGT, where the addend moves the
range to the top of the signed code units, so the zero register is gone, and
the last range is tested in the data register itself. At 8 bits that is
5.3% faster on [QXZ] (25752 to 27121 million
code units per second) and 7.8% faster on [0-9]{6} (33312 to 35898), with
.{2}[0-9]{6} 9.0% faster at an offset. The one pattern to lose is
.{3}[QXZ], 2.7% slower (25718 to 25036), steady over repeated runs. It runs
one instruction fewer per block than before and the same class at offset
zero gains 4.6%, so I take it to be the loop alignment again, as with the
dense register packing I mentioned earlier.

I have not tried moving the mask out after each compare. PMOVMSKB only issues
on one port, and it would turn each POR into a PMOVMSKB and an OR, so I would
expect it to be slower, and with the zero register gone the scan uses eleven
of the sixteen vector registers, so there is room to spare.

zherczeg (approved):
LGTM
Nice improvement!

zherczeg:
I just remembered something. That is windows. It expects you to save some simd registers. But deciding at function entry time that you will use those registers is a nightmare. So some simd code is disabled on windows at the moment. If you use more than 6 registers, this might be also an issue here.

NWilson:
> I just remembered something. That is windows. It expects you to save some simd registers. But deciding at function entry time that you will use those registers is a nightmare. So some simd code is disabled on windows at the moment. If you use more than 6 registers, this might be also an issue here.

That's a good point. However, it has been considered. The new code is protected by a Windows-check in the `#if` - JIT_HAS_FAST_FORWARD_START_BITS_SIMD is not provided on Windows. There are some compiler changes, but I think it's all safe, and the entrypoints to the new code are all conditional.

There's also an explicit assertion at the entrypoint:

```
/* Three registers for the scan itself, and two more for the constants of
   every range it may be asked to handle. */
SLJIT_COMPILE_ASSERT(3 + X86_START_BITS_MAX_RANGES * 2 <= SLJIT_NUMBER_OF_SCRATCH_VECTOR_REGISTERS,
  not_enough_vector_registers);
```

I didn't check it manually, but GPT is confident that's the right number of registers the code actually uses.

NWilson:
~~After merging your negative-class fix, I'll push a merge commit to this branch.~~

Actually... I'll let you fix the conflict, so you can check it yourself.

mattst88:
Rebased!

NWilson:
@mattst88 GitHub doesn't re-run CI when the main branch changes.

I was a sheep and merged this because the CI was green - but it looks like the build fails on x86 when the Alpha and SIMD PRs are both present.

Are you able to fix that?

mattst88:
Yep, I'll look into it right away.

NWilson:
@mattst88 

GPT is pretty confident this is the fix.

EDIT: GPT was clearly wrong!

It looks like a chunk of code was mis-merged.

```diff
diff --git a/src/pcre2_jit_compile.c b/src/pcre2_jit_compile.c
index 62e994a9..4a0a47f5 100644
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -7129,16 +7129,6 @@ if (common->match_end_ptr != 0)
   SELECT(SLJIT_GREATER, STR_END, TMP1, 0, STR_END);
   }

-#ifdef JIT_HAS_FAST_FORWARD_START_BITS_SIMD
-if (JIT_HAS_FAST_FORWARD_START_BITS_SIMD && common->mode == PCRE2_JIT_COMPLETE
-    && fast_forward_start_bits_simd(common, start_bits))
-  {
-  if (common->match_end_ptr != 0)
-    OP1(SLJIT_MOV, STR_END, 0, RETURN_ADDR, 0);
-  return;
-  }
-#endif
-
 start = LABEL();

 partial_quit = CMP(SLJIT_GREATER_EQUAL, STR_PTR, 0, STR_END, 0);
```

PR #942, opened 2026-08-10
bound utf scan and line advance past a \K match in pcre2grep

sage-mode-hunter:
Advancing past a zero-progress \K match in pcre2grep can run startoffset off the end of the subject:
- the UTF-8 continuation-byte scan (the --only-matching and multiline/colour paths) had no upper bound, so under -U (PCRE2_MATCH_INVALID_UTF) a subject ending in continuation bytes reads past the line buffer
- the line-advance loops right after can then spin, since startoffset points past the last line
- bounded both scans against length, the same way the two backward scans in this file already are, and let the advance loops stop once they reach the buffer end

Repro under ASAN: a run of 0x80 bytes ending in one match, a small --buffer-size, -U, and pattern (?<=\K.), reads one byte past main_buffer at the multiline scan.


NWilson:
@sage-mode-hunter I have studied this code a bit, and made some changes to your PR.

You didn't include any testcases. I have used AI to add quite a few, and looked at their results to make sure they match what is expected. The testcases also definitely exercised your ASAN condition, and it now appears fixed.

Feel free to review the changes I made. I will explain them.

I don't know whether you used a tool or performed a manual fix; either way, let me know if you're satisfied that the fixes I added address the issues you found.

src/pcre2grep.c
```diff
@@ -2910,16 +2910,18 @@ while (ptr < endptr)
         oldstartoffset = pcre2_get_startchar(match_data);
         if (startoffset <= oldstartoffset)
           {
-          if (startoffset >= length) goto END_ONE_MATCH;  /* Were at end */
+          if (oldstartoffset >= length) goto END_ONE_MATCH;  /* We're at end */
```
NWilson, line 2913: The change to `oldstartoffset >= length` has actually no effect. It's a tidy-up, but in the case where this holds, it makes no difference.

src/pcre2grep.c
```diff
@@ -2910,16 +2910,18 @@ while (ptr < endptr)
...
           }
 
         /* If the current match ended past the end of the line (only possible
         in multiline mode), we must move on to the line in which it did end
-        before searching for more matches. */
+        before searching for more matches. An offset that is within a line's
+        terminating CRLF sequence still belongs to that line. */
 
-        while (startoffset > linelength)
+        while (endlinelength != 0 && startoffset >= linelength + endlinelength)
```
NWilson, line 2924: The logic here should match the case below. The real issue is the underflow here: when subtracting `linelength + endlinelength` from the (unsigned!) startoffset, it must not wrap.

However, using the same `startoffset > linelength + endlinelength` condition as below isn't possible here, because we have to actually advance to next line in order to count `linenumber` correctly. Hence - we use `>=`. In turn, that could enter an infinite loop on the final trailing line of the input, which the `endlinelength != 0` guards against.

src/pcre2grep.c
```diff
@@ -3080,8 +3082,10 @@ while (ptr < endptr)
 
           if (startoffset <= oldstartoffset)
             {
+            if (oldstartoffset >= length) break;  /* We're at end */
```
NWilson, line 3085: This fix is necessary. It's the better fix.

Instead of guarding against startoffset advancing too far, your commit instead allowed it to go too far, but then guarded the while-loop below. That is less safe than doing the bounds check before advancing startoffset.

sage-mode-hunter:
Reviewed your changes and re-ran my original repro against the branch with an ASAN build (a run of 0x80 bytes, --buffer-size=8, -aUM, --allow-lookaround-bsk '(?<=\K.)'): clean, and RunGrepTest passes here too, including the new U11-U14 cases.

Agree with all three changes. Checking oldstartoffset against length before advancing is the better shape; my version let startoffset run past and relied on guards further down, which is a weaker invariant. The reworked line-advance condition makes sense as well, the unsigned underflow was the real bug there and I hadn't clocked that >= is forced by the linenumber accounting. The oldstartoffset swap in the first hunk reads as equivalent to me, since startoffset <= oldstartoffset holds in that branch.

As for tool vs manual, a bit of both: I use some tooling for triage and sanity checking, but the fix was worked out by hand from the ASAN trace. Satisfied this addresses everything I hit, and the tests were a real gap in my PR, so thanks for filling that in.

PR #943, opened 2026-08-10
pcre2test: correctly set history prompt with readline()

carenas:
when building pcre2test with either GNU readline or libedit, the prompt will break as shown in (press up, down, up in the second prompt, which now includes some of the previous line as the prompt):

```
PCRE2 version 10.47 2025-10-21 (8-bit)
  re> 12345678901234567
** Invalid pattern delimiter '1' (x31).
  re> 12345678912345678901234567
```

### DIFF
--- a/doc/html/pcre2test.html
+++ b/doc/html/pcre2test.html
@@ -214,9 +214,11 @@ If an unknown option is given, an error message is output; the exit code is 0.
 </p>
 <p>
 <b>--colo[u]r[=&#60;always,auto,never&#62;]</b>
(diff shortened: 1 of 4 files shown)


NWilson:
Nice! Thanks a lot Carlo.

I found an additional issue with libedit: it chews up the prompt when the escape sequence is the final character, and this is a documented restriction. I moved the final space character out of the escapes.

I also added a non-empty test to NO_COLOR (https://no-color.org/), and synced the docs.

PR #946, opened 2026-08-12
pcre2test: honor no_jit when used with jitfast

carenas:
teach `pcre2test` to avoid calling `pcre2_jit_match()` directly if `no_jit` was requested, making the use of both options work as expected in recent tests, and improve those so they trigger correctly.

### DIFF
--- a/src/pcre2test_inc.h
+++ b/src/pcre2test_inc.h
@@ -3165,7 +3165,8 @@ for (;;)
       dat_context, dfa_workspace, DFA_WS_DIMENSION);
     }
 
-  else if ((pat_patctl.control & CTL_JITFAST) != 0)
+  else if ((pat_patctl.control & CTL_JITFAST) != 0 &&
+           (dat_datctl.options & PCRE2_NO_JIT) == 0)
     capcount = pcre2_jit_match(compiled_code, pp, ulen, dat_datctl.offset,
       dat_datctl.options, match_data, dat_context);
 
@@ -4747,7 +4748,8 @@ if (dat_datctl.replacement[0] != MOD_STR_UNSET)
         dat_datctl.offset, dat_datctl.options, match_data,
         use_dat_context, dfa_workspace, DFA_WS_DIMENSION);
       }
-    else if ((pat_patctl.control & CTL_JITFAST) != 0)
+    else if ((pat_patctl.control & CTL_JITFAST) != 0 &&
+             (dat_datctl.options & PCRE2_NO_JIT) == 0)
       {
       (void)pcre2_jit_match(compiled_code, pp, arg_ulen, dat_datctl.offset,
         dat_datctl.options, match_data, use_dat_context);
@@ -4990,7 +4992,7 @@ for (gmatched = 0;; gmatched++)
   /* When matching is via pcre2_match(), we will detect the use of JIT via the
   stack callback function. */
 
-  jit_was_used = (pat_patctl.control & CTL_JITFAST) != 0;
+  jit_was_used = FALSE;
 
   /* Do timing if required. */
 
@@ -5021,7 +5023,8 @@ for (gmatched = 0;; gmatched++)
         }
       }
 
-    else if ((pat_patctl.control & CTL_JITFAST) != 0)
+    else if ((pat_patctl.control & CTL_JITFAST) != 0 &&
+             (dat_datctl.options & PCRE2_NO_JIT) == 0)
       {
       start_time = clock();
       for (i = 0; i < timeitm; i++)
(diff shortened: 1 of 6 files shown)


zherczeg (approved):
LGTM

src/pcre2test_inc.h
```diff
@@ -5208,30 +5216,21 @@ for (gmatched = 0;; gmatched++)
       }
 
     /* If PCRE2_COPY_MATCHED_SUBJECT was set, check that things are as they
-    should be, but not for fast JIT, where it isn't supported. */
+    should be. */
 
     if ((dat_datctl.options & PCRE2_COPY_MATCHED_SUBJECT) != 0)
       {
-      if ((pat_patctl.control & CTL_JITFAST) != 0)
```
NWilson, line 5215: I liked this previous code.

What happens when you request copy_matched_subject AND JIT? pcre2test should pass those options to the PCRE2 API in order to be able to exercise what would happen if a customer did that.

So I like what you did with the test - but I we should retain the following case:

```
/abc/jitfast
    abc\=copy_matched_subject
```

carenas: we already have that combination covered by testinput2:

```
/foobar/
    the foobar thing\=copy_matched_subject
    the foobar thing\=copy_matched_subject,zero_terminate

/foobar/g
    the foobar thing foobar again\=copy_matched_subject
```

but I'll add more tests

NWilson: It's the jitfast + copy_matched_subject case which (I thought) we didn't have covered.

src/pcre2test_inc.h
```diff
@@ -3165,7 +3165,8 @@ for (;;)
       dat_context, dfa_workspace, DFA_WS_DIMENSION);
     }
 
-  else if ((pat_patctl.control & CTL_JITFAST) != 0)
+  else if ((pat_patctl.control & CTL_JITFAST) != 0 &&
+           (dat_datctl.options & PCRE2_NO_JIT) == 0)
```
NWilson, line 3169: This change is good. The no-jit exclusion makes sense to me.

src/pcre2test_inc.h
```diff
@@ -5208,30 +5216,21 @@ for (gmatched = 0;; gmatched++)
...
-        if ((match_data->flags & PCRE2_MD_COPIED_SUBJECT) != 0)
-          cfprintf(clr_test_error, outfile,
-            "** PCRE2 error: flag set after unsupported copy_matched_subject\n");
-        }
-      else
-        {
-        if ((match_data->flags & PCRE2_MD_COPIED_SUBJECT) == 0)
-          cfprintf(clr_test_error, outfile,
-            "** PCRE2 error: flag not set after copy_matched_subject\n");
+      if ((match_data->flags & PCRE2_MD_COPIED_SUBJECT) == 0)
```
NWilson: This block of changes here looks wrong. Should this be reverted back?

What happens when you just do:

```
/abc/jitfast
  abc\=copy_matched_subject
```

The code I put in before said "copy_matched_subject does nothing when jitfast is active" which is true. Now you've put it back to asserting that copy_matched_subject works, regardless of jitfast?

src/pcre2test_inc.h
```diff
@@ -6043,6 +6044,36 @@ if (test_compiled_with_jit)
 
   pcre2_match_data_free(test_match_data);
   test_match_data = NULL;
+
+
+/* Check that fast JIT ignores PCRE2_COPY_MATCHED_SUBJECT */
+
+  test_match_data = pcre2_match_data_create(100000, test_gen_context);
```
NWilson, line 6067: We don't need these tests here, do we?

I don't see why this sequence of tests can't be done in the testinput17 file, rather than hardcoded.

src/pcre2test_inc.h
```diff
@@ -4990,7 +4992,7 @@ for (gmatched = 0;; gmatched++)
   /* When matching is via pcre2_match(), we will detect the use of JIT via the
   stack callback function. */
 
-  jit_was_used = (pat_patctl.control & CTL_JITFAST) != 0;
+  jit_was_used = FALSE;
```
NWilson, line 4995: I'm not convinced by this. The code claims that the callback function detects the use of JIT. However there are options that make it so that we don't install the callback.

I think we want this condition to be `jit_was_used = JITFAST && !NO_JIT`.

Here's a possible test case:

```
# Verify JIT reporting when the match context is NULL.

/abc/jitfast
    abc\=null_context
    abc\=null_context,no_jit
```

src/pcre2test_inc.h
carenas: at least the one that checks that PCRE2_NO_JIT is ignored by `pcre2_jit_match()` can't be reached other way

src/pcre2test_inc.h
carenas: using the callback is the documented (see `pcre2it.3`) way to detect that JIT gets called, but you are correct that it needs a context, so fixed that by adding a fallback.

NWilson (approved):
Thanks Carlo, I like it!

Issue #867, opened 2025-12-25
Partial DFA match semantics: improve docs or deprecate feature


sjshuck:
I apologize for the vague title.  I opened an issue in 2021, bug 2761 in the old exim issue tracker, and I don't have access to that.  I wanted to keep tracking it here as I've had that in my list of things to poke at since then.

I have emails from exim with Philip's and others' contributions to that thread, but not the whole thread (and especially not my own OP).

NWilson:
I'm afraid I don't have access to that either. 

Are you able to reproduce or reconstruct an outline of the problem?

sjshuck:
I can't reproduce it right now but I think I was trying to match over a chunk boundary using `PCRE2_PARTIAL_HARD` and `PCRE2_PARTIAL_SOFT` and neither was fulfilling my expectation "of achieving the same results that would happen if the entire string was available for searching all the time" (quoting [the docs](https://pcre.org/current/doc/html/pcre2partial.html#TOC1)).  Here's a quote from Philip in the bug thread:

> Consider the pattern `/1234|3789/`. If the first segment is "`ABC123`" there is a partial match for "`1234`". There is no partial match for the second alternative, because such a match does not start at the same point in the subject string. Attempting to continue with the string "`7890`" does not yield a match because only those alternatives that match at one point in the subject are remembered.

In short, even assuming that the user held onto all chunks comprising a prospective match, it's unclear which offset to jump back to to retry the match at an advanced starting position.  Non-partial matching takes care of retrying automatically.  The partial matching documentation doesn't indicate this extra responsibility of the user or how best to accomplish it.

I'm not sure what the issue is here, whether just docs or an actual technical impossibility given the current implementation.  I'd have to sit down and attempt a PoC to know.  For now I'll just leave it with these comments in the thread from Philip and Petr Pisar, respectively, which seem to contradict the above quoted docs:

> For multi-segment matching it is probably best not to make use of the "restart"
feature at all, but instead to retain the partially matched string and do a
normal match when some more text has been joined on. You can do this with DFA
or non-DFA matching.

> An inability to remember an arbitrary long string
is an inherent limitation of any deterministic finite automaton. If a users
needs to remember the strings, he needs to use a different matching algorithm
than DFA.

PhilipHazel:
[...]

PR #947, opened 2026-08-12
doc: say that PCRE2_DFA_RESTART does not remove the need to retain data

karpovantonme:
Follow-up to #867

the multi-segment section for `pcre2_dfa_match()` explains the `1234|3789` limitation, but it still reads as if restarting frees the caller from keeping earlier text. It does not: a lookbehind reaches back into the segment that was discarded, and the bound on how much to keep is the same `PCRE2_INFO_MAXLOOKBEHIND` that the `pcre2_match()` section above already describes

so the added paragraph says that, and points at retaining a bounded tail and running a fresh match as the approach that achieves the same results as searching the whole subject at once. Nothing new is claimed here, the wording follows what was concluded in #867

two files: `doc/pcre2partial.3`, and the generated `doc/html/pcre2partial.html` regenerated with `maint/132html` the way `maint/UpdateAlways` calls it. `maint/CheckMan` is clean on the page

if you would rather deprecate `PCRE2_DFA_RESTART` outright, say so and I will close this

### DIFF
--- a/doc/html/pcre2partial.html
+++ b/doc/html/pcre2partial.html
@@ -389,15 +389,18 @@ followed by "no match" when PCRE2_DFA_RESTART is used on the second buffer, you
 can then try a new match starting at offset <i>n+1</i> in the first buffer.
 </p>
 <p>
-Note that PCRE2_DFA_RESTART does not remove the need to keep earlier subject
-data. A lookbehind assertion still refers to characters that PCRE2 no longer
-holds, so the amount that must be retained is bounded in the same way as for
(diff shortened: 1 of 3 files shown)


NWilson:
Thank you very much! I have made a couple of minor corrections, but your text is helpful.

karpovantonme:
Thanks for taking it and for the corrections 🙏🏼

rebased onto main, your commit kept as is. the red Bazel job on windows was a `503 Service Unavailable` fetching `bazel-skylib-1.8.2.tar.gz` rather than anything in the diff, so the fresh run should be clean once it gets an approval to start

PR #857, opened 2025-12-01
Fix regression in handling of repeated backslash escapes in pcre2_substitute

NWilson:
Fixes #854

PR #948, opened 2026-08-12
keep check_escape error offset within the pattern for \x{ and \o{

sage-mode-hunter:
An unterminated \x{ or \o{ escape at the end of a pattern makes check_escape report an error offset past the end of the pattern:
- ESCAPE_FAILED_FORWARD increments ptr unconditionally, but the \x{ and \o{ scanners reach it with ptr already at ptrend
- pcre2_compile then returns erroroffset = patlen + 1 (offset 5 for the 4-unit pattern \x{2), out of range for the caller
- with --enable-debug this trips PCRE2_ASSERT(ptr <= pattern + patlen); the \c paths sharing this label only arrive with ptr < ptrend, so they stay correct
Guarded the increment against ptrend and added a regression to test 2.

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -2020,7 +2020,7 @@ else
     else
       {
       *errorcodeptr = ERR64;
-      goto ESCAPE_FAILED_FORWARD;
+      if (ptr < ptrend) goto ESCAPE_FAILED_FORWARD;
       }
     break;
 
@@ -2107,7 +2107,7 @@ else
         else
           {
           *errorcodeptr = ERR67;
-          goto ESCAPE_FAILED_FORWARD;
+          if (ptr < ptrend) goto ESCAPE_FAILED_FORWARD;
           }
         }   /* End of \x{} processing */
 
@@ -2224,10 +2224,7 @@ return escape;
 /* Some errors need to indicate the next character. */
 
 ESCAPE_FAILED_FORWARD:
-/* Point past the offending character, but not past the end of the pattern. The
-\x{ and \o{ scanners can reach here with ptr at ptrend, and the reported error
-offset must stay within the pattern. */
-if (ptr < ptrend) ptr++;
+ptr++;
 #ifdef SUPPORT_UNICODE
 if (utf) FORWARDCHARTEST(ptr, ptrend);
 #endif
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -2224,7 +2224,10 @@ return escape;
 /* Some errors need to indicate the next character. */
 
 ESCAPE_FAILED_FORWARD:
-ptr++;
+/* Point past the offending character, but not past the end of the pattern. The
+\x{ and \o{ scanners can reach here with ptr at ptrend, and the reported error
+offset must stay within the pattern. */
+if (ptr < ptrend) ptr++;
 #ifdef SUPPORT_UNICODE
 if (utf) FORWARDCHARTEST(ptr, ptrend);
 #endif
--- a/testdata/testinput2
+++ b/testdata/testinput2
@@ -3987,6 +3987,10 @@
 
 /^A\x{/
 
+/\x{2/
+
+/\o{7/
+
 /[ab]++/B,no_auto_possess
 
 /[^ab]*+/B,no_auto_possess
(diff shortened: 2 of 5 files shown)


testdata/testinput2
```diff
@@ -8351,4 +8351,11 @@ a)"xI
     abc\=ps,substitute_replacement_only,replace=>$_<
     abc\=ps,substitute_replacement_only,replace=>$'<
```
carenas, line 8357: move to 3984 with the other similar tests

carenas (changes requested):
interesting finding.

note that an offset + 1 of the pattern length is still a valid erroroffset though, see #460 for when those asserts were added

testdata/testinput5
```diff
@@ -3616,4 +3616,9 @@
     \x{e1}
     \x{c1}
 
+# An unterminated \N{U+ escape at the end of the pattern must
+# report an error offset that lies within the pattern.
+
+/(*UTF)\N{U+2/
```
NWilson, line 3622: This is the third construct `\N{...` which his the same error.

src/pcre2_compile.c
```diff
@@ -2020,7 +2020,7 @@ else
     else
       {
       *errorcodeptr = ERR64;
-      goto ESCAPE_FAILED_FORWARD;
+      if (ptr < ptrend) goto ESCAPE_FAILED_FORWARD;
```
NWilson, line 2023: Actually, I feel that the bug is doing `goto ESCAPE_FAILED_FORWARD` when we haven't checked that there is a character to consume.

There are four `goto ESCAPE_FAILED_FORWARD` instances. Two of them are correct. It's just these two here that were missing the check.

NWilson (approved):
Super-helpful! Thank you very much for finding and reporting this, and sending in a fix.

src/pcre2_compile.c
carenas: it is difficult to see, but it seems that most of the issues were part of the refactoring to sync the erroroffsets with Perl, so this should cover all those 4 goto, right?

```C
diff --git a/src/pcre2_compile.c b/src/pcre2_compile.c
index c64fd273..aedcf62c 100644
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -2224,10 +2224,13 @@ return escape;
 /* Some errors need to indicate the next character. */
 
 ESCAPE_FAILED_FORWARD:
-ptr++;
+if (ptr < ptrend)
+  {
+  ptr++;
 #ifdef SUPPORT_UNICODE
-if (utf) FORWARDCHARTEST(ptr, ptrend);
+  if (utf) FORWARDCHARTEST(ptr, ptrend);
 #endif
+  }
 goto EXIT;
 }
```
the logic for other similar escapes (ex: verbs) is also affected, and looking at the logic we had in 10.45 doesn't seem to show the reported issues.

NWilson: As I said, there are four gotos, and two of them are correct. It was just these two that needed fixing.

testdata/testinput2
sage-mode-hunter: Done, moved them up next to the /^A\x{/ group (with the matching output block).

src/pcre2_compile.c
sage-mode-hunter: Yes, that covers all four, and it should be behaviorally identical: the two \c sites can only reach the label with ptr < ptrend (\c bails out earlier with ERR2 when the pattern ends), so guarding at the label vs at the two broken call sites gives the same offsets either way. I kept the call-site checks per Neil's point that the bug is jumping forward with nothing left to consume, but I don't feel strongly, either shape works.

You're right about the history: this came in with the erroroffset rework in #857, so 10.45 is unaffected. I also probed the truncated verb forms ((*MARK, (*MARK:ab, (*:ab, (*LIMIT_HEAP=4) against a debug build and they all report offset == patlen, which is still a valid offset per the asserts, so those look fine as-is.

NWilson: Thank you Karthik.

The `ESCAPE_FAILED_FORWARD` gotos should behave the same as the `FAILED_FORWARD` (non-escape) ones. And those are all correct - I've just re-reviewed them. The pattern used elsewhere is to guard the goto.

The actual code at the `ESCAPE_FAILED_FORWARD` label should match the code at the `FAILED_FORWARD` label too.

I'm satisfied we don't have any further bugs here, and that the testcases cover everything.

sage-mode-hunter: Sounds good. The two label bodies already line up (both do ptr++ then FORWARDCHARTEST under SUPPORT_UNICODE, differing only in the exit target), so I've left the branch as is.

PR #949, opened 2026-08-12
maint: add a match throughput benchmark driver

mattst88:
Adds a benchmark driver for match throughput, `maint/pcre2bench.c`, plus
`maint/RunBenchmark` to compile and run it.

## Why

Work on the JIT's SIMD fast-forward paths for a new architecture needs a
yardstick that is the same from one architecture to the next, and that says
which of the `JIT_HAS_FAST_*` hooks a given change actually moves. I wrote this
while adding an Alpha SIMD block, where the question under review was how much
of the new code paid for itself; the answer came from disabling each hook in
turn and re-running. That is a question anyone adding a fast-forward
implementation will have to answer, so the driver seemed worth sharing rather
than keeping locally.

## What it does

Builds a 4MB deterministic pseudo-text corpus and scans it end to end with 18
patterns, reporting best-of-five throughput as TSV, one row per pattern. The
patterns separate the fast-forward paths from everything else: absent first
code unit (pure fast-forward cost, no false starts), character ranges, ranges
at a relative offset, caseless, character pairs, a few realistic mixed
patterns, and two controls that use no fast-forward at all as a noise floor.

A `short` mode slices the same corpus into 8 to 512 byte runs and reports
per-call cost instead, which is where the fixed prologue and loop-setup cost
shows up. A `nojit` mode runs the interpreter for comparison.

The corpus comes from a fixed xorshift64* seed, so it is bit-identical on every
host and build. Every row carries its match count, and the driver aborts if the
count moves between repeats, so a build under test can be checked against a
reference for identical results rather than only for speed.

Typical use:

    maint/RunBenchmark build-before before > before.tsv
    maint/RunBenchmark build-after  after  > after.tsv

## Decisions I would rather flag than have you find

**Not wired into any of the four build systems.** `maint/ucptest.c` sets the
precedent of a maintainer C program that you compile by hand, and adding a
target to autotools, CMake, Bazel and Zig for a tool like this seemed a poor
trade. `RunBenchmark` locates the generated `pcre2.h` and `libpcre2-8` under a
given build directory, handling both the CMake and autotools layouts, so nobody
has to get the include and link flags right by hand.

**Not in CI, and I do not think it should be.** Shared runners are noise
dominated. For calibration, on a dedicated idle machine I measured up to 17.6%
run-to-run spread on the noisiest pattern in this set. Anything CI could report
from that would be false signal.

**Will not build on MSVC.** It uses `clock_gettime(CLOCK_MONOTONIC)`, which
UCRT does not provide. `RunBenchmark` is `/bin/sh`, so the tool is Unix-only in
practice anyway, but `maint/` does carry `.ps1` scripts, so I would rather say
so than let you assume otherwise. Happy to add a `QueryPerformanceCounter`
branch if you want it.

**`RunBenchmark` passes `-Wl,-rpath`.** Fine on ELF and Mach-O, not universal.
It only matters for shared builds.

## Impact on the tree

None outside `maint/`. `maint/` is not in `EXTRA_DIST`, so these files do not
reach the tarball and `manifest-tarball` is unaffected; the `RunManifestTest`
invocations in CI check install directories, not the source tree. No job
compiles or lints `maint/` contents. No ChangeLog entry, matching the other
recent infrastructure commits.

## Testing

Built the driver `-Wall -Wextra` clean and ran all four modes against an x86-64
static JIT build. Match counts are stable across repeats on every pattern, and
back-to-back runs of the same binary agree to within 1.3%.

### DIFF
--- a/maint/pcre2bench.c
+++ b/maint/pcre2bench.c
@@ -1,0 +1,571 @@
+/*************************************************
+*      PCRE2 match throughput benchmark driver   *
+*************************************************/
+
+/* This program measures whole-buffer scan throughput for a set of patterns
+chosen to exercise the JIT's start-of-match fast-forward paths (single
+character, character pair, character range, caseless) as well as control cases
+that bypass fast-forward entirely.  It exists to compare two builds of the
+library, for example before and after a change to the SIMD fast-forward code
+for one architecture.
+
+The subject is a deterministic pseudo-text corpus, generated identically on
+every host and build, so runs from different builds can be compared directly.
+Output is TSV on stdout.  The match count is reported for each pattern and must
+be identical across every build under test.
+
+Usage: pcre2bench [label] [nojit] [short]
+
+  label   a string used in the first output column; defaults to "build"
+  nojit   run the interpreter with PCRE2_NO_JIT instead of the JIT
+  short   run the short-subject sweep instead of the throughput benchmarks
+
+The code unit width is chosen at compile time by defining
+PCRE2_CODE_UNIT_WIDTH to 8, 16 or 32; it defaults to 8.  All the text here is
+ASCII, and the corpus is a fixed number of code units rather than of bytes, so
+the same corpus is generated at every width and the match counts are directly
+comparable between them.  Throughput is therefore reported in code units, not
+in bytes, and the width appears as an output column.
+
+This program is not built by any of the PCRE2 build systems.  Compile it
+against an already-built tree with
+
+  cc -O2 -DPCRE2_CODE_UNIT_WIDTH=8 -o pcre2bench maint/pcre2bench.c \
+    -I<build>/src -L<build>/.libs -lpcre2-8
+
+or use maint/RunBenchmark, which works the include and link flags out for a
+CMake or an autotools build tree and then runs the result. */
+
+#ifndef PCRE2_CODE_UNIT_WIDTH
+#define PCRE2_CODE_UNIT_WIDTH 8
+#endif
+
+#include <pcre2.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+
+#ifndef SUBJECT_UNITS
+#define SUBJECT_UNITS (4u * 1024u * 1024u)
+#endif
+
+#define MIN_SECONDS 0.5
+#define REPEATS 5
+
+/* Longest pattern or error message handled by widen() and narrow(). */
+
+#define TEXT_MAX 256
+
+static double
+now_seconds(void)
+{
+struct timespec ts;
+clock_gettime(CLOCK_MONOTONIC, &ts);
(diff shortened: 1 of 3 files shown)


maint/pcre2bench.c
```diff
@@ -0,0 +1,520 @@
...
+  pcre2_match_data *match_data;
+  long matches;
+  double best = 0.0;
+  int rep;
+
+  code = pcre2_compile((PCRE2_SPTR)benchmarks[i].pattern,
+    PCRE2_ZERO_TERMINATED, 0, &errorcode, &erroroffset, NULL);
+  if (code == NULL)
+    {
+    unsigned char message[256];
```
carenas: PCRE2_UCHAR message[128];

carenas (commented):
FWIW for portability we use clock() in `pcre2test`, but since this is a "maint" tool this is likely better.

the build script is IMHO not needed, as you could just put a comment of how to compile it, like we use in ucptest, but it doesn't do harm either.

IMHO, it would be better if it would had been PCRE2_CODE_UNIT_WIDTH agnostic.

NWilson:
Thanks Matt! This is another PR where I think I'll just pause it for a week until after 10.48 is out.

But many thanks, I'll look at it soon

mattst88:
> Thanks Matt! This is another PR where I think I'll just pause it for a week until after 10.48 is out.

Totally fine. Absolutely no rush from my side. Thanks!

maint/pcre2bench.c
mattst88: Fixed. `PCRE2_UCHAR message[TEXT_MAX]`, with `TEXT_MAX` as the size argument rather than `sizeof(message)` — that parameter counts code units, not bytes, so the old call overstated the buffer by 4x at width 32.

mattst88:
All three taken.

Width: the driver is now `PCRE2_CODE_UNIT_WIDTH` agnostic, defaulting to 8. Its text is all ASCII, so patterns are widened on the way into the library and error messages narrowed on the way out. The corpus is a fixed number of code units rather than of bytes, so the same text is generated at every width; 8, 16 and 32 give identical match counts in both modes, with and without the JIT. Throughput is reported in code units (`mcu_per_s`, `ns_per_cu`) and the width is an output column, so two TSVs can't be compared across widths by accident. `RunBenchmark` takes `-8`, `-16`, `-32`.

Build script: added the compile line to the header comment, as `ucptest.c` has. Kept the script, since what it does is find the generated `pcre2.h` and the library under either a CMake or an autotools layout.

clock(): left as `clock_gettime(CLOCK_MONOTONIC)`. `clock()` measures CPU time; what's reported here is throughput, so wall time is the right figure.

NWilson (approved):

NWilson:
This is great, thanks Matt! It's likely that I'll be doing a lot more benchmarking in the future, since I have grand plans for refactoring piece-by-piece. This will be a good foundation for a perf test suite.

PR #951, opened 2026-08-14
pcre2grep: make path traversing functions static everywhere

carenas:


NWilson (approved):

NWilson:
It doesn't matter much for binaries like pcre2grep and pcre2test, but it's always worth having good hygiene.

Thanks!

PR #954, opened 2026-08-20
guard end-before-start match length in pcre2grep -o and --output

sage-mode-hunter:
**Length underflow on an end-before-start match in pcre2grep**

With `--allow-lookaround-bsk`, a \K inside a lookaround can leave a match whose end offset is before its start (`(?=foo\K)` on `XfooY` gives ovector 4..1). The -o and --output paths take the printed length as an unsigned `end - start`, which wraps near SIZE_MAX and runs print_match/fwrite off the end of the line buffer. The colour and multiline paths already swap for this exact case; these two did not. Guarded both so a reversed match counts as empty, which is what the existing `plen > 0` / `capturesize > 0` tests were already after.

Repro under ASAN (small --buffer-size to land on the redzone):
```
printf 'XfooY\n' | pcre2grep --buffer-size=24 --allow-lookaround-bsk -o '(?=foo\K)' -
printf 'XfooY\n' | pcre2grep --buffer-size=24 --allow-lookaround-bsk --output '$0' '(?=foo\K)' -
```
Both fault in the length-based write (pcre2grep.c:2889 for -o, :2216 for --output) before the patch and run clean after. Added test 161 to RunGrepTest.


src/pcre2grep.c
NWilson: Here's one needing the same treatment.

Create local variables, swap if not ordered.

NWilson: This usage too.

NWilson: This should be testing which is larger, and doing the reversed subtraction if they are not ordered.

NWilson: Another one here to swap/flip.

NWilson (commented):
This is great, thank you!

However... we do have a convention already in a couple of other places in pcre2grep, where these pathological "reversed matches" are just printed as `[min(start,end) ... max(start,end)]`.

Would it be OK to swap these two you found to behave the same, just for consistency?

I also asked GPT to check for others, and it found some. It should be quick & easy (hopefully!) to make the same change in each place.

sage-mode-hunter:
Makes sense, switched both to the swap convention and did the same in the four spots you marked (--line-offsets, --file-offsets, and the callout argslen/memcpy pair). For the argslen one I kept it as a conditional subtraction since it's only a length, the rest use start/end locals with the same swap block as the colour path. Test 161 now expects "foo" and covers --line-offsets / --file-offsets too. RunGrepTest is green with and without ASAN.

src/pcre2grep.c
sage-mode-hunter: Done, start/end locals with a swap, shared by both offset options.

sage-mode-hunter: Covered by the same block as --line-offsets now.

sage-mode-hunter: Done, reversed subtraction when end < start so it agrees with the memcpy pass below.

sage-mode-hunter: Swapped here too.

NWilson (approved):
Amazing, thank you very much!

NWilson:
I pushed a small commit adding the matching update `RunGrepTest.bat`

PR #957, opened 2026-08-22
Update sljit compiler

zherczeg:
The last update was 11 months ago. Let's see if the latest compiler causes any problems.


NWilson:
Those build failures on `Dev / Clang` are known - I just merged the PR fixing them.

This all looks good - go ahead and merge if you're happy with it!

PR #958, opened 2026-08-22
Add support for installing PDBs with CMAKE_SHARED_LIBRARY_PREFIX_C

moyer-avs:
Hello,

We are building pcre2 with a custom prefix for shared libraries on Windows.
We use `CMAKE_SHARED_LIBRARY_PREFIX_C` to set the prefix, resulting in both the dll and pdb being prefixed with the content of the variable.

The pdb installation then fails, because the prefix is not taken into account in the `install()` command.

This PR intends to fix it.

```bash
cd ./pcre2
cmake -G Ninja -DCMAKE_BUILD_TYPE=Debug \
    -DBUILD_STATIC_LIBS=OFF \
    -DINSTALL_MSVC_PDB=ON \
    -DBUILD_SHARED_LIBS=ON \
    -DCMAKE_SHARED_LIBRARY_PREFIX_C="prefix." \
    -B build
cmake --build build/
cmake --install build/ --prefix C:/test/pcre2
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -910,7 +910,6 @@ endif()
 
 set(TARGETS)
 set(DLL_PDB_FILES)
-set(DLL_PDB_DEBUG_FILES)
 
 # 8-bit library
 
@@ -988,8 +987,7 @@ if(PCRE2_BUILD_PCRE2_8)
       set_target_properties(pcre2-8-shared PROPERTIES LINK_DEPENDS ${PROJECT_BINARY_DIR}/src/libpcre2-8.sym)
     endif()
     list(APPEND TARGETS pcre2-8-shared)
-    list(APPEND DLL_PDB_FILES $<TARGET_PDB_FILE_DIR:pcre2-8-shared>/pcre2-8.pdb)
-    list(APPEND DLL_PDB_DEBUG_FILES $<TARGET_PDB_FILE_DIR:pcre2-8-shared>/pcre2-8d.pdb)
+    list(APPEND DLL_PDB_FILES $<TARGET_PDB_FILE:pcre2-8-shared>)
 
     add_library(pcre2-posix-shared SHARED ${PCRE2POSIX_HEADERS} ${PCRE2POSIX_SOURCES})
     target_include_directories(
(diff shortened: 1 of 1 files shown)


NWilson:
Thank you! I can see how and why this is failing.

We last changed this configuration in PR #370, where the author commented on the existence of `$<TARGET_PDF_FILE:target>`. However I didn't pick up on it realise the problem. The current situation is a bit of a patch on top of the previous system, which was also a bit broken.

Basically, the CMake is file is "guessing" what the exact filename is going to be for the PDB file. By adding CMOKE_SHARED_LIBRARY_PREFIX_C, you've improved the guess, but there are _many_ other CMake properties the user could have set, to customise the file basenames, prefixes, suffixes, and so on.

Would you be able to fix it to request the _actual_ pathname from CMake?

Replace:

```diff
-    list(APPEND DLL_PDB_FILES $<TARGET_PDB_FILE_DIR:pcre2-8-shared>/pcre2-8.pdb)
-    list(APPEND DLL_PDB_DEBUG_FILES $<TARGET_PDB_FILE_DIR:pcre2-8-shared>/pcre2-8d.pdb)
+    list(APPEND DLL_PDB_FILES "$<TARGET_PDB_FILE:pcre2-8-shared>")
```

We only need to track one filename - and CMake shouldn't ever give us the wrong one!

Delete all references to `DLL_PDB_DEBUG_FILES`.

Then, we only need to install them when they exist (in Debug and RelWithDebInfo configurations):

```cmake
if(MSVC AND INSTALL_MSVC_PDB)
  install(
    FILES ${DLL_PDB_FILES}
    DESTINATION ${CMAKE_INSTALL_BINDIR}
    CONFIGURATIONS Debug RelWithDebInfo
  )
endif()
```

moyer-avs:
I tested the proposed changes, they work as expected.
I amended my commit.

NWilson:
Thank you! I will do some testing of my own, and merge.

NWilson:
I have tested with the four different build types on Windows; using INSTALL_MSVC_PDB and with/without your CMAKE_SHARED_LIBRARY_PREFIX_C.

Everything appears to work.

Issue #959, opened 2026-08-23
alt_extended_class + lookbehind: wrong error 125 and internal error 170 crash


YuFuSweetie:
# alt_extended_class + lookbehind: wrong error 125 and internal error 170 crash

## Environment

- PCRE2 10.47 (2025-10-21)
- Windows 11, 8-bit library, UTF+UCP enabled
- Option: `PCRE2_ALT_EXTENDED_CLASS` (UTS#18 compatible character classes)

## Bug 1: UTS#18 extended class + lookbehind anywhere in the pattern → compilation fails, pcre2test aborts

When `PCRE2_ALT_EXTENDED_CLASS` is enabled and the pattern uses a UTS#18 extended class (e.g. difference `--` or union `||`), the presence of any lookbehind assertion in the pattern causes compilation to fail with an internal error, and pcre2test aborts the whole run ("flash exit").

### Pattern 1: difference class + negative lookbehind

```
/[\d--1]a(?<!x)/alt_extended_class,utf,ucp
    a
```

Result:

```
Failed: error 170 at offset 0: internal error: unknown meta code in check_lookbehinds()
** Error code 170 not implemented in error_direction().
   error_direction() should usually return '1' for newly-added errors,
   and the offset should be just to the right of the bad character.
** pcre2test run abandoned
```

### Pattern 2: union class + empty lookbehind

```
/[a||b](?<=)/alt_extended_class,utf,ucp
    a
```

Result: same error 170, pcre2test run abandoned.

## Bug 2: nested character class used in a lookbehind → error 125 "length not limited"

When a nested character class (union of classes, e.g. `[12[34]]`) is used in a lookbehind assertion, compilation fails with error 125, although the class matches a single character and is clearly fixed-length.

### Pattern

```
/(?<=[12[34]])/alt_extended_class,utf,ucp
    1
```

Result:

```
Failed: error 125 at offset 0: length of lookbehind assertion is not limited
        here: |-->| (?<=[12[34...
```

carenas:
This also affects Perl's extended character classes (AKA regex_sets), for example:

```
/(?[\d-[1]])a(?<!x)/
```

NWilson:
Great bug report! I'm working on the fixes right now.

NWilson:
Fixed in #961, #962

PR #961, opened 2026-08-23
Fix compilation of patterns incorporating ECLASS and lookbehinds

NWilson:
Reported in #959

### DIFF
--- a/src/pcre2_compile.c
+++ b/src/pcre2_compile.c
@@ -10167,6 +10167,11 @@ for (; *pptr != META_END; pptr++)
     case META_COMMIT:
     case META_DOLLAR:
     case META_DOT:
+    case META_ECLASS_AND:
+    case META_ECLASS_NOT:
+    case META_ECLASS_OR:
+    case META_ECLASS_SUB:
+    case META_ECLASS_XOR:
     case META_FAIL:
     case META_PLUS:
     case META_PLUS_PLUS:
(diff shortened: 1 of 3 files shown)


carenas (commented):
I had the following tests instead in my version, which are IMHO more realistic and cover also the perl class syntax:

```
diff --git a/testdata/testinput1 b/testdata/testinput1
index 64b697f3..91e2c7a1 100644
--- a/testdata/testinput1
+++ b/testdata/testinput1
@@ -7154,6 +7154,13 @@ $/x
 \= Expect no match
     z
 
+/(?[\d-[1]]).(?<!x)/
+    0a
+    90
+\= Expect no match
+    1a
+    0x
+
 /^.{4}/s
     abcdef
     abcde
diff --git a/testdata/testinput2 b/testdata/testinput2
index a7c1c846..c205e68e 100644
--- a/testdata/testinput2
+++ b/testdata/testinput2
@@ -7407,6 +7407,13 @@ a)"xI
 /[^[[^]~~[^]]]/B,alt_extended_class,allow_empty_class
     a
 
+/[\d--1].(?<!x)/alt_extended_class
+    0a
+    90
+\= Expect no match
+    1a
+    0x
+
 # --------------
 
 ```

NWilson:
Thanks @carenas! That's a good testcase, I've added it to my PR.

carenas (approved):
the Perl compatible extended class syntax test could have gone to test1 as was originally suggested, though

NWilson:
Most of the perl eclass tests are in testinput2. I put it with the existing block.

PR #962, opened 2026-08-23
Fix compilation of ECLASS inside lookarounds

NWilson:
Reported in #959

### DIFF
--- a/testdata/testinput2
+++ b/testdata/testinput2
@@ -7400,6 +7400,16 @@ a)"xI
 
 /[z&/alt_extended_class
 
+/(?<=[12[34]])/alt_extended_class
+    1
+    2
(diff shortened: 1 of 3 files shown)


testdata/testinput2
```diff
@@ -7020,6 +7020,13 @@ a)"xI
 
 # EXTENDED CHARACTER CLASSES (UTS#18)
 
+/(?<=[12[34]])/alt_extended_class
+    1
+    2
+    3
+    4
```
carenas: \= Expect no match

The following should also fail to match:

1]
[]

NWilson: OK, added

carenas (approved):

PR #963, opened 2026-08-24
fix stale byte in negated glob class with glob_no_wild_separator

sage-mode-hunter:
Converting a negated glob class with PCRE2_CONVERT_GLOB_NO_WILD_SEPARATOR emits one extra character:
- convert_glob_parse_range writes "[^" plus the wildcard separator into the shared out_str scratch, then always emits a count of len + 1
- the no-wildcard-separator flag skips the separator store, so that count now includes a leftover byte from an earlier write into out_str
- [!a] converts to [^sa] (the "s" survives from "(?s)"), and a*[!b] to [^Cb] (from "(*COMMIT)")

Count only the units actually written. Added glob_no_wild_separator negated-class cases to test 24; they fail before the change and pass after.

### DIFF
--- a/src/pcre2_convert.c
+++ b/src/pcre2_convert.c
@@ -679,9 +679,10 @@ if (*pattern == CHAR_EXCLAMATION_MARK
       len++;
       }
     out->out_str[len] = (uint8_t) separator;
+    len++;
     }
 
-  convert_glob_write_str(out, len + 1);
+  convert_glob_write_str(out, len);
   }
 else
   convert_glob_write(out, CHAR_LEFT_SQUARE_BRACKET);
--- a/testdata/testinput24
+++ b/testdata/testinput24
@@ -330,6 +330,12 @@
 
 /??a??/
 
+/[!a]/
+
+/[^xy]/
+
+/a*[!b]/
+
 #pattern convert=unset
 #pattern convert=glob,convert_glob_escape=0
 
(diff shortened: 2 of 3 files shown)


NWilson:
I'd like to accept this fix, but I've unfortunately made the 10.48 branch, and I don't want to add any delay.

This will have to wait a week!

sage-mode-hunter:
No rush at all, happy for this to land whenever the 10.48 dust settles.

fallenmi (approved):
Reviewed exact base `92baa6cb647d89385aa9d550bc6614df881d8c2b` and exact head `4a9498de86e4378f42c932ee7ee956e0bc002d97`.

The length accounting is correct in both branches: `len` now advances only when the separator is actually written, so `glob_no_wild_separator` emits exactly `[^`, while the existing separator and escaped-separator paths retain their previous output length.

I verified the bug independently with three converter cases. The exact base produced `[^sa]`, `[^sxy]`, and `[^Cb]`; the exact head produced the intended `[^a]`, `[^xy]`, and `[^b]`. On both exact trees, PCRE2 test 24 passed for the 8-, 16-, and 32-bit libraries, and the head returned the corrected oracle output in all three widths. `git diff --check` is clean, and all 32 current non-skipped CI checks pass.

Disclosure: this review was prepared with Codex assistance; I independently verified the complete diff, exact SHAs, converter outputs, local tests, and current CI state before submitting it.

NWilson (approved):
Thank you again Karthik, and @fallenmi for the AI review.

The fix is clear and tested. Merging.

PR #966, opened 2026-08-27
pcre2test: fix off-by-one heap write in pattern expand buffer

sage-mode-hunter:
The expand pattern modifier can fill pbuffer8 to its exact size and then write the terminating zero one byte past the end:
- the growth loop at pcre2test_inc.h:2292 stops once count*length fits exactly, leaving no room for the trailing NUL stored just after at :2310
- `/\[a]{50000}/expand` (50000 is the default pbuffer8 size) overflows the buffer by one byte; ASAN reports a heap-buffer-overflow WRITE of size 1 in process_pattern
- the subject-side `\[...]{n}` replicator lower in the same file already reserves the byte with `>=`

### DIFF
--- a/src/pcre2test_inc.h
+++ b/src/pcre2test_inc.h
@@ -2289,7 +2289,7 @@ else if ((pat_patctl.control & CTL_EXPAND) != 0)
     expanding buffers always keeps buffer and pbuffer8 in step as far as their
     size goes. */
 
-    while (pt + count * length > pbuffer8 + pbuffer8_size)
+    while (pt + count * length >= pbuffer8 + pbuffer8_size)
       {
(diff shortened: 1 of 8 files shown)


NWilson:
Nice find, thank you!

Would you mind adding the testcase `/\[a]{50000}/expand` to pin the behaviour? Thanks.

sage-mode-hunter:
Added it to testinput8, next to the existing `/\[()]{65535}/expand` case. The compile result is link-size dependent here (too large at link size 2, compiles fine at 3 and 4), so it fits with the per-link-size output files, which I regenerated for all widths.

NWilson (approved):
Really nice, thanks Karthik!

PR #967, opened 2026-08-31
Add pcre2_get_subject

jcking:
Add `pcre2_get_subject` to access the copied subject when `PCRE2_COPY_MATCHED_SUBJECT` is used or the original subject. Currently it is not possible to access the copied subject when `PCRE2_COPY_MATCHED_SUBJECT` is used, making the use of `pcre2_get_ovector_pointer` and `pcre2_get_ovector_count` infeasible without creating yet another copy of the original subject.

I was not exactly sure how to go about updating the documentation or whether it would be preferred to use `pcre2_get_match_data_subject`.

### DIFF
--- a/src/pcre2.h.generic
+++ b/src/pcre2.h.generic
@@ -772,7 +772,9 @@ PCRE2_EXP_DECL PCRE2_SIZE *PCRE2_CALL_CONVENTION \
 PCRE2_EXP_DECL PCRE2_SIZE PCRE2_CALL_CONVENTION \
   pcre2_get_startchar(pcre2_match_data *); \
 PCRE2_EXP_DECL int PCRE2_CALL_CONVENTION \
-  pcre2_next_match(pcre2_match_data *, PCRE2_SIZE *, uint32_t *);
+  pcre2_next_match(pcre2_match_data *, PCRE2_SIZE *, uint32_t *); \
+PCRE2_EXP_DECL PCRE2_SPTR PCRE2_CALL_CONVENTION \
+  pcre2_get_subject(pcre2_match_data *);
 
 
 /* Convenience functions for handling matched substrings. */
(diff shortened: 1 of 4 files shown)


NWilson:
Interesting.

It seems a reasonable addition. However, I'm trying to think what the use-case would be. If you have called PCRE2 - then you _have_ the subject, so you wouldn't normally need to retrieve it.

The `PCRE2_COPY_MATCHED_SUBJECT` has fairly limited usefulness (since it makes an unnecessary copy). It does have usefulness for applications like language bindings for memory-safe languages. In those contexts, you may not be able necessarily to 'pin' the subject string and keep it alive from GC (Java JNI bindings are one example I know of that have that problem). If the bindings use `PCRE2_COPY_MATCHED_SUBJECT`, then the `pcre2_substring_get_*` functions can be provided in a safe way (using the copied subject string).

Can you explain what you are using `PCRE2_COPY_MATCHED_SUBJECT` for?

jcking:
Pseudo C++ code for explanation:

```
class MyMatches {
  // contains pcre2_match_data*
};

class MyPattern {
  // contains pcre2_code*
};

bool match(const char* subject, const MyPattern& pattern, MyMatches& matches);

bool match(std::string_view subject, const MyPattern& pattern, MyMatches& matches);

bool match(std::string&& subject, const MyPattern& pattern, MyMatches& matches);
```

The first two overloads are views and the underlying subjects would outlive the function calls (caller guaranteed) while the third would not and would need to make a copy of the subject for `MyMatches`. Without this overload if you passed `std::string&&` or `std::string` the `std::string_view` overload would be selected and the subject would be immediately destroyed after the function call resulting in a dangling pointer to the subject.

The easiest most performant way to resolve this is with `PCRE2_COPY_MATCHED_SUBJECT` which only makes a copy if the match is successful, otherwise you have to preemptively make a copy. `MyMatches` uses the subject and ovector to return the matches and operates like a range over the matches.

NWilson:
I don't completely understand.

Your MyMatches object knows the subject string? It's not clear why you have to obtain it from PCRE2.

If you are using the ovector, then you don't need `PCRE2_COPY_MATCHED_SUBJECT`, because that argument is just to alter the hidden internal subject copy that PCRE2 uses for the `pcre2_substring_*` APIs.

If you're not using `pcre2_substring_get`, it shouldn't matter to you what string pointer PCRE2 has, because it's not being read? When `pcre2_match` returns, your code has the subject and the offsets, without needing to ask PCRE2 for them.

jcking:
Ah, I suppose I could just avoid using `pcre2_substring_*` APIs and make the copy myself if there is a match. That should work. Never mind then.

NWilson:
OK, if you are using `pcre2_substring_*` APIs that does make sense. In particular the nametable lookups are nice, so there is a good rationale for wanting to keep the string alive.

NWilson:
I'm reopening this.

I think I _do_ like this idea of an API addition.

Don't worry about the conflicts.

There are a lot of things I need to do anyway when I add a new API - symbol versioning declarations, documentation, etc.

I'm happy to take the PR from here.

Many thanks for the contribution!

NWilson:
@jcking I'm afraid I can't push the conflict resolution to your branch. The merge from main updates various GitHub actions CI files, and pushing changes to those (even via a merge commit) is not permitted into someone else's repo.

I will close this PR, and open a fresh one.

However I will retain your commit authorship.

PR #971, opened 2026-09-03
Delete the Windows Run(Grep)Test.bat scripts and provide Run(Grep)Test.py

### LINKED ISSUE #652: Replace RunTest.bat and RunGrepTest.bat with Powershell
Then developers wouldn't need to use windows, since Powershell is available on Linux as well. It's also a sane modern language, compared with .bat, which is virtually impossible to reason about and debug.

I don't think there should be any problem requiring pwsh 5+ to run the tests on windows. It's preinstalled on win 10, and available for older platforms.

NWilson:
Delete the old Windows RunTest.bat, RunGrepTest.bat files. These are hard to maintain for various reasons, around the legacy bat script quoting rules, and hard to keep in sync with the Unix shell scripts.

As a replacement, I have created RunTest.py and RunGrepTest.py. These are cross-platform and identical in behaviour to the shell scripts. From now on, they are the only way to run the test suite on Windows, and can optionally be used on other platforms. I am not (at this time) removing the shell scripts, and Unix/Mac/Linux clients are not forced to use Python.

The main consequence is that Windows clients now need to have Python installed. This can be avoided by turning off the tests (with PCRE2_BUILD_TESTS=OFF in CMake). I do not recommend shipping untested builds, so I expect Windows clients to install Python as necessary.

In future, having a Python test harness could allow us to do more flexible testing, with a wider choice of test suites, including access to test tooling and libraries such as fuzzers.

Fixes #652


maint/UpdateAlways
```diff
@@ -261,8 +260,6 @@ txt_files=(
   )
 
 crlf_files=(
```
NWilson, line 262: Ah, RunShell.bat was meant to have been added here.

RunShell.bat
```diff
@@ -0,0 +1,31 @@
...
+  if not defined RUNSHELL_SCRIPT (
+    echo RunShell.bat: no script was specified. 1>&2
+    exit /b 1
+  )
+  cd /d "%~dp0"
+) else (
+  set "RUNSHELL_SCRIPT=%~1"
+)
+
+if exist "%SystemRoot%\System32\wsl.exe" (
```
NWilson, line 18: Ah. Just because `wsl.exe` exists doesn't mean it's usable. We should query for whether an installation exists firs.

RunGrepTest
```diff
@@ -117,12 +124,23 @@ utf8=$?
...
 fi
 
+case "$pcre2grep_exe" in
+  *.exe|*.EXE) pcre2grep="$pcre2grep --lf";;
+esac
+
+# Set up the srcdir-relative paths for testing from the srcdir.
+
+builddir_from_srcdir="`realpath --relative-to="$srcdir" "$builddir"`"
+pcre2grep_from_srcdir="`realpath --relative-to="$srcdir" "$pcre2grep"`"
```
NWilson, line 142: `--relative-to` isn't in POSIX. I'll need a replacement for that.

PR #972, opened 2026-09-07
pcre2test: fix getall heap over-read with allcaptures/allvector

sage-mode-hunter:
`getall` over-reads the substring list when the reported capture count was widened for the ovector dump.

- pcre2_substring_list_get sizes the stringlist/lengths arrays from the match return and terminates the list with a NULL
- the getall display loop instead ran to capcount, which allcaptures (maxcapcount + 1) and allvector (oveccount) raise above that count
- /(a)|(b)/ on "a" with getall,allcaptures then reads lengths[2] one PCRE2_SIZE past a 68-byte block (ASAN heap-buffer-overflow at copy_and_get)

Splits the overloaded `capcount` into `match_rc` (raw match return, left untouched), `capcount_rc` (pairs held in the ovector) and `capcount_display` (widened by allcaptures/allvector for the ovector dump). `copy_and_get` receives `capcount_rc`, so the getall loop runs to the expected count and keeps its NULL-terminator check. Regression added to test 2.

### DIFF
--- a/src/pcre2test_inc.h
+++ b/src/pcre2test_inc.h
@@ -3810,14 +3810,16 @@ if ((dat_datctl.control & CTL_GETALL) != 0)
     }
   else
     {
-    for (i = 0; i < capcount; i++)
+    /* The list is sized from the match return, and terminated by a NULL. Walk
+    it to that terminator rather than to capcount, which allcaptures/allvector
+    may have raised above the number of pairs the match actually returned. */
+
+    for (i = 0; stringlist[i] != NULL; i++)
       {
       fprintf(outfile, "%2dL ", i);
       pchars(clr_none, stringlist[i], lengths[i], utf, outfile);
       putc('\n', outfile);
       }
-    if (stringlist[i] != NULL)
-      cfprintf(clr_test_error, outfile, "** string list not terminated by NULL\n");
     pcre2_substring_list_free(stringlist);
     }
   }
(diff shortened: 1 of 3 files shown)


NWilson:
Nice catch! Thanks.

Would it be possible to keep the assertion that `pcre2_substring_list_get()` returns the expected number of entries, but instead pass in the correct value for `capcount`?

I think what we want is:
* A variable `match_rc` that stores the return code of the match, and is never touched after that (ie it stays zero, negative - whatever match returned).
* A variable `capcount_rc` that holds the number of captures stored; equal to match_rc if > 0; equal to oveccount if match_rc = 0.
* A variable `capcount_display` that holds the number of captures to print; updated by `allcaptures` and `allvector` to allow printing more than rc.

Then the fix is simply to pass in `capcount_rc` to the copy_and_get code.

We're just reusing one variable for too many things.

sage-mode-hunter:
Done, split into `match_rc` / `capcount_rc` / `capcount_display` as you laid out. `copy_and_get` takes `capcount_rc` now and the NULL-terminator check is back in.

One side effect to flag: for a zero return, `capcount_rc` uses the match data's oveccount (what `pcre2_substring_list_get` uses) rather than `dat_datctl.oveccount`. Those only differ with `ovector=0`, where a DFA overflow used to print nothing after the "too small" line and now prints the pair that was set. No existing expected output changes; full suite is clean under ASan.

NWilson:
Amazing! Thank you Karthik

NWilson (approved):

PR #973, opened 2026-09-07
Add Solaris locale test output variant

vlmarek:
Solaris French locales classify some high-byte characters differently from the upstream PCRE2 locale-test outputs.

Solaris fr treats these bytes as alphabetic/word characters:

  \xb2  ²  superscript two
  \xb3  ³  superscript three
  \xb9  ¹  superscript one
  \xbc  ¼  one quarter
  \xbd  ½  one half
  \xbe  ¾  three quarters

Thanks for pcre2!


RunTest
```diff
@@ -620,18 +620,28 @@ for bmode in "$test8" "$test16" "$test32"; do
             outfile2=$testdata/testoutput3A
             outfile3=$testdata/testoutput3B
             outfile4=$testdata/testoutput3C
+            outfile5=test3outputD
```
NWilson: Unfortunately, there are several other places where the testoutput files will need to be registered. You could search for "testoutput3A" to find some more of them.

RunTest
```diff
@@ -620,18 +620,28 @@ for bmode in "$test8" "$test16" "$test32"; do
...
             outfile=test3output
             outfile2=test3outputA
             outfile3=test3outputB
             outfile4=test3outputC
+            outfile5=test3outputD
             sed "s/fr_FR/$loc/" $testdata/testinput3 >test3input
             sed "s/fr_FR/$loc/" $testdata/testoutput3 >test3output
             sed "s/fr_FR/$loc/" $testdata/testoutput3A >test3outputA
             sed "s/fr_FR/$loc/" $testdata/testoutput3B >test3outputB
             sed "s/fr_FR/$loc/" $testdata/testoutput3C >test3outputC
```
NWilson, line 635: Could we use the same `sed` command on testoutput3D, please, and avoid using iconv? The file for this particular test should be ISO-8859-1 encoded (unless Solaris has some very unexpected encoding choice), so it should be possible to construct and check in the correct output file without needing iconv.

NWilson (changes requested):
I am very sorry this is causing problems!

The testoutput3 data is brittle and quite problematic as a test. It _is_ useful, but hard to get working across different OS codepages. I am genuinely sorry that has caused you trouble.

vlmarek:
Thank you for looking into the PR, this should be more in line with the rest of the tests.

NWilson (approved):
Excellent, thank you very much!

(I had to push a merge commit before I could complete the PR, due to a conflict. Your branch was based off a release branch, which is fine, but merging to main needed a fix.)

vlmarek:
Thank you for your help!

Issue #674, opened 2025-01-25
Modernise code style


mvorisek:
I propose to use clang_format [1] CS formatter to enforce unified CS and minimize diff in a long term.

There are [2] multiple CS rulesets available out of the box matching well formed C CS by industry leaders.

The CS should be normalized in a few iterations (like WS first etc.) and then asserted by CI.

[1] https://clang.llvm.org/docs/ClangFormat.html
[2] https://clang.llvm.org/docs/ClangFormatStyleOptions.html#basedonstyle

NWilson:
I'm afraid this is a "won't fix".

I understand why people feel strongly about code formatting (I do myself sometimes). For teams of 10 or more people working together, I would generally agree that using automated formatting is the way to go, to smooth things out for everyone.

However, we have a very small number of contributors to PCRE2 currently, and we are all happy enough with the current style. It is very unusual, and it took me a while to get used to, but I cannot see any advantage at the moment of reformatting.

It would not be possible to enforce the current style automatically either, since it is followed loosely in the current codebase.

Features and quality come first. The current formatting is not a barrier in any way to actual code improvements.

mvorisek:
Let's unify at least indentation and EOL.

NWilson:
EOL is LF throughout, and indentation is 2 spaces. Those are consistently followed already.

PhilipHazel:
I am, of course, to blame for the current style. I adopted it when I first learned BCPL in 1968/9, before C existed. That's my excuse. However, any change now is up to @NWilson and other current contributors.

zherczeg:
I like the idea of changing the braces, that is probably the most unusal part of the coding style. I am not sure any tool can do that, so we would probably need some scripts to do it though.

mvorisek:
ClangFormat is very configurable, as long as the current CS unified enough, ClangFormat can be configured to match the current CS.

In the ClangFormat docs there is:
```
clang-format -style=llvm -dump-config > .clang-format
```

This should generate `.clang-format` file will all the options of the base CS (`llvm` in the example). The individual format options can be then tweaked to match the current CS.

Surely, there will be some minimal unavoidable changes. But once this is done, the CS can be fully asserted by CI and modernized like improved placement of the braces.

NWilson:
[...]

PR #978, opened 2026-09-11
Bulk reformat: trailing comments

NWilson:
This is preparatory to updating the codebase's indentation style.

The problem is that `/* ... */` are treated differently than `// ...` by code formatters. The line-ending `//...` comments are more tightly attached to a line of code (which is good), whereas the block comments can be moved around more freely when code is wrapped.

Therefore, I have used a simplistic Python script to bulk-convert all single-line block comments which are at the end of a line; which are not on preprocessor lines; and which have code to the left of them.

After running the Python, I visually inspected the diff and fixed up problematic cases by hand.

Part of #674


NWilson:
My Python:

```python
#!/usr/bin/env python3

import re
import sys
from pathlib import Path


def indent_braces(text):
  lines = text.splitlines(keepends=True)
  inside_function = False

  for index, line in enumerate(lines):
    # If it starts with '^\s*#' then ignore the line
    if line.lstrip(" \t").startswith("#"):
      continue

    # Convert trailing /*...*/ comments to // style. Spec: must have non-whitespace
    # before, and be at the end of the line
    m = re.match(r"^(.*?)/\*(.*?)\*/(?=\s*$)", line)
    if m and m.group(1).strip():
      trailpart = line[m.end():]
      line = f"{m.group(1)}//{m.group(2).rstrip()}{trailpart}"

    lines[index] = line

  return "".join(lines)


for filename in sys.argv[1:]:
  path = Path(filename)
  path.write_text(indent_braces(path.read_text()))
```

PR #980, opened 2026-09-11
Fix the JIT prefix scan of negated classes at 16 and 32 bits

mattst88:
In the 16-bit and 32-bit libraries a negated class accepts every code unit above 255, whatever its bitmap says about 255 itself. `scan_prefix()` only took the class to accept those code units when bit 255 was set, so a negated class which leaves out `\xff`, such as `[^\x00-\x40\x42-\xff]`, was taken to accept `A` alone. The JIT then fast-forwarded to the next `A` and missed a `\x{100}` before it, which the interpreter matches:

```
$ pcre2test -16 -jit
/[^\x00-\x40\x42-\xff]/
\x{100}
No match
```

This treats a negated class as accepting everything above the bitmap at those widths, as `pcre2_study()` already does when it builds the start bitmap. 10.47 is affected as well, so it may be worth a backport.

I found this while extending #941 to 16 and 32 bits, and the next revision of that PR builds on this change. The new tests in testinput11 fail under the JIT without the fix. RunTest and pcre2_jit_test pass at all three widths.

### DIFF
--- a/src/pcre2_jit_compile.c
+++ b/src/pcre2_jit_compile.c
@@ -5881,7 +5881,7 @@ fast_forward_char_data *chars_end = chars + MAX_N_CHARS;
 PCRE2_SPTR cc_stack[SCAN_PREFIX_STACK_END];
 fast_forward_char_data *chars_stack[SCAN_PREFIX_STACK_END];
 sljit_u8 next_alternative_stack[SCAN_PREFIX_STACK_END];
-BOOL last, any, class, caseless;
+BOOL last, any, class, caseless, accept_above;
 int stack_ptr, step_count, repeat, len, len_save;
 sljit_u32 chr; /* Any unicode character. */
 sljit_u8 *bytes, *bytes_end, byte;
@@ -6206,6 +6206,14 @@ while (TRUE)
   if (class)
     {
     bytes = (sljit_u8*) (cc + 1);
+    /* The bitmap covers the first 256 code units. Whatever it says about 255,
+    a negated class accepts every code unit above them and a positive class
+    none, and the 8-bit library has none to accept. */
+#if PCRE2_CODE_UNIT_WIDTH != 8
+    accept_above = (*cc == OP_NCLASS);
+#else
+    accept_above = FALSE;
+#endif
     cc += 1 + 32 / sizeof(PCRE2_UCHAR);
 
     SLJIT_ASSERT(last == TRUE && repeat == 1);
@@ -6255,7 +6263,7 @@ while (TRUE)
 
     do
       {
-      if (bytes[31] & 0x80)
+      if (accept_above)
         chars->count = 255;
       else if (chars->count != 255)
         {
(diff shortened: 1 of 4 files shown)


src/pcre2_jit_compile.c
```diff
@@ -6206,6 +6206,14 @@ while (TRUE)
   if (class)
     {
     bytes = (sljit_u8*) (cc + 1);
+    /* Code units above the bitmap are accepted when its highest bit is set.
+    A negated class accepts all of them regardless, since the bitmap only
+    covers the first 256. */
```
zherczeg: Isn't the type (class/nclass) represents the acceptance of chars > 255? If class can accept characters > 255 then nclass is meaningless.

mattst88: Yes, you're right. The opcode alone decides it, as it does in the
interpreter, where a code unit above 255 fails an OP_CLASS and passes an
OP_NCLASS without the bitmap being looked at. Bit 255 was never the right
test: the fix now sets accept_above from the opcode at 16 and 32 bits, and
to FALSE at 8 bits, where there is nothing above the bitmap.

That also covers the other direction. A positive class which includes \xff,
such as [A\xff], used to make its position count as any character, which was
only a missed optimization, and it now fast-forwards to its two characters.
I added it to testinput11 next to the negated case, with \x{1ff} in the
subject to check that only the code unit 0xff itself is taken.

NWilson:
I thought that bit 255 has no special meaning in a class. It does in some of the 'study' code. But OP_CLASS and OP_NCLASS should be treating all the bits in the bitmap exactly the same way. Only class/nclass distinction should matter.

I may be misremembering... I haven't verified that.

zherczeg (approved):
LGTM

zherczeg:
Yes, we are talking about the same thing.

NWilson:
This is a fairly serious match correctness bug - we'll need to see if it's a regression, and how far back it goes in history.

Anything that could case a 'reasonable' regex (like Matt's example) to return incorrect matches is bad enough that I'll triage it for backporting to all the old releases.

NWilson:
@mattst88 Just to let you know - after I have merged all open PRs, I plan to re-indent the whole codebase!

Any branches you have will need to be reformatted and de-conflicted.

Would it be possible for you to pause opening PRs? (Well, you can open them, but any created after today will be held until I'm done.)

I really apologise for this disruption.

https://groups.google.com/g/pcre2-dev/c/Znj5dAtwIAE

mattst88:
No problem -- sounds good to me. I don't have anything else in the queue at the moment.

NWilson:
I confirm affected versions:

10.48
10.47
10.46
10.45
10.44
10.43
10.42
10.41
10.40
10.39
... and likely earlier as well.

I will fix in 10.49-DEV, and produce a patch for backport.

EDIT: _Matt_ has fixed in 10.49-DEV. Many thanks!

PR #985, opened 2026-09-12
Check for the subject end before the first start-bits load

### LINKED ISSUE #984: The new x86 JIT start-bitmap scan can read past (at) the subject end
The x86 SIMD start-bitmap scan can be entered with `STR_PTR == STR_END`. It then performs a 16-byte load from the end pointer. OpenBSD detects this as a segmentation fault under the hardened malloc configuration used by CI.

### Reproduction

The first failure in test 4 is `testdata/testinput4:1859`:

```text
/^[a-z\x{500}-\x{1000}]{3,}[a-h]|x/utf
\= Expect no match
    ab\x{600}ijklm
```

Independent examples occur in test 5 and test 10:

```text
/[Ā-Ą]/utf
\= Expect no match
    \x{105}
```

```text
/[\x{105}-\x{109}]/IBi,utf
\= Expect no match
    \x{100}
```

### Cause

For the test 4 pattern, the start bitmap contains `a-z` and UTF-8 lead bytes `0xd4-0xe1`. Its 40 accepted byte values are below `X86_START_BITS_MAX_COVERED`, so `fast_forward_start_bits_simd()` is selected.

The sequence is:

1. The start scan finds successive candidate characters in `ab\x{600}ijklm`.
2. Matching from the final candidate, `m`, fails.
3. The outer retry check sees that `m` is before `STR_END` and enters `mainloop_entry()`.
4. `mainloop_entry()` advances over `m`, leaving `STR_PTR == STR_END`.
5. It calls the start scan again. The scalar scan checks the endpoint, but the new SIMD call occurs before that check.
6. The SIMD scan performs its first aligned load despite there being no readable code unit.

The earlier outer check is therefore insufficient: `mainloop_entry()` advances `STR_PTR` before calling the start scan.

### Diagnostic trap

This temporary patch was applied to the buggy revision and built successfully:

```diff
diff --git a/src/pcre2_jit_simd_inc.h b/src/pcre2_jit_simd_inc.h
index faad8150..0d635ca3 100644
--- a/src/pcre2_jit_simd_inc.h
+++ b/src/pcre2_jit_simd_inc.h
@@ -542,6 +542,7 @@ struct sljit_label *start;
 struct sljit_label *restart;
 #endif
 struct sljit_jump *quit;
+struct sljit_jump *subject_available;
 start_bits_range ranges[X86_START_BITS_MAX_RANGES];
 sljit_s32 tmp1_reg_ind = sljit_get_register_index(SLJIT_GP_REGISTER, TMP1);
 sljit_s32 data_ind = sljit_get_register_index(reg_type, SLJIT_VR0);
@@ -599,6 +600,10 @@ if (offset > 0)
   add_jump(compiler, &common->failed_match, CMP(SLJIT_GREATER_EQUAL, STR_PTR, 0, STR_END, 0));
   }
 
+subject_available = CMP(SLJIT_LESS, STR_PTR, 0, STR_END, 0);
+sljit_emit_op0(compiler, SLJIT_BREAKPOINT);
+JUMPHERE(subject_available);
+
 /* Load the constants of every range, two registers apiece from VR3 upwards,
    of which a single character needs only the first. */
 value = SLJIT_SIMD_REG_128 | SLJIT_SIMD_ELEM_32 | SLJIT_SIMD_LANE_ZERO;
```

Running test 4 under GDB stopped on `SIGTRAP` with:

```text
STR_PTR (rsi) = 0x555555850310
STR_END (rbx) = 0x555555850310
```

Tests 5 and 10 independently reached the same equality. The generated code immediately following the trap initializes SIMD constants and executes `movdqa (%rsi),%xmm0`.

### Suggested fix

Check the effective scan position unconditionally before the first SIMD load. This covers both the ordinary start bitmap (`offset == 0`) and prefix-class scans (`offset > 0`).

```diff
diff --git a/src/pcre2_jit_simd_inc.h b/src/pcre2_jit_simd_inc.h
index faad8150..8e94ba2c 100644
--- a/src/pcre2_jit_simd_inc.h
+++ b/src/pcre2_jit_simd_inc.h
@@ -589,15 +589,12 @@ if (common->match_end_ptr != 0)
   SELECT(SLJIT_LESS, STR_END, TMP1, 0, STR_END);
   }
 
-/* At offset zero the block the first load reads from is known to hold at least
-   one code unit of the subject, so the load cannot cross into an unmapped page
-   and the end of the subject is checked only once a candidate is found. That no
-   longer holds once STR_PTR is moved forward. */
+/* Move to the requested offset, then ensure that the first vector load has at
+   least one readable code unit. */
 if (offset > 0)
-  {
   OP2(SLJIT_ADD, STR_PTR, 0, STR_PTR, 0, SLJIT_IMM, IN_UCHARS(offset));
-  add_jump(compiler, &common->failed_match, CMP(SLJIT_GREATER_EQUAL, STR_PTR, 0, STR_END, 0));
-  }
+
+add_jump(compiler, &common->failed_match, CMP(SLJIT_GREATER_EQUAL, STR_PTR, 0, STR_END, 0));
 
 /* Load the constants of every range, two registers apiece from VR3 upwards,
    of which a single character needs only the first. */
```

With this guard, the OpenBSD workflow passes and the protected-page reproducer returns `PCRE2_ERROR_NOMATCH` instead of faulting.

mattst88:
The Alpha and x86-64 start-bits scans checked STR_PTR against STR_END before
their first load only when the offset was nonzero. mainloop_entry() can
advance STR_PTR to STR_END before calling the scan, and if STR_END is on a
page boundary the aligned load then reads the next page. Check the end of the
subject unconditionally.

Tested on Alpha under qemu-alpha with the subject placed right before an
unmapped page: patterns such as `[a-c](?=\d)` against `zzzzzzza` fault
without the fix and return no match with it. `pcre2_jit_test` and `RunTest`
pass on Alpha, and ctest passes on x86-64.

Fixes #984

### DIFF
--- a/src/pcre2_jit_simd_inc.h
+++ b/src/pcre2_jit_simd_inc.h
@@ -589,15 +589,12 @@ if (common->match_end_ptr != 0)
   SELECT(SLJIT_LESS, STR_END, TMP1, 0, STR_END);
   }
 
-/* At offset zero the block the first load reads from is known to hold at least
-   one code unit of the subject, so the load cannot cross into an unmapped page
-   and the end of the subject is checked only once a candidate is found. That no
-   longer holds once STR_PTR is moved forward. */
 if (offset > 0)
-  {
   OP2(SLJIT_ADD, STR_PTR, 0, STR_PTR, 0, SLJIT_IMM, IN_UCHARS(offset));
-  add_jump(compiler, &common->failed_match, CMP(SLJIT_GREATER_EQUAL, STR_PTR, 0, STR_END, 0));
-  }
+
+/* STR_PTR may equal STR_END, and the block the first load reads could then be
+   in an unmapped page. */
+add_jump(compiler, &common->failed_match, CMP(SLJIT_GREATER_EQUAL, STR_PTR, 0, STR_END, 0));
 
 /* Load the constants of every range, two registers apiece from VR3 upwards,
    of which a single character needs only the first. */
--- a/src/pcre2_jit_simd_inc.h
+++ b/src/pcre2_jit_simd_inc.h
@@ -3584,15 +3584,12 @@ if (common->match_end_ptr != 0)
   SELECT(SLJIT_LESS, STR_END, TMP1, 0, STR_END);
   }
 
-/* At offset zero the quadword the first load reads from is known to hold at
-   least one code unit of the subject, so the load cannot cross into an
-   unmapped page and the end of the subject is checked only once a candidate is
-   found. That no longer holds once STR_PTR is moved forward. */
 if (offset > 0)
-  {
   OP2(SLJIT_ADD, STR_PTR, 0, STR_PTR, 0, SLJIT_IMM, IN_UCHARS(offset));
-  add_jump(compiler, &common->failed_match, CMP(SLJIT_GREATER_EQUAL, STR_PTR, 0, STR_END, 0));
-  }
+
+/* STR_PTR may equal STR_END, and the quadword the first load reads could then
+   be in an unmapped page. */
+add_jump(compiler, &common->failed_match, CMP(SLJIT_GREATER_EQUAL, STR_PTR, 0, STR_END, 0));
 
 #if defined SUPPORT_UNICODE
 restart = LABEL();


NWilson:
Great! Unconditional checks are safe, and it's fixed in both sites.

I'm very disturbed by the fact that the OpenBSD build is the only one that caught this. We test with valgrind in our CI, which has pretty similar checks for overreads. But nothing failed in CI.

That seems to be because the entire JIT SIMD code is removed when SUPPORT_VALGRIND is defined: and our CI makes sure to turn _on_ the Valgrind build-time support when we run the tests under Valgrind!

@zherczeg - Surely we should fix that, right? This issue is caught immediately on Linux when we run `./RunTest -valgrind` on a build where SUPPORT_VALGRIND=OFF. But we don't test that configuration.

Nor should we test that configuration! Shouldn't we just remove `#if !(defined SUPPORT_VALGRIND)` in `pcre2_jit_simd_inc.h`?

zherczeg (approved):
LGTM

PR #986, opened 2026-09-13
Fix build warning

ltrzesniewski:
I'm not 100% sure about this one, but I noticed GCC 15.2.0 [complains](https://github.com/ltrzesniewski/pcre-net/actions/runs/34720489826/job/103625331689#step:7:87) when building in an Alpine image (with musl), while GCC 13.3.0 is [fine](https://github.com/ltrzesniewski/pcre-net/actions/runs/34720489826/job/103625331714#step:6:86) with that (with glibc).

Historically, UTF-8 allowed up to 6 bytes per code point, but the spec later limited that to 4. Using 6 bytes in this buffer fixes the warning.

I know PCRE2 requires valid UTF-8 and performs a validity check beforehand, but better be safe than sorry, right? :)


NWilson:
Ooh, this is interesting. That is some very fine reasoning it's done there.

Our GETUTF8 macro does read up to 6 bytes, so purely based on local reasoning, it's absolutely right that the character value can be larger than 0x10FFFF (locally). And hence char_othercase can output a uint32_t value that (locally) requires the full 6 byte of buffer space.

I don't think we have a problem in practice: as you say, the input is pre-validated, so we know that in fact the character cannot be larger than 0x10FFFF.

I don't have strong opinion on the best way of silencing the warning, or asserting, or whatever. This PR is certainly a reasonable idea.

Ultimately GETUTF8 has unreachable branches confusing GCC, and it could have the 5/6 byte forms trimmed out. That would also remove the issue.

PR #989, opened 2026-09-16
Add C11 compiler detection (optional) and <stdatomic.h> detection

NWilson:
* The minimum C level remains C99, but C11 is detected and used if available. Both CMake and Autoconf detect it. Bazel cannot detect it, so assumes it. Zig forces use of C11, which is safe as it is modern.
* stdatomic.h is detected in CMake and Autoconf, and assumed to be available on non-Windows platforms by Bazel and Zig.
* Autoconf minimum version is raised to 2.71 (Ubuntu 22.04!). This should not impact downstream clients, since the tarballs and release tags have the Autoconf output included, and clients have never needed to run Autoconf themselves.

I am expecting in future to make (conditional!) use of some C11 features, so I'd like to get this change in, while it's a no-op, in order to probe and validate for support.


NWilson:
Blocked on: https://github.com/google/oss-fuzz/pull/16148

PR #996, opened 2026-09-21
Fix large stack allocation

zherczeg:


zherczeg:
@NWilson I don't know how to test this. We don't have a string generator in pcre2test. Even with that, it is hard to detect that something is wrong.

NWilson:
We do have a syntax `/PREFIX\[REPEAT]{COUNT}SUFFIX/expand` option that should be usable.

I would expect (hope!) that Valgrind shouts very loudly if there are any out-of-bounds accesses, so just having some very large `\g{}` tests should catch it.

However... I'm not sure if we should be disclosing this immediately (so maybe don't commit the precise testcase yet).

src/pcre2_jit_compile.c
```diff
@@ -3643,6 +3645,8 @@ allocate_stack(compiler_common *common, sljit_s32 size)
   DEFINE_COMPILER;
 
   SLJIT_ASSERT(size > 0);
+  if (size > (STACK_GROWTH_RATE / 2))
```
NWilson: I'm a bit confused by the units here. Just below, you adjust STACK_TOP by `size * SSIZE_OF(sw)`, which makes it look like it's counting in units of s32's. But it looks like STACK_GROWTH_RATE is just in plain bytes? In which case the check isn't catching much.

In general, tracking `large_stack_allocation` seems unnecessary - and dangerous, if the detection gets out of sync.

Could you just use the `large_stack_allocator` branch, unconditionally? Adding one or two instructions to do the arithmetic safely seems a good tradeoff. Less complexity, and fewer rarely-exercised branches.

src/pcre2_jit_compile.c
```diff
@@ -15006,7 +15010,15 @@ jit_compile(pcre2_code *code, sljit_u32 mode)
...
   OP1(SLJIT_MOV, SLJIT_R0, 0, ARGUMENTS, 0);
-  OP2(SLJIT_SUB, SLJIT_R1, 0, STACK_LIMIT, 0, SLJIT_IMM, STACK_GROWTH_RATE);
+  if (common->large_stack_allocation)
+  {
+    SLJIT_COMPILE_ASSERT((STACK_GROWTH_RATE & (STACK_GROWTH_RATE - 1)) == 0, stack_growth_must_be_poser_of_2);
+    OP2(SLJIT_SUB, SLJIT_R1, 0, STACK_TOP, 0, SLJIT_IMM, STACK_GROWTH_RATE / 2);
+    OP2(SLJIT_AND, SLJIT_R1, 0, SLJIT_R1, 0, SLJIT_IMM, ~(STACK_GROWTH_RATE - 1));
+  }
+  else
+    OP2(SLJIT_SUB, SLJIT_R1, 0, STACK_LIMIT, 0, SLJIT_IMM, STACK_GROWTH_RATE);
```
NWilson, line 15026: If you unify the branches, this would work well I think:

```
OP2(SLJIT_SUB, SLJIT_R1, 0, STACK_TOP, 0, STACK_LIMIT, 0); // how much we need, negative
OP2(SLJIT_AND, SLJIT_R1, 0, SLJIT_R1, 0, SLJIT_IMM, ~(sljit_sw)(STACK_GROWTH_RATE - 1)); // round to multiple
OP2(SLJIT_ADD, SLJIT_R1, 0, SLJIT_R1, 0, STACK_LIMIT, 0); // shift down STACK_LIMIT
```

For small growth, this gives exactly the same behaviour as before: decrease STACK_LIMIT by one STACK_GROWTH_RATE. But for larger amounts of growth, it decreases STACK_LIMIT by as many increments of STACK_GROWTH_RATE as required.

Your two SUB/ADD instructions are a bit different. They don't seem to guarantee that the new STACK_LIMIT is aligned relative to the old one - only that STACK_TOP is aligned. But I don't know if the actual addresses are necessarily forced to be aligned so tightly? There is a 'simple' malloc-based allocator in the codebase that isn't doing any alignment.

zherczeg: Yes, I also realized the alignment issue.

src/pcre2_jit_compile.c
zherczeg: I prefer to keep the optimized variant when the other is unnecessary. The JIT very rarely needs large allocations.

NWilson (approved):
Looks good.

I think the SUB 'minimum extra space after allocation' can cause allocation to fail unnecessarily. If there is a stack growth maximum, then that SUB could cause the allocation to fail. For example, if the amount required is 15,000 bytes, that would round up nicely to 16*1024, but adding in an extra GROWTH_RATE/2 bumps it up to the next increment, which could fail to be satisfied.

I don't know however whether it is required or not for the stack to have some 'headroom'. If it _is_ required to have headroom, then the standard-size branch doesn't guarantee it (if you request STACK_GROWTH_RATE - 1 then it's just rounded up by one byte).

zherczeg:
I am aware it is possible that it may fail in a few cases. Technically, we could simply allocate STACK_TOP amount of memory, and the code would work. I would assume however, that in most cases deep recursion causes stack increase, and it is better to allocate some extra space to reduce the number of allocations. However, I never measured its actual gain.

zherczeg:
There was a discussion recently about a regex benchmark. Maybe it could be used to measure the perf difference....

zherczeg:
The minimal case would be `OP1(SLJIT_MOV, SLJIT_R1, 0, STACK_TOP, 0);`

zherczeg:
Based on the comments above, the proper fix should be in the jit compiler:
https://github.com/zherczeg/sljit/pull/375

It should be fail safe, and allow allocating the maximum amount of the stack.

NWilson:
It's good to fix it in sljit if that's cleaner.

However - I will need a small/simple fix that's local to PCRE2 for backporting to all the older versions, as this is a serious issue.

If you're happy with the code in this PR (happy that it works) then I don't mind whether it's actually merged or not, as long as it's safe to backport.

zherczeg:
If you want a minimal patch, the code in `common->large_stack_allocation` could be enough.

zherczeg:
Rework in #1007

